PR #1, opened 2016-03-04
[copr-dist-git] Don't assume the SCM repo has the same name as the package

danvratil:
Mock(-scm) expects the cloned directory to has the same name as the package
it's building. To make sure it is so we explicitly pass the package name to
git clone as a destination directory name.

This allows (amont other things) to have a single git repo for multiple spec
files.

### DIFF
--- a/dist-git/dist_git/dist_git_importer.py
+++ b/dist-git/dist_git/dist_git_importer.py
@@ -324,9 +324,9 @@ class MockScmProvider(SrpmBuilderProvider):
 
     def scm_option_get(self):
         return {
-            "git": "git_get='git clone --depth 1 {}'",
-            "svn": "git_get='git svn clone {}'"
-        }[self.task.mock_scm_type].format(self.task.mock_scm_url)
+            "git": "git_get='git clone --depth 1 {} {}'",
+            "svn": "git_get='git svn clone {} {}'"
+        }[self.task.mock_scm_type].format(self.task.mock_scm_url, self.task.package_name)
 
 
 class PyPIProvider(SrpmBuilderProvider):


clime:
Can you provide copr-dist-git import-task description in the following form?

```
{
    "branch": "master",
    "project": "example",
    "source_json": "{\"scm_branch\": \"\", \"scm_url\": \"git@localhost:example.git\", \"scm_type\": \"git\", \"spec\": \"example.spec\"}",
    "source_type": 4,
    "task_id": "5-master",
    "user": "clime"
}
```

This will represent a test case for which the current codebase fails (`scm_url` and `spec` in the `source_json` field are the most import to me at the moment). You might also directly add it into http://pkgs.devel.redhat.com/cgit/tests/copr/tree/Regression/dist-git/fedata/data/import-tasks.json or I'll integrate it later.

danvratil:
Sorry for taking so long:

```
{
    "branch": "master",
    "project": "example",
    "source_json": "{\"scm_branch\": \"\", \"scm_url\": \"git@localhost:example.git\", \"scm_type\": \"git\", \"spec\": \"mypackage.spec\"}",
    "source_type": 4,
    "task_id": "5-master",
    "user": "clime"
}
```
(or do you want a real-world data? I don't know if I can provide any persistent source data for that).

The idea is that package name (which `dist_git_importer.py` extracts from the `spec` key) is different from the name of the repository in `scm_url`. I don't know how the tests work, but this will only actually test the change if the git repo exists and you get all the way to running mock.

I don't have access to devel.redhat.com anymore, so please integrate it for me :)

clime:
Oki, I understand the problem and I'll make a nice test case out of it :). I meant real-world data that I would encapsulate later but I got the point and found one my own repositories that I can reproduce the problem on (https://github.com/clime/example2 with subpkg/example.spec). Anyway, thank you for the fix and the first pull request!

PR #7, opened 2016-05-29
Separate logfile paths

immanetize:
I wanted to install copr-backend and copr-frontend on the same instance, for testing, and there was a file ownership conflict for /var/log/copr.  Since the packages have been separated, and there are seemingly no other technical reasons for these packages to conflict, this set of commits enables copr-backend and copr-frontend to have their own, unshared paths in /var/log.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -175,7 +175,7 @@ Vagrant.configure(2) do |config|
 
     # ..
     frontend.vm.provision "shell",
-      inline: "sudo chown -R copr-fe:copr-fe /var/log/copr-frontend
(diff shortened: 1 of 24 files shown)


xsuchy:
Looks good to me. Just remove the changes in *.spec files. Those will be made by Tito during release.

immanetize:
Neat, I didn't realize Tito did that.  I've simply checkout out the specfiles from master into this branch.

xsuchy:
I squashed those commits about spec file into one.
Merged.  Thank you.

PR #13, opened 2016-09-10
fix for python-progress API

praiskup:
Revert "W: 36,23: Redefining built-in 'max' (redefined-builtin)"

This reverts commit 42c0942b759d58b22c2c2e5551eb67088cf5d9ac.

I'm rather doing pull request, because I'm not sure we are able to avoid this pylint warning.

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -160,7 +160,7 @@ class Commands(object):
         progress_callback = None
 
         if os.path.exists(args.pkgs[0]):
-            bar = ProgressBar(maxval=os.path.getsize(args.pkgs[0]))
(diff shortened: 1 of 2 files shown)


xsuchy:
I dont get it. Why this should be reverted. Can you elaborate?

praiskup:
Because otherwise the progress bar does not work.  Should be reproducible, please try to upload rather large src.rpm (~10MB).

praiskup:
You should have 'python-progress' package installed to be able to reproduce.  While we are on it, we could probably make that package hard dependency on Fedora?

xsuchy:
Merged. Thank you.

PR #15, opened 2016-09-20
Backend: Fix chroot_setup_cmd regex for custom chroot

mizdebsk:
Braces need to be escaped in order not to be interpreted as character
class by Ansible.

CC @msimacek

### DIFF
--- a/backend/backend/mockremote/builder.py
+++ b/backend/backend/mockremote/builder.py
@@ -188,8 +188,7 @@ class Builder(object):
         buildroot_custom_cmd = (
             "dest={cfg_path}"
             " line=\"config_opts['chroot_setup_cmd'] = 'install {pkgs}'\""
-            " regexp=\"config_opts['chroot_setup_cmd'] = ''$\""
-            " backrefs=yes"
+            " regexp=\"config_opts\\['chroot_setup_cmd'\\] = ''$\""
         )
         set_networking_cmd = (
             "dest={cfg_path}"


praiskup:
Which would mean that this regexp never worked before?  Or has something changed in ansible?

praiskup:
Btw., in `mock` there is now new option "additional_chroot_pkgs" .. that can be used instead of this, that could lead to much easier regexp.

mizdebsk:
> Which would mean that this regexp never worked before? Or has something changed in ansible?

Correct, this never worked, AFAIK.

> Btw., in mock there is now new option "additional_chroot_pkgs" .. that can be used instead of this, that could lead to much easier regexp.

The idea of custom chroot is to define which repos are used and which packages are installed (overriding mock defaults), not to provide additional packages. Therefore chroot_setup_cmd should be the right way to do it.

praiskup:
> The idea of custom chroot is to define which repos are used and which packages are installed (overriding mock defaults), not to provide additional packages. Therefore chroot_setup_cmd should be the right way to do it.

I'm afraid I don't understand.  Via `chroot_setup_cmd` is, at the moment at least, chosen the additional list of packages to install into "minial" build chroot.   Copr deals with custom repos on a different place.

mizdebsk:
> I'm afraid I don't understand. Via chroot_setup_cmd is, at the moment at least, chosen the additional list of packages to install into "minial" build chroot. Copr deals with custom repos on a different place.

The "additional packages" field in Copr has different meanings for custom and traditional chroots - for custom chroots it is exhaustive list of packages to install. Actually, there are 2 separate Ansible runs, for Custom and traditional chroots.

See:
- https://github.com/mizdebsk/copr/blob/39c766c0b6f52bd9d8048d185863e800493302b7/backend/backend/mockremote/builder.py#L182-L192
- https://github.com/mizdebsk/copr/blob/39c766c0b6f52bd9d8048d185863e800493302b7/backend/backend/mockremote/builder.py#L203-L206.

praiskup:
Thanks for the info, after off-github discussion, I see now why the `chroot_setup_cmd` needs to be overwritten (because you don't want to install '@build', or whatever is the equivalent for particular mock config).  That means that `chroot_additional_packages` could help "traditional" chroots only.

PR #21, opened 2016-10-11
Revert "[frontend] Bug 1381790 - rename Rawhide to F26 in Copr and

praiskup:
create F27 when Fedora branches instead"

Don't remove `fedora-rawhide-x86_64` chroot.

Also, migration scripts are consumed by Red Hat internal Copr, and
we likely are not able to move to latest upstream source withing
corect period => Migration scripts must be usable regardless of the
time when those are run (in this case even when `rawhide == f28`).

This reverts commit 9ef053b3e88c88697876b2222f1015d27289e089.


clime:
> and we likely are not able to move to latest upstream source withing corect period

There is no correct period. Afterwards, F27 will be added.

>Migration scripts must be usable regardless of the time when those are run

Holds.

clime:
> There is no correct period. Afterwards, F27 will be added.

Better put, every period is correct.

praiskup:
But there is hardwired '26' in the migration script?

praiskup:
Note also that you not only set this policy to your instance, the code is consumed by at least Red Hat internal copr, and I definitely won't remove Fedora Rawhide repository from internal copr (until Fedora itself will have Rawhide).  This is blocker for me.

praiskup:
In other words -- if you really want to change "policy" for Fedora's branching, please do it locally, withing your deployment scripts.  Thanks!

praiskup:
@clime, @msuchy, what should I do then if I still want to have in my frontend `fedora-rawhide-x86_64` chroot?

praiskup:
Please, can you answer?  I don't feel comfortable with reverting without your ACK.  And this is clear misconception - it doesn't make sense to rename chroot in migration script, because that affects every copr instance .... apart from that, changing data by migration script should be done only because you've changed table layout.  This migration changes **just one data field**.

To be honest, I don't understand why the discussion on copr-devel mailing list was **completely ignored**, because everyone there said that this is wrong idea:
https://lists.fedorahosted.org/archives/list/copr-devel@lists.fedorahosted.org/thread/W4AE6EQBXVGQ4MH7MJ5JLW43OZKKHPNH/

Also, I'm bit outdated WRT copr migration scripts -> is it supposed that I'll be able to upgrade postgresql databse running frontend 1.78 or are there other similar issues?

praiskup:
Sorry, I've fixed the revert so the later migration script works.

praiskup:
Yeah, to be discussed -> there's still at least one migration which needs to be reverted to unblock internal copr.  Sorry for delays :(.

praiskup:
Note that removing particular migrations by patches is a lot of work (as will be seen from this PR) and maintaining fork would mean that we would have to backport all migration scripts case-by-case, with some exclusions.  That's way too much work :(.

I would prefer having data edits done by the good old `manage.py`, but I'm open to ideas, @clime, what would be appropriate approach?  Having an opt-in for data migrations?  I'm not even sure that this is possible, but I'd like to consult this so we can move forward.

clime:
What I plan to do in a future version is to move data migrations unrelated to a particular scheme migration into a separate alembic branch so that users can do `alembic upgrade head` for db scheme updates and otherwise use ./manage.py as they are used to.

praiskup:
That sounds cool!

Is it possible to do this "in future", especially after release -> so you'll do 'alembic upgrade head' on production instance?  I mean, is it possible to move the data migrations after release, without breaking the 'alembic upgrade' feature for later copr releases?

clime:
> I mean, is it possible to move the data migrations after release, without breaking the 'alembic upgrade' feature for later copr releases?

Yes, the latest migration is:    414a86b37a0f: "add unique constraint on (copr_id, name) in package table" which will stay on the main branch.

praiskup:
Mostly resolved by commit:
9b97593b3b44a6207b66a90f477997abad2012a8
(in https://pagure.io/copr/copr)

PR #22, opened 2016-10-12
[frontend] generate mock profile (RFC)

praiskup:
Return plain text '.cfg' file on
/coprs/<project>/mock-profile/<chroot>.cfg


praiskup:
For example, the output looks like:
http://209.132.178.105:55555/coprs/praiskup/test-project/mock-profile/fedora-rawhide-x86_64.cfg

For other testing: http://209.132.178.105:55555/coprs

praiskup:
@mizdebsk what do you think?  Does the format of mock profile make sense?

praiskup:
Added links pointing to mock configs, fixed several issues.  PTAL.

cli/copr_cli/main.py
```diff
@@ -309,6 +309,27 @@ def action_fork(self, args):
         result = self.client.fork_project(source=args.src, username=username, projectname=copr, confirm=args.confirm)
         print(result.message)
 
+
+    def action_mock_profile(self, args):
+        """ Method called when the 'list' action has been selected by the
```
mizdebsk: s/list/mock-config/

mizdebsk:
I think that such feature would be useful for build reproducibility, esp. in debugging failing/bad builds.

Regarding format of mock config: I think that it would be better to provide full config, without relying on contents of standard mock conf files. Users can (and often do) modify files in /etc/mock. And you wouldn't need latest mock to use that config.

Regarding the code itself:
1. naming consistency - "mock profile" in API/web vs "mock config" in CLI. Personally I prefer "mock config" due to above reasons, and to mimic "koji mock-config" command.
2. API v2 - AFAICS, you added only legacy API. What about API v2? What about updating API docs?

clime:
This was already discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1357564. For debugging of failing/bad builds, all available mock configs were added into the results directory.

praiskup:
Yeah, this is still RFC.

> I think that such feature would be useful for build reproducibility, esp. in debugging failing/bad builds.

To have reproducible builds, there's full copy-pasted config in backend's result directory.  So this is rather point where the immediate configuration is reflected.

From long term POV, I'dl like to let third-party guys provide copr builders, so they don't want to wait for the pool of VMs generally available (that's a bit of problem internally).  This is first of many steps to let copr builder be more independent from backend and frontend (builder should be able to get configuration from frontend itself), e.g. through API. But it is a long way there... 

> Regarding format of mock config: I think that it would be better to provide full config, without relying on contents of standard mock conf files. Users can (and often do) modify files in /etc/mock. And you wouldn't need latest mock to use that config.

That's right.  OTOH, as @clime said - we have full config dump in build directory.

> Regarding the code itself:
1. naming consistency - "mock profile" in API/web vs "mock config" in CLI. Personally I prefer "mock config" due to above reasons, and to mimic "koji mock-config" command.

Good point.

> 2. API v2 - AFAICS, you added only legacy API. What about API v2? What about updating API docs?

Yeah, this is RFC.  I would really appreciate if @clime was able to discuss politely, I am trying to help and closing the ticket without proper discussion does not help cooperation.

praiskup:
> This was already discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1357564. For debugging of failing/bad builds, all available mock configs were added into the results directory.

I've actually started from that bug-report.  Can you please have a look at the proposal and say what's wrong with that?

praiskup:
> mock configs were added into the results directory.

This is good thing for reproducing the failed builds. But I want to have API call in copr FE so I can immediately get the info for copr builder (similarly to koji), without waiting for the first build in copr.

xsuchy:
I discussed this with @praiskup offline and here are summary:
* People are more likely interested in real configs we used during building. That was already implemented by  https://bugzilla.redhat.com/show_bug.cgi?id=1357564#c4
* It should be clearly stated that this is *very similar* config, useful for debugging.
* I'm not sure if webUI is right way to show it (adds more complexity)
* I am suggesting to add this functionality to copr-cli. You will pass username/copr/chroot and copr-cli will fetch needed data from API as json and copr-cli can construct that config.

praiskup:
Thanks, agreed -> I'll fix the pull request.

praiskup:
Please take another look.

praiskup:
Ping :)

clime:
The code looks good. I have only a few remarks and I miss a test in `beaker-tests/Sanity` to show how this works by example. Do you think you could create a remote pull request already here https://pagure.io/copr/copr/, where we could continue?

praiskup:
@clime I can move this PR, but the pagure project doesn't seem to be updated to this repo.  Can we force push the latest github commit into pagure and make this repo "read only"?

praiskup:
Or @xsuchy, seems like there's only one commit here which is not in pagure, could we --force push the "pagure" repo here, and freeze?

clime:
I can just cherry-pick the commit.

praiskup:
Repositories will be de-synced, however .. shrug.

praiskup:
Merged in https://pagure.io/copr/copr

PR #23, opened 2016-10-12
Fork even dist-git data (RhBug: 1368458)

FrostyX:
I have re-implemented the forking of build data on dist-git - based on FrostyX/copr@1e73866 and also on feodra-copr/copr@5b22368 so I could get rid of abusing `Action` and all the over-factoring around that.

This PR will also fix [RhBug 1368458](https://bugzilla.redhat.com/show_bug.cgi?id=1368458)

### DIFF
--- a/dist-git/dist_git/dist_git_importer.py
+++ b/dist-git/dist_git/dist_git_importer.py
@@ -429,9 +429,7 @@ class ForkProvider(GitProvider):
         cmd = ["fedpkg", "--dist", self.task.branch, "--path", self.git_dir, "srpm"]
         log.debug(" ".join(cmd))
 
-        # @TODO use VM.run(...)
-        proc = Popen(cmd, stdout=PIPE, stderr=PIPE, cwd=self.tmp)
-        output, error = proc.communicate()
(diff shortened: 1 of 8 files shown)


xsuchy:
:+1:

xsuchy:
Please rebase and merge.

clime:
Disagreed. It's not necessary to fork dist-git repo at the moment. The respective bugzilla #1368458 has been already covered by simply not allowing the resubmit action. This will be unlocked again when dist-git builds support copr-dist-git as the source.

clime:
This pull request generally contains several problems, the main one being the `FORK` "source_type", already sounding strange, is supposed to work with copr-dist-git repos but `fedpkg` is used instead of `fedpkg-copr`, which makes some branches like `mga6` unsupported. 

It is also useful to keep the status of forked builds as `forked` instead of `succeeded` in the frontend because it gives a clear message to user about the build source. It is useful because content in the backend build result directory is totally different from what it would be when a full build was done.

Last but not least, the current implementation of dist-git builds could be used to achieve this by employing just the import phase of the whole dist-git build process (like you do it for the `FORK` builds). It will also need `fedpkg-copr` but it can be done.

praiskup:

PR #25, opened 2016-10-19
API for translating module NVR to DNF repo

FrostyX:
The PR adds API endpoint and appropriate `python-copr` method for translating the pair of user (or group) and module NVR ---> URL to module DNF repo file.


FrostyX:
You mean returning something like this?

    http://localhost:5002/results/@copr/coprtestmodule/modules

Not sure if it is necessary, probably not. But IMHO it is much more useful to return URL of the repofile. When you curl the currently returned URL, you will get this:

    [group_copr-coprtestmodule]
    name = Copr modules repo for @copr/coprtestmodule
    url = http://localhost:5002/results/@copr/coprtestmodule/modules
    enabled = 1

and then you:
1. Have to save this whole repofile as `/etc/fm.modules.d/foo.cfg` to be able `dnf module enable ...`
2. Can easily parse the `url = ` part from the repofile if you need to. On the other hand returning only the url and constructing the whole repofile wouldn't be as good.

But I understand your objections. The commit (and PR title) says that it return url to a repo. It should say that it returns url to a repofile, right?

clime:
Well, I thought this feature is good for linking to dependent modules when building new ones. In that case, returning just url should be fine.

xsuchy:
I agree that returning just the url should be better. As it will be mostly consumed by Rida, it will remove the unnecessary parsing. If you can be on safe side, you can implement both.

FrostyX:
I missed the fact that when we are adding external repositories, we add baseurl links from the repofiles, not links to repofiles themselves. Going to change it. TY

clime:
Can you, please, fix the TODOs in the code:

+    # @TODO Current premise is that Copr's name == NVR
+    # @TODO Get rid of it after creating `module` database table

FrostyX:
I would love to, but I think that we need to add `module` table(s) as we discussed it some time ago. So far, all the module information is stored as encoded yaml file in `action.data` and querying that would be slow and ugly.

Basically there are two possible options now:
1. Create `ModulesLogic.get(owner, nvr)` method and move those TODOs there. It won't actually solve anything, but the TODO will be in the logic, not in the API, which might (maybe not) be slightly better.
2. Create the proper DB table (definitely within new PR)

Let's discuss it on Monday meeting, shall we?

clime:
The url classification for modules should be owner/project/nvr. That's how it is done. We can also support owner/nvr but that by returning list of module repos and let a client cope with it.

Also, module repo should be a property (read @property :)) of a module and returned as such but having this method for the time being is quite fine by me.

Please, if you may, move this request to https://pagure.io/copr/copr by filling a "Remote pull request" for the final phase before merge as we are moving there and I'd like to use it already for this.

praiskup:

PR #26, opened 2016-10-30
Module table

FrostyX:
This PR:
- Creates `module` table and binds the `Module` with it
- Seeds the `module` table with current modules from `action` table
- Updates everything related to get it working with new model (not the interesting part)
- Task to create module on backend should be untouched and `action` dict returned by `/backend/waiting/` should remain exactly the same, except for the fact, that it was constructed from both `action` and `module` tables.

Suggested table doesn't cover everything and I am absolutely sure that we will need to extend this table with some other properties in the future. I've focused on the things that we _need_ now, such as unique constraint on user/nvr, #25 and also the extensibility ensured by storing the [yaml_b64](#diff-85e42e918c6b68f6d912b320723632e7R1130)


FrostyX:
I made some changes on this:

- Refactored and commented the migration script, @xsuchy 
- Added support for groups (i.e. when module is built from group copr, it is named as `@group/nvr`
- Couple of fixes that were rebased into previous commits

So, what do you think?

xsuchy:
:+1:

clime:
My main objection is that you don't need those two unique constraints on (module, user) and (module, group) and you don't need to keep the owner records in the table. It is good enough to have just `copr_id` in the table with a single unique constraint (module, copr). Module owner is the same as the project owner.

Please, if you may, move this request to https://pagure.io/copr/copr by filling a "Remote pull request" for the final phase before merge as we are moving there and I'd like to use it already for this. It would be also useful if you referenced this original thread in the "Initial comment" for the remote pull request.

praiskup:

PR #2373, opened 2022-11-21
Build packages using GitHub actions

praiskup:


praiskup:
Unfortunately, we can't test this workflow without merging it first.  So to actually fix any potential bugs I'd appreciate an early merge.

But I believe it should work, at least according to my tests in [my fork](https://github.com/praiskup/copr/pull/3).  That one can not succeed because the copr-side configuration is set to clone from this repo, not from my fork.  So the custom method will naturally fail on an assertion.

.github/workflows/copr-builds.yml
```diff
@@ -0,0 +1,55 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
```
FrostyX, line 15: Unwanted C&P or expected?

FrostyX:
Briefly reviewed and LGTM.
I'd say merge and let's see the bugs in action.

PR #2382, opened 2022-11-24
backend, distgit: require redis.service to be started

FrostyX:
Currently we do this in ansible

    service: name="{{ item }}" enabled=yes state=started
    with_items:
    - redis       # TODO: .service in copr-backend should depend on redis

which isn't the best solution.


praiskup:
+1

PR #2379, opened 2022-11-23
Use OpenPGP v4 signatures for signing packages

pmatilai:
v3 signatures which obs-signd currently creates by default were obsolete in 1998 already, but needed for interoperability in the early millennium. Everything from the last 15 years supports v4 signatures, lets get on with the times.

This requires at obs-signd 2.5.2 (released in 2018) or newer. Perhaps worth noting that the -4 switch is undocumented in obs-sign man pages.

For more background, see https://bugzilla.redhat.com/show_bug.cgi?id=2141686 and https://github.com/openSUSE/obs-sign/issues/43

PR #2383, opened 2022-11-24
backend: use OpenPGP v4 signatures

praiskup:
From Panu's statement in #2379:
v3 signatures which obs-signd currently creates by default were obsolete in 1998 already, but needed for interoperability in the early millennium.  Everything from the last 15 years supports v4 signatures, lets get on with the times.

This requires at obs-signd 2.5.2 (released in 2018) or newer. Perhaps worth noting that the -4 switch is undocumented in obs-sign man pages.

For more background, see
https://bugzilla.redhat.com/show_bug.cgi?id=2141686 and openSUSE/obs-sign#43

Closes: #2379
Closes: #2383

### DIFF
--- a/backend/copr_backend/sign.py
+++ b/backend/copr_backend/sign.py
@@ -88,7 +88,7 @@ def get_pubkey(username, projectname, log, outfile=None):
 
 
 def _sign_one(path, email, hashtype, log):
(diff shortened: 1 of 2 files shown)


FrostyX:
+1,
we tested it with @praiskup, and it works even for RHEL6.

PR #2385, opened 2022-11-25
A couple of beaker tests fixes

FrostyX:
Discovered when upgrading to F37


beaker-tests/Sanity/copr-cli-basic-operations/runtest.sh
```diff
@@ -485,7 +485,7 @@ rlJournalStart
         rlRun "copr-cli delete ${NAME_PREFIX}wrong-name" 1
 
         # test building for armhfp
-        rlRun "copr-cli create --chroot fedora-$FEDORA_VERSION-armhfp ${NAME_PREFIX}ArmhfpBuild"
+        rlRun "copr-cli create --chroot fedora-$PREV_FEDORA_VERSION-armhfp ${NAME_PREFIX}ArmhfpBuild"
```
praiskup, line 488: Can you use 36 constant instead?  And then, once this test starts failing again - it means that we dropped the fedora-36 chroot and we'll drop the entire test (in-line comment would be useful).

FrostyX: That's a good idea, updated.

praiskup:
Thank you!

PR #2396, opened 2022-11-29
doc: notes for the yesterday's F35 => F37 move

praiskup:
The LetsEncrypt certificates need to be copied form old server => batcave => new server.  We need to assure the backed-up certificates are not too old.

Also, DistGit is now storing all the data, including logs, on one (large) volume.


FrostyX (approved):
+1

PR #2400, opened 2022-11-29
fix database pvc path

pkking:
fix database pvc path not match, change the pod restart strategy to recreate because the pvc is claim as ReadWriteOnce
also make nginx mark the log.gz can be shown directly

Signed-off-by: Li Chaoran <pkwarcraft@gmail.com>

### DIFF
--- a/kubernetes/kustomize/backend.yaml
+++ b/kubernetes/kustomize/backend.yaml
@@ -4,6 +4,8 @@ metadata:
   name: copr-backend
 spec:
   replicas: 1
+  strategy:
+    type: Recreate
   selector:
     matchLabels:
       component: copr-backend
--- a/kubernetes/kustomize/database.yaml
+++ b/kubernetes/kustomize/database.yaml
@@ -4,6 +4,8 @@ metadata:
   name: copr-database
 spec:
   replicas: 1
+  strategy:
+    type: Recreate
   selector:
     matchLabels:
       component: copr-database
@@ -53,7 +55,7 @@ spec:
                   key: database-name
                   name: copr-database
           volumeMounts:
-            - mountPath: /var/lib/pqsql/data
+            - mountPath: /var/lib/pgsql/data
               name: postgresql-data
       volumes:
         - name: postgresql-data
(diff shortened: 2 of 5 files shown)


pkking:
@TommyLike could you please check the PR :)

kubernetes/kustomize/config/backend/nginx.conf
```diff
@@ -14,4 +14,12 @@ server {
 	    port_in_redirect off;
 		autoindex on;
 	}
+
```
TommyLike, line 17: If this is a common issue for backend nginx configuration, you should seperate this change and change the nginx configuration located in the docker folder as well.

TommyLike (commented):
Thanks @pkking

kubernetes/kustomize/config/backend/nginx.conf
pkking: I will try it with docker and if it works, let the backend-httpd pod use the file from container looks reasonable.

FrostyX:
Thank you for the PR,
let us know when you are satisfied with the PR and we will merge.

pkking:
@TommyLike reworked
Move the nginx.conf from kubernetes to docker, so that the nginx in docker-compose and k8s will benefit
@FrostyX @praiskup please also take a look at the backend_httpd nginx.conf changes, it will print the nginx log directly to stdout/stderr :)

TommyLike:
@pkking you need move the configuration change in docker folder into seperated PR I guess.

praiskup:
> Gotcha, will make this PR clearly.

I typically do `git fetch origin && git rebase -i origin/main` ... (before saving and closing the editor, reorganize the patches, drop some, etc.).

pkking:
> > Gotcha, will make this PR clearly.
> 
> I typically do `git fetch origin && git rebase -i origin/main` ... (before saving and closing the editor, reorganize the patches, drop some, etc.).

Thanks, reworked the PR using this way, super easy :) @TommyLike could you please take a look again?

TommyLike:
@TommyLike thanks for fixing this! looks good to me now.

praiskup (approved):

PR #2405, opened 2022-11-29
frontend, backend: fix traceback for non-existing tasks

### LINKED ISSUE #2390: copr backend traceback
```
[2022-11-28 19:40:15,409][ ERROR][PID:350650][/usr/bin/copr-backend-process-build.managed.pid-350650][background_worker_build.py:handle_task:792] Unexpected exception
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 786, in handle_task
    self.handle_build()
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 778, in handle_build
    self._get_build_job()
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 344, in _get_build_job
    self.job = BuildJob(resp.json(), self.opts)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/copr_backend/job.py", line 75, in __init__
    for key, val in task_data.items():
                    ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

FrostyX:
Fix #2390

First, take a look at the frontend change. The
`BuildsLogic.get_srpm_build_task` never raises an exception. It returns `None` for non-existing tasks instead. The route therefore returns 200 status code even for non-existing tasks, and simply returns `nil`. I am changing this to return 404 instead.

This prompted the backend changes because exceptions started appearing.

### DIFF
--- a/backend/copr_backend/frontend.py
+++ b/backend/copr_backend/frontend.py
@@ -8,7 +8,8 @@ import logging
 from copr_common.request import SafeRequest, RequestError
 from copr_backend.exceptions import FrontendClientException
 
-MIN_FE_BE_API = 4
+# The frontend counterpart is in `backend_general:send_frontend_version`
+MIN_FE_BE_API = 5
 
 class FrontendClient:
     """
@@ -75,7 +76,7 @@ class FrontendClient:
             response = request.send(url, method=method, data=data)
             return response
         except RequestError as ex:
-            raise FrontendClientException from ex
+            raise FrontendClientException(str(ex)) from ex
 
     def update(self, data):
         """
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -1,6 +1,4 @@
 import flask
-import sqlalchemy
-
 from copr_common.enums import StatusEnum
 from coprs import db, app
 from coprs import models
@@ -9,7 +7,7 @@ from coprs.logic.builds_logic import BuildsLogic
 from coprs.logic.complex_logic import ComplexLogic, BuildConfigLogic
 from coprs.logic.packages_logic import PackagesLogic
 from coprs.logic.coprs_logic import MockChrootsLogic, CoprChrootsLogic
-from coprs.exceptions import MalformedArgumentException, ObjectNotFound
+from coprs.exceptions import CoprHttpException, ObjectNotFound
 from coprs.helpers import streamed_json
 
 from coprs.views import misc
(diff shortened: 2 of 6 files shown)


praiskup:
Would you mind bumping `MIN_FE_BE_API`?

praiskup:
Why backend even asked for those build IDs?

FrostyX:
> Why backend even asked for those build IDs?

I am thinking of something like 

1. User submitted a build
2. Backend added it to its priority queue
3. Because of one of the post-release bugs, the building was stuck
4. User deleted the build because it wasn't working
5. Before updating the priority queue, backend picked the build and started it

I need to check if this was the case and if this situation can happen. If yes, we should fix it.

FrostyX:
> I am thinking of something like

Now it looks more like an issue caused by #2398.

praiskup:
So are we going to skip this PR then?

FrostyX:
We should probably skip this for the next mini-release. But the change is valid, I think.

backend/copr_backend/frontend.py
```diff
@@ -8,7 +8,8 @@
 from copr_common.request import SafeRequest, RequestError
 from copr_backend.exceptions import FrontendClientException
 
-MIN_FE_BE_API = 4
+# The frontend counterpart is in `backend_general:send_frontend_version`
+MIN_FE_BE_API = 5
```
praiskup, line 12: Further reading this code, I think I was wrong and this bump is not needed?  The frontend/backend changes in this PR are independent, right?

Anyways, the documentation here is very nice, having it (as a separate commit) woudl be nice.

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -348,13 +349,13 @@ def _stream():
 @backend_ns.route("/get-build-task/<task_id>/")
 @backend_ns.route("/get-build-task/<task_id>")
 def get_build_task(task_id):
-    try:
-        task = BuildsLogic.get_build_task(task_id)
-    except MalformedArgumentException:
```
praiskup, line 353: We are actually handling the same problem, and raising this exception inside `BuildsLogic.get_build_task`.  I'd prefer to handling the problem with just one method.  Perhaps we'd just change [first()](https://github.com/fedora-copr/copr/blob/5f30af8106c226823fb25eb447506b829c3d18d2/frontend/coprs_frontend/coprs/logic/builds_logic.py#L390) call to `one()`?  Then we should get the `NoResultFound` and it would be handled on line 357 here.

praiskup:
Thank you for this patch, it makes the code more robust..

backend/copr_backend/frontend.py
FrostyX: I think doing the `MIN_FE_BE_API` bump is a good idea because, at this moment, the backend can't handle a non-200 response. If we start returning 404 and 500 status codes from frontend, the `FrontendClient` will start raising exceptions. At the same time `_get_build_job` doesn't expect them and checks for `if resp.status_code != 200:` instead. I think we need both frontend and backend changes at the same time.

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
FrostyX: Makes sense, I updated the PR, WDYT?
The change is in a separate commit for easier review.

praiskup (approved):

PR #2406, opened 2022-11-29
backend, dist-git: start services after redis.service

praiskup:
On top of Requires=, we need also After= so redis is started first. Long term (for the OpenShift scenario) we should make the Redis client connections resilient against the temporary Redis server outages.

The copr-backend.target depends on three services which have the dependency resolved between each other (actions/builds depend on log, log depends on redis).

With copr-dist-git.service it is much simpler, we just depend on redis.service.


FrostyX:
+1

PR #2421, opened 2022-12-05
add openeuler logo

pkking:


praiskup:
Thank you for the quick PR, @pkking!

The icon will be downsized a lot, so I don't think it is wise to provide the text in the logo.  When downsized, it would be unreadable and the icon would be unnecessarily small because of that. E.g. look at the [Red Hat](https://github.com/fedora-copr/copr/blob/74f7586a90bfce30fa4dbe54120532270a5c1bf9/frontend/coprs_frontend/coprs/static/rhel_logo.png). Have  a look [how logos look when downsized](https://copr.fedorainfracloud.org/coprs/praiskup/ping/).

pkking:
Ah, i agree, the logo need to be downsized :)

PR #2430, opened 2022-12-08
switch distgit to newest container

pkking:
New distgit pod will share logs between containers.
Distgit is using tcp port 5001.


TommyLike:
@pkking thanks for simplifying this, now we have fewer images. looks good to me.

PR #2432, opened 2022-12-10
frontend: add hover for table rows

FrostyX:
Fix RHBZ 2152268

Our v3 bootstrap supports `.table-hover` class for tables but its has a blue color which is too inconsistent with the rest of our colorscheme. I am adding the hover via CSS and using the gray color that is used by bootstrap v4.


praiskup:
+1 (haven't tested)

nikromen (approved):

PR #2436, opened 2022-12-12
frontend: add commands for adding and removing warning banner

### LINKED ISSUE #2319: Show warning banner on copr frontend before release
Original issue: https://pagure.io/copr/copr/issue/2319
Opened: 2022-09-26 12:27:44
Opened by: nikromen

nikromen:
TODO:
- [x] docs update

Fixes #2319

### DIFF
--- a/doc/how_to_release_copr.rst
+++ b/doc/how_to_release_copr.rst
@@ -10,6 +10,15 @@ good to know what's ahead.
 Keep amending this page if you find something not matching reality or expectations.
 
 
+Create warning banner about release
+-----------------------------------
+
+
+Use script in frontend's ``manage.py`` to announce an outage::
+
+    ./manage.py warning-banner --outage_time "2022-12-31 13:00-16:00 UTC" --ticket 1234
+
+
 Pre-release
 -----------
 
@@ -421,6 +430,11 @@ It is often good idea to put new (filtered) ``%changelogs`` entries there.
 
 Final steps
 ...........
+
+Remove the warning banner from frontend page by::
+
+    ./manage.py warning-banner --remove
+
 
 Check if the `MODIFIED bugs <https://bugzilla.redhat.com/buglist.cgi?bug_status=POST&bug_status=MODIFIED&classification=Community&list_id=4678039&product=Copr&query_format=advanced>`_
 (that are not ON_QA) are fixed in released Copr or not, move them ON_QA.
--- a/frontend/coprs_frontend/coprs/templates/banner-include.html
+++ b/frontend/coprs_frontend/coprs/templates/banner-include.html
@@ -1,0 +1,16 @@
+<div class="alert alert-danger" role="alert">
+    {% if outage_time %}
+        <b>Warning!</b> Scheduled outage of Copr servers - {{ outage_time }}.
+    {% endif %}
+    {% if ticket %}
+        For more information please refer to
+        <a href="https://pagure.io/fedora-infrastructure/issue/{{ ticket }}">this ticket</a>.
+    {% endif %}
+    <!--
+    Possible rest of short html code. Something like:
+    "Check <a href="#" class="alert-link">this document</a> for details."
+    -->
+    {% if rest %}
+        {{ rest }}
+    {%  endif %}
+</div>
--- a/frontend/coprs_frontend/commands/warning_banner.py
+++ b/frontend/coprs_frontend/commands/warning_banner.py
@@ -1,0 +1,51 @@
+"""
+Command for managing warning banner on the top of frontend page.
+"""
+
+
+import os
+import sys
+from pathlib import Path
+
+import click
+from jinja2 import Template
+
+from coprs.constants import BANNER_LOCATION
+
+
+@click.command()
+@click.option("--outage_time", "-o", default=None, help="start time of the outage")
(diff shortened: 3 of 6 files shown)


frontend/coprs_frontend/commands/add_warning_banner.py
```diff
@@ -0,0 +1,33 @@
...
+from flask import render_template
+
+
+@click.command()
+@click.option("--outage_time", "-o", default=None, help="start time of the outage")
+@click.option("--ticket", "-t", default=None, help="fedora infra ticket ID")
+@click.option("--rest", default=None, help="additional code")
+def add_warning_banner(
+    outage_time: Optional[str], ticket: Optional[int], rest: Optional[str]
+) -> None:
```
FrostyX, line 14: We will need to discuss the amount of explicit types in our codebase. Although I very much like static typing and the advantages it brings ... in cases like this, it is just unnecessary clutter. I will make my case in https://github.com/fedora-copr/debate and we can debate there.

PS: I am not at all saying that I don't want to use explicit types. I quite like them, even though I think python chose an ugly syntax :D

frontend/coprs_frontend/commands/add_warning_banner.py
```diff
@@ -0,0 +1,33 @@
...
+    if outage_time is None and rest is None:
+        print("Error: `outage_time` or `rest` should be present.", file=sys.stderr)
+        return
+
+    with open("/var/lib/copr/banner-include.html", "w") as banner:
+        banner.write(
+            render_template(
+                "banner-include.html", outage_time=outage_time, ticket=ticket, rest=rest
+            )
+        )
```
FrostyX, line 33: This would be shorter :-)

```python
text = render_template("banner-include.html", outage_time=outage_time,
                       ticket=ticket, rest=rest)
banner.write(text)
```

frontend/coprs_frontend/commands/remove_warning_banner.py
```diff
@@ -0,0 +1,14 @@
+import os
+
+import click
+
+from coprs.constants import BANNER_LOCATION
+
+
+@click.command()
+def remove_warning_banner() -> None:
```
FrostyX, line 9: I would prefer merging this with `add_warning_banner` and simply having `warning_banner` with an optional parameter `--remove`. Or at least having these two functions in the same file.

FrostyX (commented):
Nice, thank you for the PR, this would be a helpful feature during release. I made a few notes, and also the documentation change is needed.

frontend/coprs_frontend/commands/add_warning_banner.py
FrostyX: My type hinting proposal: https://github.com/fedora-copr/debate/pull/3

frontend/coprs_frontend/coprs/templates/banner-include.html
```diff
@@ -0,0 +1,14 @@
+<div class="alert alert-danger" role="alert">
+    {% if outage_time and ticket %}
+        <b>Warning!</b> Scheduled outage of Copr servers - {{ outage_time }}. For more information
+        please refer to <a href="https://pagure.io/fedora-infrastructure/issue/{{ ticket }}">this
```
xsuchy, line 4: Hmm, this is to hardcoded. Especially for people who run their own instance. I would have nothing against it if this file will be in /etc/copr/ and marked as %config

doc/how_to_release_copr.rst
```diff
@@ -10,6 +10,15 @@ good to know what's ahead.
 Keep amending this page if you find something not matching reality or expectations.
 
 
+Create warning banner about release
+-----------------------------------
+
+
+Use script in frontend's ``manage.py`` to announce an outage::
+
+    ./manage.py manage-warning-banner --outage_time "2022-12-31 13:00-16:00 UTC" --ticket 1234
```
xsuchy, line 19: manage manager warning banner? This is nitpicking, but can we have just
`manage.py warning-banner` ?

frontend/coprs_frontend/coprs/templates/banner-include.html
```diff
@@ -0,0 +1,14 @@
+<div class="alert alert-danger" role="alert">
+    {% if outage_time and ticket %}
+        <b>Warning!</b> Scheduled outage of Copr servers - {{ outage_time }}. For more information
```
xsuchy, line 3: hmm, if ticket is not specified then the main headline will not be displayed? Can you split it to two independent if?

frontend/coprs_frontend/coprs/templates/banner-include.html
nikromen: yes it is hardcoded. I meant this to be really specific and strict for our release use cases since we do them often and it is a very specific banner (and we want basically every time the same warning message).

what do you mean by marked as %config? what do I achieve from it?

xsuchy: When the file is in spec file marked as %config then it can be modified by user and it will not be overwritten during `dnf upgrade` but .rpmnew file will be created instead.

xsuchy:
I still have a few ideas on how to enhance it, but ... yeah, this works and is better than before. +1 from me.

PR #2438, opened 2022-12-13
fix: pod strategy set to recreate as attached volume is ReadWriteOnce

pkking:
The new pod will hangs in attaching the pvs since the old one still holding it.


pkking:
@TommyLike please reviews ...

TommyLike:
@pkking thanks for your fix:)

Issue #2389, opened 2022-11-28
copr-sign fails with obs-sign 2.6.1


praiskup:
https://bodhi.fedoraproject.org/updates/FEDORA-2022-4779bd5643

we manually downgraded in production for now to obs-signd-2.5.10-4.20210907git5c32050.fc37.x86_64 (on keygen, server-side)

obs-signd-2.6.1-1.x86_64 works fine on copr-be (client side)

praiskup:
We should reproduce on stage, and ask @xsuchy for help maybe.

FrostyX:
I think the issue was

```
[2022-11-30 21:23:08,316][WARNING][PID:3452181] Command '/bin/sign -4 -h sha256 -u tmz#git@copr.fedorahosted.org -r /var/lib/copr/public_html/results/tmz/git/srpm-builds/05076356/git-2.39.0-0.1.rc1.src.rpm' failed with: gpg-agent: 67108871 Bad secret key <GPG Agent>
gpg-agent: 67108871 Bad secret key <GPG Agent>
```

FrostyX:
Somehow, the `obs-signd` package upgraded, so I downgraded it again and did

```
[root@copr-keygen ~][PROD]# dnf versionlock add obs-signd-2.5.10-4.20210907git5c32050.fc37.x86_64
```

praiskup:
```
$ dnf history info 26
Transaction ID : 26
Begin time     : Wed 30 Nov 2022 08:30:12 PM UTC
Begin rpmdb    : c98648fabfbf90727b5ef8f1704f878fff4fde5baf7fc5d16dee395612e39c01
End time       : Wed 30 Nov 2022 08:30:59 PM UTC (47 seconds)
End rpmdb      : c40d9be380733f05f3e950111a1d130d9da9c68d93257aadcf36ce3ae3dde4b7
User           : fedora Cloud User <fedora>
Return-Code    : Success
Releasever     : 37
Command Line   : -y update
Comment        : 
Packages Altered:
    Install  kernel-core-6.0.10-300.fc37.x86_64                @updates
    Upgrade  bluez-5.66-4.fc37.x86_64                          @updates
    Upgraded bluez-5.65-3.fc37.x86_64                          @@System
    Upgrade  fontconfig-2.14.1-2.fc37.x86_64                   @updates
    Upgraded fontconfig-2.14.1-1.fc37.x86_64                   @@System
[...]

PR #2453, opened 2023-01-04
docker: make keygen container work with obs-sign 2.6.1

FrostyX:
See #2389


docker/keygen/files/signd-entrypoint
```diff
@@ -13,6 +13,7 @@ gpg: /usr/bin/gpg-copr
 allowuser: $(id -u)
 # we can't open a privileged (for the remote keygen server)
 allow-unprivileged-ports: true
+gnupghome: /var/lib/copr-keygen/gnupg/
```
praiskup, line 16: I have no objections against the change, but it would be nice to have a commit message explaining how this helps

FrostyX:
The issue was elsewhere https://github.com/fedora-copr/copr/issues/2389#issuecomment-1371020534, closing.

PR #2457, opened 2023-01-05
docs: mdadm --add helped us during the F35 => F37 move

praiskup:


FrostyX:
I don't remember the exact command we executed, but LGTM

praiskup:
I've found it in command-line history actually.  Thank you for the review!

PR #2460, opened 2023-01-09
frontend: fix homepage user count

### LINKED ISSUE #2448: Homepage user count issue
Recently i notice our new deployed homepage displayed a wrong user count like `openEuler Copr hosts 4 projects from
0 openEuler contributors`, i checked the db, there are 2 user records, should `get_multiple_with_projects` go without `~` ?
```
    @classmethod
    def get_multiple_with_projects(cls):
        """
        Return all users that have at least one project (deleted projects
        counts as well)
        """
        return User.query.filter(~User.coprs.any())
```

FrostyX:
Fix #2448

The tilde doesn't make any sense, it had to be a copy-paste error or something. According to the documentation
https://docs.sqlalchemy.org/en/20/core/operators.html tilde means `NOT`, so we were returning users without projects, not users with projects.

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/users_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/users_logic.py
@@ -28,7 +28,7 @@ class UsersLogic(object):
         Return all users that have at least one project (deleted projects
         counts as well)
         """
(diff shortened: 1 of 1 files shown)


praiskup:
+1

PR #2462, opened 2023-01-10
Revert "Add badges for LGTM.com"

xsuchy:
This reverts commit ebfa21d5cc038136cb8f05da3dc36e2a99916083.

Because of https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/ LGTM has been replaced code scanning feature directly in GH.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,6 @@ python-copr-common | [![badge](https://copr.fedorainfracloud.org/coprs/g/copr/co
 
 [//]: # (Please generate this table by ./build_aux/generate-build-status-table)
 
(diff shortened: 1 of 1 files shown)


nikromen (approved):
I can see that we don't have codeql (and its corresponding advanced settings in `.github/workflows`) enabled. So we shouldn't forget to do it.

Otherwise +1 since the lgtm grade and alert counter is no longer available.

praiskup:
Using LGTM was requested by Red Hat Security team.  This might bring a regression in project compliance.  I'd prefer having an alternative first.

FrostyX:
I am sorry, I added the LGTM.com badges quite recently, and didn't notice that the service is being shut down.

The recommended replacement seems to be a GitHub built-in code-scanning feature. I am enabling it in PR #2464.

xsuchy:
> I'd prefer having an alternative first.

The service already does not work. No matter if this PR is merged or not. :)

Issue #2463, opened 2023-01-10
unable to set chroot-specific configuration "additional_modules"


TomasTomecek:
We have just implemented configuration of chroot-specific options in Packit. Unfortunately, getting this exception:

> Module name 'foreman:el8 ruby:2.7 nodejs:17' must consist of two parts separated with colon, e.g. module:stream

Which is strange, because this is the call we are making:

```
[2023-01-10 15:29:48,061: INFO/ForkPoolWorker-1] Trying to get packit-stg/TomasTomecek-foreman-1 Copr project. The project will be created if it does not exist.
[2023-01-10 15:29:48,312: INFO/ForkPoolWorker-1] There is chroot-specific configuration for centos-stream-8-x86_64
[2023-01-10 15:29:48,562: INFO/ForkPoolWorker-1] Update packit-stg/TomasTomecek-foreman-1 centos-stream-8-x86_64: {'additional_repos': ['http://koji.katello.org/releases/yum/foreman-nightly/el8/x86_64/'], 'additional_modules': ['foreman:el8', 'ruby:2.7', 'nodejs:17']}
```

is this syntax wrong?

```
'additional_modules': ['foreman:el8', 'ruby:2.7', 'nodejs:17']
```

Context: https://github.com/TomasTomecek/foreman/pull/1/checks?check_run_id=10555208826

praiskup:
Tha API documentation has a bug from eece7ac2724b1289a3069095a286884db0cc90b3 - the API option accepts a comma-separated list of modules, same as the Copr cli:

    $ copr edit-chroot --help | grep -A2 -- '^  --modules' 
      --modules MODULES     comma separated list of modules that will be enabled
                            or disabled in the given chroot, e.g.
                            'module1:stream,!module2:stream'

praiskup:
Thank you for the report!

evgeni:
@praiskup but the `get-chroot` call gives an array?

```console
% copr get-chroot @theforeman/pr-testing-playground/centos-stream-8-x86_64
{
    "additional_modules": [
        "ruby:2.7,",
        "nodejs:12,",
        "foreman:el8"
    ],
    "additional_packages": [],
    "additional_repos": [],
    "comps_name": null,
    "delete_after_days": null,
    "isolation": "unchanged",
    "mock_chroot": "centos-stream-8-x86_64",
    "ownername": "@theforeman",
    "projectname": "pr-testing-playground",
    "with_opts": [],
    "without_opts": []
}
```

[...]

PR #2465, opened 2023-01-11
python: fix docs for project_chroot additional_modules option

praiskup:
Fixes: #2463

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -1416,6 +1416,54 @@ class PagureIntegrationForm(BaseForm):
             self.repo_url.data = repo_url
 
 
+class AbstractSeparatedListField(wtforms.Field):
+    """
+    For API queries (uploaded JSONs), we prefer sending data as list() over
+    comma/space separated string of string items.  In Web-UI though, we prefer
+    sending comma-separated strings (easier to specify in the available UI
+    widgets, ie text-area or normal input).
+
+    Though WTForms accepts just MultiDict for both string and list variants.
+    The string-variant is just specified as one key-value pair, while lists have
+    multiple key-value pairs with multiple different values, for example
+    [("foo", "val1"), ("foo", "val2")].
+
+    We typically want to work with string representation, because it is easier
+    to store in DB.  So the task of this custom Field class is to transform the
+    possible list representation into a XYZ-separated string.
+    """
+
+    # boilerplate stuff for the web-UI rendering
+    widget = wtforms.widgets.TextInput()
+    def _value(self):
+        return ', '.join(self.data or [])
+
+    def process_formdata(self, valuelist: list):
+        """
+        Fill the self.data string appropriately
+        """
+        # pylint: disable=attribute-defined-outside-init
+        self.data = self.transform(valuelist)
+
+    def transform(self, valuelist: list) -> str:
+        """
+        Transform the input array into a string of preferred format. Override.
+        """
+        raise NotImplementedError
+
+
+class CommaSeparatedListField(AbstractSeparatedListField):
+    """
+    Accept list of values, or (", ")-separated strings like "a, b, c".
+    """
+    def transform(self, valuelist: list) -> str:
+        """
+        Just do [a, b] => "a, b".  Note that we don't explicitly refuse inputs
+        like ["a, b", "c"], they should work "by accident".
+        """
+        return ", ".join(valuelist)
+
+
 class ChrootForm(BaseForm):
 
     """
@@ -1432,10 +1480,11 @@ class ChrootForm(BaseForm):
 
     comps = FileField("comps_xml")
 
-    module_toggle = wtforms.StringField("Modules",
-                                        validators=[ModuleEnableNameValidator()],
-                                        filters=[StringWhiteCharactersFilter()]
-                                        )
+    module_toggle = CommaSeparatedListField(
+        "Modules",
+        filters=[StringWhiteCharactersFilter()],
+        validators=[ModuleEnableNameValidator()],
+    )
 
     with_opts = wtforms.StringField("With options")
     without_opts = wtforms.StringField("Without options")
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_project_chroots.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_project_chroots.py
@@ -1,6 +1,6 @@
 import flask
 from coprs.views.misc import api_login_required
-from coprs.views.apiv3_ns import apiv3_ns
+from coprs.views.apiv3_ns import apiv3_ns, rename_fields_helper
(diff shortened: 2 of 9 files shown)


python/copr/v3/proxies/project_chroot.py
```diff
@@ -58,7 +58,9 @@ def edit(self, ownername, projectname, chrootname, additional_packages=None, add
         :param str chrootname:
         :param list additional_packages: buildroot packages for the chroot
         :param list additional_repos: buildroot additional additional_repos
-        :param list additional_modules: additional modules for the chroot
+        :param str additional_modules: comma separated list of modules that will
```
FrostyX, line 61: I would prefer to take a list of modules instead of a comma-separated string because it makes much more sense for the `python-copr` user. If there is a bug in the API and it in fact expects a comma-separated string, then we could send `",".join(additional_modules)`.

praiskup: I believed that I will make a small change, but this is a bit complicated topic (the forms don't work ideally with the uploaded json structures).

praiskup: I mean, I eventually did a different fix - PTAL

frontend/coprs_frontend/tests/test_apiv3/test_project_chroots.py
```diff
@@ -32,4 +34,31 @@ def test_edit_project_chroot(self):
         assert chroot.isolation == "nspawn"
         assert chroot.bootstrap == "off"
         assert chroot.buildroot_pkgs == "foo bar"
-        assert chroot.module_toggle == "!aaa:devel"
+        assert chroot.module_toggle == "!aaa:devel, blah:11"
+
+        data = {
+            # Submit additional modules as comma-separated string
+            "additional_modules": "module_b:11, !module_c:12",
```
praiskup, line 41: Note that both comma-separated string and list() works now (tested here).

frontend/coprs_frontend/coprs/views/apiv3_ns/json2form.py
```diff
@@ -2,7 +2,19 @@
...
+    """
+    Load flask request data from API/CLI calls (typically Python dict uploaded
+    as JSON data or set of uploaded files) and transform it to a dict in
+    WTForms-compatible format; so we eventually can leverage the WTForms
+    validation logic not only for web-UI but also for API.
+
+    Objects of type list() are joined using a separator into one string.
+    The separator string is by default set to one space character, but fields
+    configured in the comma_separated= array are separated using ', ' string.
+    """
```
praiskup, line 16: I was kind of shocked that someone is modifying the uploaded list of strings to space-separated string;  I was blaming WTForms, but then I realized we do -> so at least this comment.  But this is not a method I'd like :-(

praiskup:
The CI failure is fixed in #2468

frontend/coprs_frontend/coprs/views/apiv3_ns/json2form.py
TomasTomecek: the docstring is greatly appreciated, I was lost when I looked at code while we discussed this

FrostyX: Yeah, I am sorry about this. I know how awfully hacky this is.
You expressed my motivations perfectly in the comment. I just didn't know how to do this correctly. 

Maybe we could move the logic to the WTForms side? Looking at the documentation, they seem to be doing similar things with "custom fields"
https://wtforms.readthedocs.io/en/2.3.x/fields/#custom-fields

```python
class TagListField(Field):
    widget = TextInput()

    def _value(self):
        if self.data:
            return u', '.join(self.data)
        else:
            return u''

    def process_formdata(self, valuelist):
        if valuelist:
            self.data = [x.strip() for x in valuelist[0].split(',')]
        else:
            self.data = []
```

FrostyX (approved):
I tried to use this, and both comma-separated strings and a list of strings worked, thank you very much.

frontend/coprs_frontend/coprs/views/apiv3_ns/json2form.py
praiskup: I'll try to experiment a bit.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_project_chroots.py
```diff
@@ -52,17 +53,17 @@ def to_build_config_dict(project_chroot):
...
-    output = input.copy()
-    for from_name, to_name in replace.items():
-        if from_name not in output:
-            continue
-        output[to_name] = output.pop(from_name)
+    output = MultiDict()
+    for key in input_md.keys():
+        new_key = replace.get(key, key)
+        for value in input_md.getlist(key):
+            output.add(new_key, value)
```
praiskup, line 66: Note: This required a fix as it previously ate the inputs and shrinked multiple values into a one (the first one).

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -1432,10 +1471,11 @@ class ChrootForm(BaseForm):
...
     comps = FileField("comps_xml")
 
-    module_toggle = wtforms.StringField("Modules",
-                                        validators=[ModuleEnableNameValidator()],
-                                        filters=[StringWhiteCharactersFilter()]
-                                        )
+    module_toggle = CommaSeparatedListField(
+        "Modules",
+        filters=[StringWhiteCharactersFilter()],
+        validators=[ModuleEnableNameValidator()],
```
praiskup, line 1477: Note: Filters are applied first, so reordered to make this more readable...

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -1416,6 +1416,45 @@ def __init__(self, api_key=None, repo_url=None, *args, **kwargs):
...
+    multiple key-value pairs with multiple different values, for example
+    [("foo", "val1"), ("foo", "val2")].
+
+    We typically want to work with string representation, because it is easier
+    to store in DB.  So the task of this custom Field class is to transform the
+    possible list representation into a XYZ-separated string.
+    """
+    def process_formdata(self, valuelist):
+        # pylint: disable=attribute-defined-outside-init
+        self.data = self.transform(valuelist)
```
praiskup, line 1437: Simple as this?  @FrostyX it is maybe too simple compared to the example you provided, not sure if it covers all the cases.

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -1416,6 +1416,48 @@ def __init__(self, api_key=None, repo_url=None, *args, **kwargs):
             self.repo_url.data = repo_url
 
 
+class AbstractSeparatedListField(wtforms.Field):
```
praiskup, line 1419: Note: Going with abstract because re-using this for the space separated fields seems reasonable in the future.

PR #2469, opened 2023-01-12
frontend: customizable group link to FAS

FrostyX:
Other Copr instances might want to point somewhere else then FAS.


FrostyX:
See also https://pagure.io/rhcopr/maint/pull-request/9

praiskup (approved):

PR #2473, opened 2023-01-14
frontend: OpenAPI first steps

FrostyX:
See #1865

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/__init__.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/__init__.py
@@ -7,6 +7,7 @@ from functools import wraps
 from werkzeug.datastructures import ImmutableMultiDict, MultiDict
 from werkzeug.exceptions import HTTPException, NotFound, GatewayTimeout
 from sqlalchemy.orm.attributes import InstrumentedAttribute
+from flask_restx import Api, Namespace, Resource
 from coprs import app
 from coprs.exceptions import (
     AccessRestricted,
@@ -21,6 +22,30 @@ from coprs.helpers import streamed_json
 
 
 apiv3_ns = flask.Blueprint("apiv3_ns", __name__, url_prefix="/api_3")
+
+
+# Somewhere between flask-restx 1.0.3 and 1.1.0 this change was introduced:
+# > Initializing the Api object always registers the root endpoint / even if
+# > the Swagger UI path is changed. If you wish to use the root endpoint /
+# > for other purposes, you must register it before initializing the Api object.
+# TODO That needs to get fixed and then we can move this route to apiv3_general
+# See https://github.com/python-restx/flask-restx/issues/452
+@apiv3_ns.route("/")
+def home():
+    """
+    APIv3 homepage
+    Return generic information about Copr API
+    """
+    return flask.jsonify({"version": 3})
+
+
+api = Api(
+    app=apiv3_ns,
+    version="v3",
+    title="Copr API",
+    description="See python client - <https://python-copr.readthedocs.io>",
+    doc="/docs",
+)
 
 
 # HTTP methods
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
@@ -1,15 +1,24 @@
+# All documentation is to be written on method-level because then it is
+# recognized by flask-restx and rendered in Swagger
+# pylint: disable=missing-class-docstring
+
 import os
 import flask
 from sqlalchemy.orm import joinedload
 
 from werkzeug.datastructures import MultiDict
 from werkzeug.utils import secure_filename
+from flask_restx import Namespace, Resource
 
 from copr_common.enums import StatusEnum
 from coprs import db, forms, models
 from coprs.exceptions import (BadRequest, AccessRestricted)
 from coprs.views.misc import api_login_required
-from coprs.views.apiv3_ns import apiv3_ns, rename_fields_helper
+from coprs.views.apiv3_ns import apiv3_ns, api, rename_fields_helper
+from coprs.views.apiv3_ns.schema import (
+    build_model,
+    get_build_params,
+)
 from coprs.logic.complex_logic import ComplexLogic
 from coprs.logic.builds_logic import BuildsLogic
 
@@ -26,6 +35,12 @@ from . import (
     DELETE,
 )
 from .json2form import get_form_compatible_data
+
+
+
+
+apiv3_builds_ns = Namespace("build", description="Builds")
+api.add_namespace(apiv3_builds_ns)
 
 
 def to_dict(build):
@@ -76,11 +91,24 @@ def render_build(build):
     return flask.jsonify(to_dict(build))
 
 
-@apiv3_ns.route("/build/<int:build_id>/", methods=GET)
-@apiv3_ns.route("/build/<int:build_id>", methods=GET)
-def get_build(build_id):
-    build = ComplexLogic.get_build_safe(build_id)
-    return render_build(build)
+@apiv3_builds_ns.route("/<int:build_id>")
+class GetBuild(Resource):
+
+    @apiv3_builds_ns.doc(params=get_build_params)
+    @apiv3_builds_ns.marshal_with(build_model)
+    def get(self, build_id):
+        """
+        Get a build
+        Get details for a single Copr build.
+        """
+        build = ComplexLogic.get_build_safe(build_id)
+        result = to_dict(build)
+
+        # Workaround - `marshal_with` needs the input `build_id` to be present
+        # in the returned dict. Don't worry, it won't get to the end user, it
+        # will be stripped away.
+        result["build_id"] = result["id"]
+        return result
 
 
 @apiv3_ns.route("/build/list/", methods=GET)
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_packages.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_packages.py
@@ -1,4 +1,9 @@
+# All documentation is to be written on method-level because then it is
+# recognized by flask-restx and rendered in Swagger
+# pylint: disable=missing-class-docstring
+
 import flask
+from flask_restx import Namespace, Resource
 
 from coprs.exceptions import (
         BadRequest,
@@ -12,7 +17,15 @@ from coprs.exceptions import (
 )
 from coprs.views.misc import api_login_required
 from coprs import db, models, forms, helpers
-from coprs.views.apiv3_ns import apiv3_ns, rename_fields_helper
+from coprs.views.apiv3_ns import apiv3_ns, api, rename_fields_helper
+from coprs.views.apiv3_ns.schema import (
+    package_model,
+    add_package_params,
+    edit_package_params,
+    get_package_parser,
+    add_package_parser,
+    edit_package_parser,
+)
 from coprs.logic.packages_logic import PackagesLogic
 
 # @TODO if we need to do this on several places, we should figure a better way to do it
@@ -23,6 +36,10 @@ from .json2form import get_form_compatible_data
 
 
 MAX_PACKAGES_WITHOUT_PAGINATION = 10000
+
+
+apiv3_packages_ns = Namespace("package", description="Packages")
+api.add_namespace(apiv3_packages_ns)
 
 
 def to_dict(package, with_latest_build=False, with_latest_succeeded_build=False):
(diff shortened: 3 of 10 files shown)


FrostyX:
After our last discussion about API documentation, I was looking forward to experimenting with all the available packages.

In this PR, I wanted to show https://apiflask.com/
It has some quirks but overall it seems quite intuitive.
However, installation requires some work. You need to install rawhide version `python-marshmallow` (you can do it from my Copr project [frostyx/apiflask](https://copr.fedorainfracloud.org/coprs/frostyx/apiflask/)). Then `pip3 install apiflask`.

You can access the pretty docs page at http://127.0.0.1:5000/docs and you can download the API spec at http://127.0.0.1:5000/openapi.json

This is just a proof-of-concept, if you prefer any other API package, we can throw this away.

FrostyX:
#### flask-restx

I also tried https://flask-restx.readthedocs.io/en/latest/index.html which looked the best at first sight, but it seems to really be only for REST APIs. I.e. it expects routes to be classes that implement `get`, `post`, `put`, etc methods. I couldn't find a way to add a route that is a function and specify its endpoint and HTTP method.

#### flask-openapi3

The https://luolingchun.github.io/flask-openapi3/ looks good but it seems to have less ways how to do things. For example, query parameters need to be defined as classes https://luolingchun.github.io/flask-openapi3/Tutorial/Request/#query . The `apiflask` package has more options how to do this.

#### flasgger

I haven't tried https://github.com/flasgger/flasgger yet. It seems to be really straightforward - maybe too much. There is no generation, one must manually write the specification either as dict, json or yaml.

praiskup:
Seems like flasgger has noticeable more stars on GH.  @jpopelka don't you remember why you've chosen flask-restyx?

praiskup:
Just for the record, I'm not a big fan of the marshmallow models (too many fixes needed over the years..).

FrostyX:
> Just for the record, I'm not a big fan of the marshmallow models (too many fixes needed over the years..).

If we don't like marshmallow schemas, then I think we have two options

- If we don't like writing them, we can probably generate the marshmallow schemas from our SQLAlchemy models - https://flask-marshmallow.readthedocs.io/en/latest/#flask_marshmallow.sqla.SQLAlchemyAutoSchema
- If we dislike the dependency, we can do the thing that Flasgger does and manually write the YAML schemas - https://github.com/flasgger/flasgger#using-external-yaml-files

FTR, even the apiflask that I use in this PR AFAIK internally uses marshmallow. I think these are marshmallow

```python
from apiflask import Schema
from apiflask.fields import Integer, String, List, Nested
```

So we need to decide what exactly we don't like about marshmallow and how much we don't like it :D

praiskup:
Writing them is easy, but I got a feeling it has an unstable API.  The
schema validators were chronically updated with newer marshmallow
versions.

Initially I thought we could live with something like python's built-in
typing, or something non-third party.  But if that's not possible, writing
the models is not the problem and I think we have to live with that.  As I
read your research -> openAPI depends on marshmallow, so the huge userbase
must mean that the marshmallow eventually gets stabilized too.

BTW. the marshmallow in Fedora is maintained by us, and I was so glad we
removed our dep on it with APIv2 drop.

jpopelka:
> @jpopelka don't you remember why you've chosen flask-restyx for the Packit project?

I think I just searched for a library to generate the API documentation (Swagger), found an article about (now unmaintained) [Flask-RestPlus](https://github.com/noirbizarre/flask-restplus) (from which [Flask-RESTX](https://github.com/python-restx/flask-restx) was later forked due to inactivity), tried it, it did what we needed, done.

> Seems like flasgger has noticeable more stars on GH.

yup, but the difference is IMHO not so big
* [flasgger](https://github.com/flasgger/flasgger) - 3.2k stars
* [Flask-RestPlus](https://github.com/noirbizarre/flask-restplus) - 2.7k stars (forked to [Flask-RESTX](https://github.com/python-restx/flask-restx) - 1.7k stars)

praiskup:
Finally on a review, sorry for the delay.  This looks very nice, thank you!

Some general notes to the `/api_3/docs` route:
- can we use our own favicon?
- the page claims APIv3 v1.0 -> would it make sense to claim v3.0?

I'd prefer to have our main documentation pointing at the API documentation directly.  Perhaps at least in the `Do you have an API` FAQ entry?  (I mean without the misleading hop to the `/api/` page).

frontend/coprs_frontend/coprs/views/apiv3_ns/__init__.py
```diff
@@ -23,6 +24,29 @@
 apiv3_ns = flask.Blueprint("apiv3_ns", __name__, url_prefix="/api_3")
 
 
+# Somewhere between flask-restx 1.0.3 and 1.1.0 this change was introduced:
+# > Initializing the Api object always registers the root endpoint / even if
+# > the Swagger UI path is changed. If you wish to use the root endpoint /
+# > for other purposes, you must register it before initializing the Api object.
+# TODO That needs to get fixed and then we can move this route to apiv3_general
```
praiskup, line 31: There's never a good time to do this.  Can we have an issue for this (we should create a new api end point at least, and twist our library to use it).

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
```diff
@@ -76,11 +91,24 @@ def render_build(build):
...
+        """
+        Get a build
+        Get details for a single Copr build.
+        """
+        build = ComplexLogic.get_build_safe(build_id)
+        result = to_dict(build)
+
+        # Workaround - `marshal_with` needs the input `build_id` to be present
+        # in the returned dict. Don't worry, it won't get to the end user, it
+        # will be stripped away.
```
praiskup, line 109: Don't you have some restx or other reference link?

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
```diff
@@ -0,0 +1,560 @@
...
+    """
+    kwargs = field.__dict__.copy()
+    return field.__class__(**kwargs)
+
+
+def optional(field):
+    """
+    Unless this is an error on my side, there is a bug in flask-restx displaying
+    all URL path parameters inside the Swagger documentation as as required.
+    This is a temporary hack until it gets fixed.
```
praiskup, line 414: Has this been reported?

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
```diff
@@ -0,0 +1,560 @@
...
+    parser = RequestParser()
+    for arg in args:
+        arg.location = "json"
+        parser.add_argument(arg)
+    return parser
+
+
+def edit_package_parser():
+    # pylint: disable=missing-function-docstring
+    parser = add_package_parser().copy()
```
praiskup, line 543: Is this a deep copy?

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_packages.py
```diff
@@ -138,30 +169,57 @@ def get_package_list(ownername, projectname, with_latest_build=False,
...
+    parser = edit_package_parser()
+
+    @api_login_required
+    @apiv3_packages_ns.doc(params=edit_package_params)
+    @apiv3_packages_ns.expect(parser)
+    @apiv3_packages_ns.marshal_with(package_model)
+    def post(self, ownername, projectname, package_name, source_type_text=None):
+        """
+        Edit a package
+        Edit an existing package within a Copr project.
```
praiskup, line 206: duplicate doc line

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
```diff
@@ -0,0 +1,560 @@
...
+    example="copr",
+)
+
+pypi_package_version_field = String(
+    description="PyPI package version",
+    example="1.128pre",
+)
+
+# TODO We are copy-pasting descriptions from web UI to this file. This field
+# is an ideal candidate for figuring out how to share the descriptions
```
praiskup, line 134: /me nods

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
```diff
@@ -0,0 +1,560 @@
...
+    "state": state_field,
+    "submitted_on": submitted_on_field,
+    "submitter": submitter_field,
+}
+
+build_model = api.model("Build", build_schema)
+
+package_builds_schema = {
+    "latest": Nested(build_model, allow_null=True),
+    "latest_succeeded": Nested(build_model, allow_null=True),
```
praiskup, line 377: We miss ["builds" field](https://github.com/fedora-copr/copr/blob/a0bee3f1fae88c638c9d27082572d3d7e14fce82/frontend/coprs_frontend/coprs/models.py#L928-L929) here.

praiskup:
It is a pity we have poor coverage for these routes; so I'm a bit afraid that we'll miss some of the fields or so.  But it is worth the risk I'd say!  I like the change.

@jpopelka would you mind taking a 5-minute look?  No big review, just check whether the steps we do here are meaningful, or if there isn't an obvious mistake somewhere.

jpopelka (commented):
Looks OK to me :+1:

FrostyX:
> can we use our own favicon?

I don't think so, the favicon path seems to be "hardcoded"
https://github.com/frostyx/flask-restx/blob/docstring-details/doc/conf.py#L182-L185

Packit uses the default favicon as well

> the page claims APIv3 v1.0 -> would it make sense to claim v3.0?

Sure

> I'd prefer to have our main documentation pointing at the API documentation directly. Perhaps at least in the Do you have an API FAQ entry? (I mean without the misleading hop to the /api/ page).

Let's do this in a separate PR, we want to show the documentation but we also want to show users how to generate their API keys.

FrostyX:
> Looks OK to me +1

Thank you very much for taking a look

> It is a pity we have poor coverage for these routes; so I'm a bit afraid that we'll miss some of the fields or so.

Same here, but at least in the future we won't have to be afraid of regressions like missing fields.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
FrostyX: Honestly, I couldn't find any. This traceback happens when you remove the workaround:

```
werkzeug.routing.exceptions.BuildError: Could not build url for endpoint 'apiv3_ns.build_get_build' with values ['chroots', 'ended_on', 'id', 'is_background', 'ownername', 'project_dirname', 'projectname', 'repo_url', 'source_package', 'started_on', 'state', 'submitted_on', 'submitter']. Did you forget to specify values ['build_id']?
```

It doesn't make any sense to me, but adding the `build_id` key fixed the traceback. Of course, the `build_id` doesn't appear in the result because it is not defined in the schema, so it gets filtered out.

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
FrostyX: I would say, yes.

> `parser_copy.args = deepcopy(self.args)`

https://flask-restx.readthedocs.io/en/latest/_modules/flask_restx/reqparse.html#RequestParser.copy

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_packages.py
FrostyX: This was kind of intentional.

Screenshot_2023-04-27_23-16-48

The first docstring line is used for the "card" header and therefore it is visible even when the route is collapsed. The rest of the docstring is displayed once the route is expanded.

So I basically wanted to write the docstring like

```python
"""
Edit a package
Edit a package within a Copr project.
"""
```

but there is a bug in the docstring parsing - https://github.com/python-restx/flask-restx/pull/536

Anyway, we can write the docstrings in any format you want. My preference would be at least a two-line docstring so that we have at least some text when the route is expanded.

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
FrostyX: I don't think the `with_all_builds` option is exposed to the API

```
Your branch is up to date with 'origin/main'
$ git grep with_all_builds frontend/
frontend/coprs_frontend/coprs/models.py:    def to_dict(self, with_latest_build=False, with_latest_succeeded_build=False, with_all_builds=False):
frontend/coprs_frontend/coprs/models.py:        if with_all_builds:
```

frontend/coprs_frontend/coprs/views/apiv3_ns/__init__.py
FrostyX: I linked the issue in the code comment

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
FrostyX: As it turned out, it was an error on my side. Updated.

FrostyX:
I added some changes. Since it is a big PR, I kept them as separate commits for easier review. I will rebase and squash them before merging.

frontend/coprs_frontend/coprs/views/apiv3_ns/schema.py
praiskup: Indeed
```
copr --debug get-package --name python-backoff --with-all-builds @copr/copr-dev 
#  Debug log enabled  #
[10:48:13] {/usr/lib/python3.11/site-packages/urllib3/connectionpool.py:228} DEBUG - Starting new HTTP connection (1): localhost:55555
[10:48:13] {/usr/lib/python3.11/site-packages/urllib3/connectionpool.py:456} DEBUG - http://localhost:55555 "GET /api_3/package?ownername=%40copr&projectname=copr-dev&packagename=python-backoff&with_latest_build=False&with_latest_succeeded_build=False HTTP/1.1" 200 458
[10:48:13] {/usr/lib/python3.11/site-packages/urllib3/connectionpool.py:228} DEBUG - Starting new HTTP connection (1): localhost:55555
[10:48:13] {/usr/lib/python3.11/site-packages/urllib3/connectionpool.py:456} DEBUG - http://localhost:55555 "GET /api_3/build/list?ownername=%40copr&projectname=copr-dev&packagename=python-backoff HTTP/1.1" 200 809
```

praiskup:
+1, let's rebase and merge - so we can continue on making it even more awesome

praiskup:
@nikromen / @xsuchy do you want to review folks?

PR #2476, opened 2023-01-16
frontend: fix batches race logic by adding locks

### LINKED ISSUE #2107: First couple of --after-build-id builds remain pending
Original issue: https://pagure.io/copr/copr/issue/2107
Opened: 2022-03-03 11:56:06

This has happened to me for third time now.

 1. I submit a copr build and note down the ID
 2. I submit several (20+) copr builds with --after-build-id from (1) and --background
 3. While build from (1) runs, all other builds from (2) are pending
 4. Once (1) finishes, the tasks go starting, running... etc usually pretty much in build ID order, except...

The first 3-5 builds remain pending. Presumably forever.

See here. The very bottom build of python-hypothesis is the build from (1).

 
[png](/copr/copr/issue/raw/files/afad2b7ddc1afa4b3becde8a41033f6f457fc65ef103379cc86d105bc69b201a-png)

---

#### [churchyard](https://accounts.fedoraproject.org/user/churchyard) commented at 2022-03-03 12:57:53:
Previously, I've canceled them and started them again. THis time, I leave them pending for investigation. https://copr.fedorainfracloud.org/coprs/churchyard/hypothesis-6.39.1/builds/

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-03-03 19:40:57:
Interesting, thank you Miro for the report.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-03-03 20:47:51:
The reason is that batch 28829 is "empty" (no assigned build).  Seems like you removed the build(s) inside that batch?  We need to handle this case but I'm curious how this happened?

---

I haven't removed, deleted nor canceled any builds in this copr project.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-03-03 21:07:06:
I think I see the issue then.

> 1. I submit a copr build and note down the ID
> 2. I submit several (20+) copr builds with --after-build-id from (1) and --background

This creates 20+ batches.  And the batch create is logic is racy (bug) ...
https://pagure.io/copr/copr/blob/5e4efa97c1a1db5fb8a42ce7f4fc5413e0767b7a/f/frontend/coprs_frontend/coprs/logic/batches_logic.py#_17-23

I suppose you create all those buidls in 2) by some script.  So the first N builds race against each other and all of them create a new branch for the build from 1).  But eventually, each build can be assigned only to a single batch - so the batches become orphaned.

This deserves fixes on multiple places, please allow us some time :-)

Otherwise, I think you want something like:
1. submit the FIRST build
2. submit SECOND build --after-build-id FIRST
3. submit the rest of the builds --with-build-id SECOND

This way you create just two batches with the same semantics?

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-03-03 21:07:37:
Work-around is to put some "sleep" command after the first build is submitted.

---

Yes, by `parallel copr build-package $COPR --nowait --background --after-build-id 3597090  --name -- <pkg1> <pkg2> ...`

---

#### [churchyard](https://accounts.fedoraproject.org/user/churchyard) commented at 2022-04-19 21:12:54:
https://copr.fedorainfracloud.org/coprs/churchyard/hypothesis-6.39.1/ interestingly appears to be a "temporary project: will be deleted ASAP" forever. May I kill the builds and let it be deleted?

FrostyX:
Fix #2107

The said reproducer didn't work for me. I used this one:

    parallel copr build-package @copr/foo --nowait --background \
        --after-build-id 123  --name \
        ::: $(for i in {1..50}; do echo "copr-cli"; done)

### DIFF
--- a/frontend/copr-frontend.spec
+++ b/frontend/copr-frontend.spec
@@ -113,6 +113,7 @@ BuildRequires: python3-redis
 BuildRequires: python3-requests
 BuildRequires: python3-sphinx
 BuildRequires: python3-sphinxcontrib-httpdomain
+BuildRequires: python3-sqlalchemy-utils
 BuildRequires: python3-whoosh
 BuildRequires: python3-wtforms >= 2.2.1
 BuildRequires: python3-ldap
@@ -170,6 +171,7 @@ Requires: python3-pylibravatar
 Requires: python3-pytz
 Requires: python3-redis
 Requires: python3-requests
+Requires: python3-sqlalchemy-utils
 Requires: python3-templated-dictionary
 Requires: python3-wtforms >= 2.2.1
 Requires: python3-zmq
--- a/frontend/copr-frontend.spec
+++ b/frontend/copr-frontend.spec
@@ -117,6 +117,7 @@ BuildRequires: python3-whoosh
 BuildRequires: python3-wtforms >= 2.2.1
 BuildRequires: python3-ldap
 BuildRequires: python3-yaml
+BuildRequires: python3-backoff
 BuildRequires: redis
 BuildRequires: modulemd-tools >= 0.6
 %endif
@@ -173,6 +174,7 @@ Requires: python3-templated-dictionary
 Requires: python3-wtforms >= 2.2.1
 Requires: python3-zmq
 Requires: python3-ldap
+Requires: python3-backoff
 Requires: xstatic-bootstrap-scss-common
 Requires: xstatic-datatables-common
 Requires: js-jquery-ui
--- a/frontend/coprs_frontend/coprs/logic/batches_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/batches_logic.py
@@ -2,10 +2,12 @@
 Methods for working with build Batches.
 """
 
+import contextlib
 import anytree
 import backoff
 
 from sqlalchemy.exc import SQLAlchemyError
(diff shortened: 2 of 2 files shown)


FrostyX:
I know almost nothing about database locking, so you might want to fact-check me.

praiskup:
With locking, I'm only afraid of deadlocks ... the lock should be released as soon as possible (because the locked batch table is read-only queried by most of the web-ui pages).

FrostyX:
I narrowed the locking as much as possible, PTAL.
Also, I remember @praiskup, that you wanted me to remove some other lock or something that we have elsewhere in the code and it doesn't work but I cannot find it. Do you remember what did you mean?

praiskup:
Thank you for the update!

> you wanted me to remove some other lock or something 

I'm not sure what I meant, I don't even remember asking you in particular, when?? :-).  I think this removed lock was discussed some time ago 79baf273758960241dc74398776c440dd9f0fe12 ?

FrostyX:
I removed the `with_for_update()`, PTAL.

FrostyX:
I tested it with your commit and it works as expected.

praiskup (approved):

praiskup:
I'm just going to rebase against master so we have a green CI.  Then we can merge on standup.

PR #2477, opened 2023-01-16
common: periodically dump the priority queue to a JSON file

FrostyX:
Many times in the past I wanted to see how exactly the priority queue looks like at a given time but we didn't have a way of doing it. I wanted to implement some script for displaying it, but that's not possible because we AFAIK keep all the priority queue state in a variable (which of course is not accessible from the outside). So what about periodically dumping the queue into a JSON file by the `Dispatcher` itself?

I think this would be good enough for all the times that I needed it in the past but in case it was not sufficient, this would allow us to implement any kind of script to visualize the queue any way we want.

### DIFF
--- a/common/copr_common/dispatcher.py
+++ b/common/copr_common/dispatcher.py
@@ -2,7 +2,9 @@
 Abstract class Dispatcher for Build/Action dispatchers.
 """
 
+import os
 import time
+import json
 import logging
 import multiprocessing
 from setproctitle import setproctitle
@@ -116,6 +118,8 @@ class Dispatcher(multiprocessing.Process):
                 was_running = worker_manager.cancel_task_id(task_id)
                 self.report_canceled_task_id(task_id, was_running)
 
+            self.dump_queue(worker_manager)
+
             # process the tasks
             self._update_process_title("processing tasks")
             worker_manager.run(timeout=timeout)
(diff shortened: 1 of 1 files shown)


FrostyX:
If you dislike the idea in general, we can drop the PR. No hard feelings :D

common/copr_common/dispatcher.py
```diff
@@ -126,3 +129,20 @@ def run(self):
 
         # reset the title
         self._update_process_title()
+
+    def dump_queue(self, worker_manager):
+        """
+        Dump the current priority queue state into a JSON file
```
xsuchy, line 135: Can you put in the docstring the reason why we are doing this? So your future yourself do not worry it is dead code.

praiskup: It is not a dead code actually ... it is dumped in every single iteration, which raises my worries.  Could we have an opt-in knob, and document there when we need it?

common/copr_common/dispatcher.py
```diff
@@ -126,3 +129,20 @@ def run(self):
...
+        queue = []
+        for entry in worker_manager.tasks.prio_queue:
+            # This should never happen but let's make sure we don't traceback
+            if not isinstance(entry, list) or len(entry) != 3:
+                return
+
+            priority, count, task = entry
+            queue.append([priority, count, task.id])
+
+        path = "/tmp/copr-{0}-queue.json".format(self.name)
```
praiskup, line 146: Not sure if `/tmp` is the ideal place, perhaps /var/log?

FrostyX:
Updated, PTAL

common/copr_common/dispatcher.py
```diff
@@ -126,3 +130,29 @@ def run(self):
...
         self._update_process_title()
+
+    def dump_queue(self, worker_manager):
+        """
+        Dump the current priority queue state into a JSON file.
+        This is useful when we want to see the exact state of the queue from
+        backend's perspective (e.g. when some tasks are not being processed).
+        """
+        # This function is usefull for debugging, there is no need to run it
+        # all the time
```
praiskup, line 141: s/usefull/useful

praiskup (approved):

PR #2480, opened 2023-01-17
frontend: sort logs in build detail

xsuchy:
so they are in the same order and do not confuse users
addressing: https://social.linux.pizza/@knurd42/109702895855625653


FrostyX (approved):
+1

praiskup (approved):
We want to sort like this `[builder-live.log, backend.log, import.log]` (from the most frequently observed file to least).  We are just lucky `sort()` helps :-).

praiskup (commented):
Ok, taking back.  It doesn't sort this way.  Would you mind replacing `sort()` with something that sorts this way?

praiskup:
That sounds good!

PR #2484, opened 2023-01-17
frontend: WTForms validate function now expect additional parameter

### LINKED ISSUE #2482: CI fails for Fedora rawhide
In Rawhide (F38) there is  `python3-flask-wtf-1.1.0-1.fc38.noarch` which causes our tests to fail. Downgrading to `python3-flask-wtf-1.0.1-3.fc38.noarch` fixes it.

We should see what changes happened in the new version and update our code.

FrostyX:
Fix #2482
See https://github.com/wtforms/flask-wtf/pull/479

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -735,8 +735,8 @@ class CoprFormFactory(object):
             def selected_chroots(self):
                 return self.chroots.data
 
-            def validate(self):
-                if not super(F, self).validate():
+            def validate(self, extra_validators=None):
+                if not super(F, self).validate(extra_validators=extra_validators):
                     return False
 
                 if not self.validate_mock_chroots_not_empty():
@@ -1178,7 +1178,7 @@ class PackageFormDistGitSimple(BasePackageForm):
                                           self.package_name.data,
                                           self.namespace.data)
 
-    def validate(self):
+    def validate(self, _extra_validators=None):
         """
         Try to check that we can generate clone_url from distgit, namespace and
         package.  This can not be done by single-field-context validator.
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/forms.py
```diff
@@ -735,8 +735,8 @@ class F(CoprForm):
             def selected_chroots(self):
                 return self.chroots.data
 
-            def validate(self):
-                if not super(F, self).validate():
+            def validate(self, extra_validators=None):
+                if not super(F, self).validate(extra_validators):
```
praiskup, line 739: Considering the old WTForms (our current deployment where the argument doesn't exist, yet) -> is it safe to pass None here?  I'm afraid it is some other argument actually.

praiskup:
Nice, thank you for quickly fixing this!

frontend/coprs_frontend/coprs/forms.py
FrostyX: I know what you mean, I was a bit concerned as well. But I just checked

The `validate` method is not defined in `python3-flask-wtf`. The class hierarchy is funny.

    F < CoprForm < BaseForm < FlaskForm < Form < BaseForm
    These are ours          | flask-wtf |     wtforms

The `validate` method is defined in `BaseForm` in `python3-wtforms`, and it already takes `extra_validators`.

```
[root@copr-fe ~][PROD]# cat /usr/lib/python3.11/site-packages/wtforms/form.py |grep "def validate"
    def validate(self, extra_validators=None):
    def validate(self, extra_validators=None):
```

praiskup: so can we do "extra_validators=extra_validators" instead, just to be sure?

FrostyX: Good idea, done.

praiskup (approved):

Issue #2486, opened 2023-01-18
add additional properties to copr messaging schema


ryanlerch:
the COPR fedora-messaging schema needs a handful of additional properties added to make the messages work perfectly with the upcoming release of the new FMN. These are:

* agent_name: https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.agent_name so this would be the name of the person that started the action (e.g. build)
* usernames: https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.usernames
* groups: https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.groups
* app_name: https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.app_name

FMN also needs **artifacts** properties ((packages)](https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.packages), (flatpacks)](https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.flatpaks), (containers)](https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.containers), and (modules)](https://fedora-messaging.readthedocs.io/en/stable/api.html#fedora_messaging.message.Message.modules)) but these are typically only to relate to Fedora "artifacts" -- i.e. something tracked in src.fp.o, so i dont think we need these in COPR

Note that FMN also needs the **summary** and **__str__** properties, but these already appear to be implemented.

FrostyX:
Thank you @ryanlerch, we will take a look into that.
I appreciate the documentation links.

ryanlerch:
> Thank you @ryanlerch, we will take a look into that. I appreciate the documentation links.

i'm working on the app_name one now, FWIW -- a PR should be in soon

the others are a little more difficult for me without knowing heaps about COPR internals

ryanlerch:
also, if it helps, we have couple other PRs in flight on other apps that implement some of these:

* https://github.com/fedora-infra/anitya-messages/pull/53
* https://github.com/fedora-infra/bodhi/pull/4951
* https://github.com/fedora-infra/noggin-messages/pull/290
* https://pagure.io/elections-messages/pull-request/1

praiskup:
Thank you for the report Ryan!

> the COPR fedora-messaging schema needs a handful of additional properties
[...]

PR #2487, opened 2023-01-18
[messaging] Add app_name property to _CoprMessage base class

ryanlerch:
Related: #2486

Signed-off-by: Ryan Lerch <rlerch@redhat.com>

### DIFF
--- a/messaging/copr_messaging/private/hierarchy.py
+++ b/messaging/copr_messaging/private/hierarchy.py
@@ -39,6 +39,16 @@ class _CoprMessage(message.Message):
     def _str_prefix(self):
         return "Copr Message"
 
+    @property
+    def app_name(self):
+        """
+        Return the name of the application that generated the message.
+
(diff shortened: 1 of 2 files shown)


messaging/copr_messaging/private/hierarchy.py
```diff
@@ -39,6 +39,16 @@ def __str__(self):
...
 
+    @property
+    def app_name(self):
+        """
+        Return the name of the application that generated the message.
+
+        Returns:
+            the name of the application (copr)
+        """
+        return "copr"
```
abompard, line 50: This is supposed to be human-readable, so you can capitalize it.

praiskup: @ryanlerch can you please update?

praiskup: I eventually updated.

PR #2492, opened 2023-01-20
backend: fix chroot version parsing with new python-packaging

FrostyX:
Fix RHBZ 2162436

### DIFF
--- a/backend/copr_backend/sign.py
+++ b/backend/copr_backend/sign.py
@@ -103,6 +103,7 @@ def gpg_hashtype_for_chroot(chroot, opts):
     Given the chroot name (in "mock format", like "fedora-rawhide-x86_64")
     return the expected GPG hash type.
     """
+    # pylint: disable=too-many-return-statements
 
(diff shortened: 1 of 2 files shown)


backend/copr_backend/sign.py
```diff
@@ -116,6 +117,8 @@ def gpg_hashtype_for_chroot(chroot, opts):
         # epel-8+ and not _all_ the package data on backend.
         if parts[0] in el_chroots:
             el_version = version_part
+            if el_version in ["rawhide"]:
+                return "sha1"
```
praiskup, line 121: This method is likely wrong.  While working on the code, I thought sha256 signatures are not supported on EL6, but they are.  I think that if rhel-rawhide was re-created, we should mark it sha256.

FrostyX: Okay, I will fix the test case then.

FrostyX:
Updated

praiskup (approved):

Issue #2416, opened 2022-12-02
Webhooks return 400


FrostyX:
(Custom) Webhooks seem to have some issues after the last release. Originally reported on IRC

> <xsnrg>  Are there any known problems with copr webhooks, or recent
>    changes I might not be aware of?
> <xsnrg>  I have been getting 400s recently on webhooks that have been
 >   working for a very long time

Looking at the logs, I can see 

```
2022-12-02 16:03:05,527 [ERROR][/usr/share/copr/coprs_frontend/coprs/error_handlers.py:42|error_handlers:handle_error][ANON] Response error: 400 The browser (or proxy) sent a request that this server could not understand.
2022-12-02 16:03:05,532 [INFO][/usr/share/copr/coprs_frontend/coprs/views/webhooks_ns/__init__.py:25|__init__:after_request][ANON] Webhook (400) data: b''
```

praiskup:
I don't think this is related to our code change, but to some flask change.  I think that json mime header in the POST request is newly needed.

praiskup:
Please retry with `curl -H "Content-Type: application/json"`.

xsnrg:
I tried with -H "Content-Type: application/json", which failed, so I also tried using curl -X POST -json which should add both the Content-Type header as well as the accept header. This also did not work.

praiskup:
Nah, thank you for the feedback.  I just tested now:
```
$ curl -H "Content-Type: application/json" --data '{}' -X POST https://copr.fedorainfracloud.org/webhooks/custom/5893/<UUID_HASH>/copr-cli/
5079464
```
The [build has started](https://copr.fedorainfracloud.org/coprs/g/copr/copr/build/5079464/).

xsnrg:
Confirmed, that works for me as well. I guess what is needed is a web page update for the examples here: https://docs.pagure.org/copr.copr/user_documentation.html#webhooks

PR #2499, opened 2023-01-24
frontend: allow uploading artificial payload in custom webhook

praiskup:
It turns out non-json payload was never actually handed-over to a custom script.  This was caused by the artificial request.json check in the HookContentStorage.__init__ method.  But with the previous Flask versions we were at at least able to successfully submit the custom webhook build even with an empty payload without specified content type.

After the migration to F37 with new Flask the check for request.json started raising the exception 400 for empty payloads (causing #2416):

    Did not attempt to load JSON data because the request Content-Type
    was not 'application/json'

The JSON check is now replaced with request.get_data() which is a safe call in any case (form submitted, json data, uploaded file).

Fixes: #2416

### DIFF
--- a/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
+++ b/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
@@ -223,14 +223,14 @@ class HookContentStorage(object):
     tmp = None
 
     def __init__(self):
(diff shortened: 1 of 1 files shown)


FrostyX:
Hmm, the previous code was quite strange anyway, ... validating one variable and then using a different one.
The change looks good to me.

PR #2500, opened 2023-01-24
keygen: fix validation for projects with dash in their name

FrostyX:
Beaker tests revealed some bugs in the keygen email validation. It
failed for projects with dashes and projects with dots in their
names.

The culprit was wrongly placed dash character in the regex, it needs
to be at the beginning otherwise it is interpreted as a range.


FrostyX:
Release blocker ... merging.

PR #2510, opened 2023-01-25
frontend: fix the "Modules:" form widget

praiskup:
The value is actually a string taken from database, there's no need to use ", ".join().


praiskup:
The _value() is IMO used only by the web-ui widget.  But API should [be tested](https://github.com/fedora-copr/copr/blob/14aab1a37882f5456f723b1e156c700cae30f38d/frontend/coprs_frontend/tests/test_apiv3/test_project_chroots.py).

PR #2513, opened 2023-01-25
doc: add 2023-01-25 release notes

FrostyX:
https://docs.pagure.org/copr.copr/release-notes/2023-01-25.html


FrostyX:
Let's try something different this time. Instead of using an etherpad for the draft, please commit your changes directly to this branch. I think you should even be able to edit the file in the web browser in the "Files changed" section, if you don't want to pull locally. Once we are done, I will rebase the branch and squash all commits together.

We will see if it's better or worse than etherpad and do the next release notes according to that :-)

praiskup (approved):

PR #2516, opened 2023-01-26
Anitya dont rebuild already running jobs

praiskup:


frontend/coprs_frontend/run/check_for_anitya_version_updates.py
```diff
@@ -80,12 +80,14 @@ def get_updated_packages(updates_messages, backend):
     for message in updates_messages:
         update = message['msg']
         project = update['project']
+        projectname = project['name'].lower()
```
FrostyX, line 67: Not sure if we want to convert package names to lowercase but AFAIK PyPI packages are not case-sensitive so converting them or not converting them should both cause no problems.

I have no preference, I only noticed it

frontend/coprs_frontend/run/check_for_anitya_version_updates.py
```diff
@@ -157,11 +157,15 @@ def is_prerelease(version: str) -> bool:
     """
     Detect a pre-release version string.
```
FrostyX, line 159: I think the function name `is_prerelease` is fine but can you please update the docstring so it is clear that it filters both pre-release and post-release versions?

FrostyX (commented):
Otherwise LGTM.
I can see there is a lot of tests and thank you for that.

frontend/coprs_frontend/run/check_for_anitya_version_updates.py
praiskup: Yea, mee neither - this was decided before - I now just moved the position of this code.

frontend/coprs_frontend/run/check_for_anitya_version_updates.py
praiskup: I was thinking about it, whether "post" isn't actually just another "pre" release :-)

praiskup: Updated.

FrostyX (approved):
Thank you for the changes, LGTM

PR #2520, opened 2023-01-30
Define some copr-specific environment variables for custom method

FrostyX:
See #557

That RFE requested mainly package name, and additionally project name and owner name. I am adding only the last two. The package name will be hard to implement because we don't have its value in copr-rpmbuild. Moreover, we can't even send it from the frontend because when doing Packages > some package > Rebuild, we don't save the package <-> build relationship to the database. It is saved only after the SRPM is built.

### DIFF
--- a/doc/custom_source_method.rst
+++ b/doc/custom_source_method.rst
@@ -85,6 +85,19 @@ directory. We cannot process SRC.RPM. Because some chroots can use technology
 which our server cannot recognize. E.g., in the past `rpm` changed compression and
 checksum algorithm and rpm from RHEL was unable to process packages from Fedora.
 
+Environment variables
+---------------------
+
+On top of the standard environment variables, Copr defines also:
+
+- ``COPR_OWNER`` - Owner of the project. It may be a user or group name
+- ``COPR_PROJECT`` - Name of the project
+- ``COPR_PACKAGE`` - Name of the package that is currently being
+  built. It may not be known, in such case, this variable contains an
+  empty string.
+- ``COPR_RESULTDIR`` - See the optional parameter **resultdir**
+
+
 Examples
 --------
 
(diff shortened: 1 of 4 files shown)


FrostyX:
> when doing Packages > some package > Rebuild, we don't save the package <-> build relationship to the database. It is saved only after the SRPM is built.

In case this is a feature, not a bug, we might have to close the RFE as a WONTFIX.

praiskup:
Thank you for the PR!

> Moreover, we can't even send it from the frontend because when doing
> Packages > some package > Rebuild, we don't save the package <-> build
> relationship to the database.

The custom method is mostly triggered using webhooks or API, rather than
using web UI.  And then the `package <-> build` relationship should be
known since beginning.  We could simply set empty `PACKAGE_NAME=` if
package name is unknown?

Moreover, in most of the cases I believe we could fix even the web-ui
- when doing package rebuild, we could cookie-remember the package users
  want to rebuild and specify (or prefill) the value in the form?
- Custom build (not custom rebuild) could also accept a package name?

But having empty PACKAGE_NAME= if uknown seems OK to me.

FrostyX:
> But having empty PACKAGE_NAME="" if uknown seems OK to me.

Updated. And I also changed the commit message.

praiskup:
Can we have a documentation fix for this?  Otherwise looks good.

FrostyX:
> Can we have a documentation fix for this?

Good idea, added.

doc/custom_source_method.rst
```diff
@@ -85,6 +85,19 @@ directory. We cannot process SRC.RPM. Because some chroots can use technology
...
+
+On top of the standard environment variables, Copr defines also:
+
+- ``COPR_OWNER`` - Owner of the project. It may be a user or group name
+- ``COPR_PROJECT`` - Name of the project
+- ``COPR_PACKAGE`` - Name of the package that is currently being
+  built. It may not be known, in such case, this variable contains an
+  empty string.
+- ``COPR_RESULTDIR`` - See the optional parameter **resultdir**
+
```
praiskup, line 99: Awesome, thanks!

praiskup (approved):

PR #2524, opened 2023-02-05
Make sure user can build in the project before uploading SRPM

### LINKED ISSUE #134: cli: avoid uploading SRPM if the build can not be started
Original issue: https://pagure.io/copr/copr/issue/134
Opened: 2017-09-15 12:51:08

Receently, there's been added cool feature which requires user to be authenticated before the copr-cli tool attempts to start uploading large SRPM files.

It would be nice if the copr-cli refused to upload srpm file if:
- the copr project doesn't exist (typo)
- the user has no rights to build into the project
- ...

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2017-09-15 14:53:00:
Related to: https://pagure.io/copr/copr/pull-request/126

### LINKED ISSUE #137: [PR] Repo hit statitistics collection fixed
Original PR: https://pagure.io/copr/copr/pull-request/137
Opened: 2017-09-19 11:26:30
Opened by: clime

FrostyX:
Fix #134

Tested for both non-existing project and someone else's project that I don't have permissions to.

We can easily drop the `self.client.base_proxy.auth_check()` from CLI because there is `@requires_api_auth` decorator on the function.

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -314,6 +314,13 @@ class Commands(object):
 
         username, projectname, project_dirname = self.parse_dirname(args.copr_repo)
 
+        # Before we start uploading potentially large source RPM file, make sure
+        # that the user can actually build in the project
+        if not self.client.project_proxy.can_build_in(username, projectname):
+            msg = ("User '{0}' is not allowed to build in '{1}/{2}'"
+                   .format(self.username, username, projectname))
+            raise CoprRequestException(msg)
+
         builds = []
         for pkg in args.pkgs:
             if os.path.exists(pkg):
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -316,7 +316,9 @@ class Commands(object):
 
         # Before we start uploading potentially large source RPM file, make sure
         # that the user can actually build in the project
-        if not self.client.project_proxy.can_build_in(username, projectname):
+        can_build = self.client.project_proxy.can_build_in(
+            self.username, username, projectname)
+        if not can_build:
             msg = ("User '{0}' is not allowed to build in '{1}/{2}'"
                    .format(self.username, username, projectname))
             raise CoprRequestException(msg)
--- a/cli/tests/test_cli.py
+++ b/cli/tests/test_cli.py
@@ -577,11 +577,14 @@ def test_create_multilib_project(config_from_file, project_proxy_add, capsys):
     assert stdout == "New project was successfully created: http://copr/coprs/jdoe/foo/\n"
 
 
+@mock.patch('copr.v3.proxies.project.ProjectProxy.can_build_in',
+            return_value=True)
 @mock.patch('copr.v3.proxies.BaseProxy.auth_check', return_value=Munch(name="test"))
 @mock.patch('copr.v3.proxies.build.BuildProxy.create_from_url')
 @mock.patch('copr_cli.main.config_from_file', return_value=mock_config)
 @mock.patch('copr_cli.main.Commands._watch_builds')
-def test_create_build_no_wait_ok(watch_builds, config_from_file, create_from_url, auth_check, capsys):
+def test_create_build_no_wait_ok(watch_builds, config_from_file,
+                                 create_from_url, auth_check, _can_build_in, capsys):
     create_from_url.return_value = Munch(projectname="foo", id=123)
 
     main.main(argv=[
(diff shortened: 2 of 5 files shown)


frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
```diff
@@ -11,6 +11,20 @@
 from . import GET, PUT, editable_copr, get_copr
 
 
+@apiv3_ns.route("/project/permissions/can_build_in/<ownername>/<projectname>")
```
praiskup, line 14: Perhaps a small thing: When I see `can_build_in`, I don't know who.  From the method name I'd expect a username as an argument.  Should we have a better name?

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
```diff
@@ -11,6 +11,20 @@
...
+@api_login_required
+def can_build_in(ownername, projectname):
+    copr = get_copr(ownername, projectname)
+    if not flask.g.user.can_build_in(copr):
+        raise BadRequest("User '{0}' is not allowed to build in '{1}'"
+                         .format(flask.g.user.name, copr.full_name))
+
+    # I don't really know what to return. Perhaps we can start with an empty
+    # dict and add information as we need it? For now, it's important that it
+    # returns 200 status code.
```
praiskup, line 24: The call has "assert"-like, which is a bit uncommon in APIs.  We also return BadRequest (error 400).
I'm curious if there's some "guidance" in some API best-practices documentation...?

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
FrostyX: Interesting perspective. I thought about it as "Can I (as a logged user) build in <ownername>/<projectname>". But you are right, why limitting this to _myself_ only. 

Maybe we can remove the `@api_login_required` decorator making this route available for everyone and instead, adding a `username` parameter?

FrostyX: What I want to change is the behavior of `ProjectProxy.can_build_in` method. Either, I want to rename it to `assert_can_build_in` changing it to return a boolean.

But I will take a look what should we return in the API call as well.

FrostyX:
I addressed the assert-like call

> Perhaps a small thing: When I see can_build_in, I don't know who. From the method name I'd expect a username as an argument.

I added another parameter for specifying who is asking. I kind of like both more and less at the same time, so it is in a separate commit. Should I use it or delete it?

> Should we have a better name?

Fine with me but I couldn't think of any better

cli/copr_cli/main.py
```diff
@@ -308,12 +308,15 @@ def action_build(self, args):
         :param args: argparse arguments provided by the user
         """
 
-        # Before we start uploading potentially large source RPM file, make sure
-        # that the user actually has a valid credentials.
-        self.client.base_proxy.auth_check()
```
praiskup, line 313: This removal has a small potential to cause trouble.  It actually did a real `/auth-check` call before - revealing that session expired (e.g. wiped-out redis DB because frontend restarted).

Keeping this has an obvious consequence - additional API call for each build request.

cli/copr_cli/main.py
```diff
@@ -308,12 +308,15 @@ def action_build(self, args):
...
-
         username, projectname, project_dirname = self.parse_dirname(args.copr_repo)
 
+        # Before we start uploading potentially large source RPM file, make sure
+        # that the user can actually build in the project
+        if not self.client.project_proxy.can_build_in(username, projectname):
+            msg = ("User '{0}' is not allowed to build in '{1}/{2}'"
+                   .format(self.username, username, projectname))
+            raise CoprRequestException(msg)
```
praiskup, line 320: Awesome, thank you!  KISS way of catching a user's mistake.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
```diff
@@ -11,6 +11,14 @@
 from . import GET, PUT, editable_copr, get_copr
 
 
+@apiv3_ns.route("/project/permissions/can_build_in/<ownername>/<projectname>")
+@api_login_required
+def can_build_in(ownername, projectname):
```
praiskup: missing doc string

cli/tests/test_cli.py
```diff
@@ -577,11 +577,13 @@ def test_create_multilib_project(config_from_file, project_proxy_add, capsys):
     assert stdout == "New project was successfully created: http://copr/coprs/jdoe/foo/\n"
 
 
-@mock.patch('copr.v3.proxies.BaseProxy.auth_check', return_value=Munch(name="test"))
+@mock.patch('copr.v3.proxies.project.ProjectProxy.can_build_in',
+            return_value=Munch(name="test"))
```
praiskup, line 581: `Munch(name="test")` as `can_build_in` raises questions :)

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
FrostyX: Something weird happened and you commented on an outdated code. The docstring is there.

cli/copr_cli/main.py
FrostyX: Good catch. 
There are two commits in this PR. This issue can be easily solved by removing the second commit. We can do it that way or any other way you prefer.

FrostyX:
Updated according to what we discussed on the meeting.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_permissions.py
```diff
@@ -5,12 +5,29 @@
...
+    """
+    user = UsersLogic.get(who).one()
+    copr = get_copr(ownername, projectname)
+    result = {
+        "who": who,
+        "ownername": ownername,
+        "projectname": projectname,
+        "can_build_in": user.can_build_in(copr),
+    }
+    return flask.jsonify(result)
```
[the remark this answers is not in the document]
praiskup: Does not sqlalchemy sanitize this for us properly?

praiskup (approved):

PR #2526, opened 2023-02-06
frontend: hardlink in rawhide_to_release

praiskup:
This seems like a low-hanging fruit.  We never touch the generated RPMs, except when we want to re-sign them.  This happens only in accidental or exceptional situations, and those files would be re-signed multiple times (so we should be careful about race-conditions).

Fixes: #971


praiskup:
I'd like to hot-fix this in production, and use it while doing the rawhide-to-release today.
This should eventually save us from 1.5T to 2T of data.

praiskup:
Jakub hot-fixed this in production, and rawhide was forked into 38.  Merging.

PR #2527, opened 2023-02-06
Faster rawhide to release 2

praiskup:


praiskup:
This finally runs relatively fast (tested with `create index praiskup_test_4 on build_chroot (mock_chroot_id, status, build_id);`).

But it doesn't look wise to query BuildChroots for each project ... :-/ 
`

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1350,9 +1350,12 @@ def new(cls, build, mock_chroot, **kwargs):
         Each freshly created instance of BuildChroot has to be assigned to
         pre-existing Build and MockChroot, hence the mandatory arguments.
         """
-        copr_chroot = coprs_logic.CoprChrootsLogic.get_by_mock_chroot_id(
-            build.copr, mock_chroot.id
-        ).one()
+        copr_chroot = kwargs.pop("copr_chroot", None)
```
FrostyX, line 1353: It probably doesn't matter, but `.pop` will modify the dict, and I think it is not the intention here. Maybe use `.get("copr_chroot")` instead?

praiskup: We use the **kwargs later, so I'm afraid that `copr_chroot` - if not popped - would be specified twice

Issue #2523, opened 2023-02-02
Weird trend in copr-backend storage consumption


praiskup:
See the stats:
https://copr-be.cloud.fedoraproject.org/stats/distro.html

We increased storage consumption a lot within the last four weeks.  We need to understand why, and break the trend.
This might block a Rawhide to F38 branching task.

![Screenshot_20230202_125639](https://user-images.githubusercontent.com/3277839/216318405-c6f5745c-9eb6-47eb-9be6-db39f22f6336.png)

praiskup:
The storage consumption is now alarming IMO.  2.8T remains

FrostyX:
2.5T remains, and nagios started reporting warnings.

praiskup:
https://nagios.fedoraproject.org/nagios/cgi-bin//extinfo.cgi?type=2&host=copr-be.aws.fedoraproject.org&service=Disk+space+on+Copr+Backend+used+for+build+results

praiskup:
This seems to be related to the weird [increase](https://copr-be.cloud.fedoraproject.org/stats/projects.html) of [xsnrg/indi-3rdparty-bleeding](https://copr.fedorainfracloud.org/coprs/xsnrg/indi-3rdparty-bleeding/).

@xsnrg can you please take a look why your project consumes now +2T compared to the previous month?
Also, is it possible to start packaging differently - e.g. unbundling?  I'm not sure why it eats so much.
TBH there are not so many packages, but they are built relatively frequently and they are large.

```
[root@copr-be fedora-rawhide-x86_64][PROD]# du --max-depth=1 -h | grep -- -aok$
313M    ./05459020-indi-aok
312M    ./05288200-indi-aok
410M    ./05522800-indi-aok
313M    ./05459190-indi-aok
313M    ./05459132-indi-aok
313M    ./05458852-indi-aok
313M    ./05496928-indi-aok
313M    ./05508305-indi-aok
313M    ./05351754-indi-aok
313M    ./05458908-indi-aok
313M    ./05459076-indi-aok
410M    ./05522931-indi-aok
410M    ./05522613-indi-aok
410M    ./05522865-indi-aok
410M    ./05518992-indi-aok
312M    ./05290128-indi-aok
312M    ./05290285-indi-aok
313M    ./05394952-indi-aok
312M    ./05290214-indi-aok
313M    ./05502834-indi-aok
313M    ./05373270-indi-aok
313M    ./05362774-indi-aok
[...]

PR #2536, opened 2023-02-09
frontend: restrict forking of large projects

### LINKED ISSUE #984: disallow forking of too-large copr projects
Original issue: https://pagure.io/copr/copr/issue/984
Opened: 2019-09-05 08:38:48

Say the project has several chroorts enabled, and each chroot has few tousands of packages;  such fork action is going to create a huge amount of traffic on backend (huge I/O) and it is mostly useless duplication of storage.  We should IMO raise exception saying **hey, this project is too large to fork, please add it as "external repository"**.  WDYT?

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2019-09-26 06:59:15:
We could have an admin-only checkbox "allow forking", and tell **it is forbidden to fork this repository, do ... instead** (internal copr case, there are large projects with low amount of packages which people fork unnecessarily).

FrostyX:
Fix #984

### DIFF
--- a/frontend/coprs_frontend/coprs/config.py
+++ b/frontend/coprs_frontend/coprs/config.py
@@ -160,6 +160,10 @@ class Config(object):
     # e.g. ou=users,dc=company,dc=com
     LDAP_SEARCH_STRING = None
 
+    # If a project has more package than this, normal users can't fork it.
+    # Only admins can.
+    FORK_PACKAGES_LIMIT = 1000
+
 
 class ProductionConfig(Config):
     DEBUG = False
--- a/frontend/coprs_frontend/coprs/logic/complex_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/complex_logic.py
@@ -38,6 +38,7 @@ class ComplexLogic(object):
     """
     Used for manipulation which affects multiply models
     """
+    # pylint: disable=too-many-public-methods
 
     @classmethod
     def delete_copr(cls, copr, admin_action=False):
@@ -87,6 +88,7 @@ class ComplexLogic(object):
 
     @classmethod
     def fork_copr(cls, copr, user, dstname, dstgroup=None):
+        cls.raise_if_cant_fork(user, copr)
         forking = ProjectForking(user, dstgroup)
         created = (not bool(forking.get(copr, dstname)))
         fcopr = forking.fork_copr(copr, dstname)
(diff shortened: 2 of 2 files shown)


FrostyX:
Hmm, I need to limit API as well.

frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
```diff
@@ -1100,6 +1101,16 @@ def render_copr_fork(copr, form, confirm=False):
 def copr_fork_post(copr):
     form = forms.CoprForkFormFactory.create_form_cls(copr=copr, user=flask.g.user, groups=flask.g.user.user_groups)()
     if form.validate_on_submit():
+        limit = app.config["FORK_PACKAGES_LIMIT"]
+        packages_count = PackagesLogic.get_all(copr.id).count()
```
praiskup, line 1105: I'm thinking about #2523.  Do you think we could have also a configuration value like "FORK_PROJECT_DENYLIST" with a manually maintained list of projects?

Speaking of convenience, it would be nice to have a `copr.is_forkable()` method
wrapping this (ideally redis-cached).

praiskup: Perhaps it is not a bad idea to add `copr.forks_disallowed` item into the DB table.
We could easily auto-enable/disable that field while we are analyzing the storage.

FrostyX:
I isolated the code into one function `raise_if_cant_fork` and called it from a place that is used by both web UI and API. And it even simplified the code, so win-win :-)

frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
FrostyX: > Speaking of convenience, it would be nice to have a copr.is_forkable() method wrapping this

100% agree, I moved the logic to a separate function, but it is `ComplexLogic.raise_if_cant_fork` which is more convenient. 

> I'm thinking about https://github.com/fedora-copr/copr/issues/2523. Do you think we could have also a configuration value like "FORK_PROJECT_DENYLIST" with a manually maintained list of projects?
> Perhaps it is not a bad idea to add copr.forks_disallowed item into the DB table.
We could easily auto-enable/disable that field while we are analyzing the storage.

This is something that I am not that thrilled about. The `FORK_PROJECT_DENYLIST` option can be implemented in a few minutes, so I have no problem with this.

The rest sounds like overkill to me.

Personally, I would prefer counting the size of backend data that is to be forked, and failing the action if it is too much. But that is not so user-friendly until we have #1108 implemented.

PR #2537, opened 2023-02-09
frontend, backend: remove libmodulemd1 dependency

FrostyX:
It is a leftover, we are using modulemd-tools for modularity things, so let it figure out the dependencies.


praiskup (approved):
thank you

praiskup (approved):

PR #2538, opened 2023-02-10
Don't branch armhfp chroots

FrostyX:
We got a report that we accidentally branched `fedora-38-armhfp` chroot. That's because we don't use the `active_only` parameter and therefore branch inactive `fedora-rawhide-armhfp`.


praiskup (approved):
thank you

PR #2542, opened 2023-02-15
keygen: fix config files installation pattern

praiskup:
copr-keygen.spec uses %defattr for copr-signer, and we mistakenly included the logrotate drop-in file there, new logrotate started to dislike this, per logrotate error:

    error: Ignoring copr-keygen because the file owner is wrong (should
    be root or user with uid 0).

The cron.daily script, if moved up a bit in the %files section doesn't need the %attr override either.

While on it, the sudoers dropin config shouldn't be '(noreplace)'.  We want to maintain/replace the file via RPM updates.


FrostyX:
LGTM

PR #2543, opened 2023-02-15
cli: simplify `mock-config` output

praiskup:
For Mock, config_opts['dnf.conf'] option is just an alias to the config_opts['yum.conf'] option.  Since 2020:
https://github.com/rpm-software-management/mock/commit/5d620ad326da0f656883f090c588bcfe36e92858

### DIFF
--- a/cli/copr_cli/build_config.py
+++ b/cli/copr_cli/build_config.py
@@ -41,7 +41,7 @@ config_opts['chroot_additional_packages'] = '
 {%- endif %}
 {%- if repos %}
 
-config_opts[config_opts['package_manager'] + '.conf'] += \"\"\"
+config_opts['dnf.conf'] += \"\"\"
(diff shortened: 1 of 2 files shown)


cli/copr_cli/build_config.py
```diff
@@ -41,7 +41,7 @@
 {%- endif %}
 {%- if repos %}
 
-config_opts[config_opts['package_manager'] + '.conf'] += \"\"\"
+config_opts['dnf.conf'] += \"\"\"
```
FrostyX, line 44: Maybe a dumb question, but we are still releasing `copr-cli` for EPEL7. Won't this cause problems?

praiskup: The dnf.conf is equivalent to config option yum.conf, so we should be fine even on el7, even building for el7.

PR #2544, opened 2023-02-16
replacing pagure with github in our projects after copr migration to GH

nikromen:
some unchanged leftovers from our migration


praiskup (approved):
Awesome, thank you!

PR #2545, opened 2023-02-16
python-copr: add option for follow_fedora_branching

### LINKED ISSUE #2312: Missing python-copr option support for the "follow_fedora_branching" feature
Original issue: https://pagure.io/copr/copr/issue/2312
Opened: 2022-09-13 14:40:06

We seem to have this in the API_V3 route on the frontend side, but it doesn't seem to be part of the client library, nor client.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-10-19 17:50:56:
Needed by: https://github.com/packit/packit-service/issues/1651

nikromen:
Fixes #2312


praiskup:
Looks good, but can we have this in `copr create` and `copr modify` too?

python/copr/v3/proxies/project.py
```diff
@@ -154,7 +156,7 @@ def edit(self, ownername, projectname, chroots=None, description=None, instructi
              contact=None, additional_repos=None, unlisted_on_hp=None, enable_net=None,
              auto_prune=None, use_bootstrap_container=None, devel_mode=None,
              delete_after_days=None, multilib=None, module_hotfixes=None,
-             bootstrap=None, bootstrap_image=None, isolation=None,
+             bootstrap=None, bootstrap_image=None, isolation=None, follow_fedora_branching=True,
```
FrostyX, line 159: I think this must be `follow_fedora_branching=True` otherwise running the edit action without specifying the `follow_fedora_branching` parameter, it would enable it for the project even if it was previously disabled.

FrostyX:
There is still an issue, take a look

```python
copr create test-follow-branching
# project settings say it is ON
copr modify test-follow-branching --description "foo" --follow-fedora-branching off
# project settings say it is OFF
copr modify test-follow-branching --description "foo"
# project settings say it is ON (and shouldn't be)
```

I think this is because of  `default="on"` for the `parser_modify`. The `--appstream` and `--fedora-review` parameters probably suffer the same bug.

nikromen:
thanks for pointing to it @FrostyX 

should be fixed now, PTAL

FrostyX:
LGTM, thank you for the updates.

PR #2547, opened 2023-02-16
frontend: allow automation (cron) to remove builds

praiskup:
Previously we somewhat "faked" the user who is requesting build removals in the 'manage.py clean-old-builds' cron task.  That was faked to the "build.copr.user" which - for the group projects - might not be part of the group anymore and the pedantic check from a4e7a5aa467cad6132eb9d00170 stops us.

Newly, for automation task, we can use "automation_user=str()" argument that will effectively disable the user.can_build_in() check.

While on it, tweak the logging and error reporting a bit so there's is more obvious what is going on.

Fixes: #2523

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -1144,26 +1144,36 @@ class BuildsLogic(object):
     @classmethod
     def check_build_to_delete(cls, user, build):
         """
-        :type user: models.User
+        :type user: models.User or models.AutomationUser
         :type build: models.Build
         """
-        if not user.can_edit(build.copr) or build.persistent:
+        if build.persistent:
             raise InsufficientRightsException(
-                "You are not allowed to delete build `{}`.".format(build.id))
+                f"User '{user.username}' can not delete a persistent "
+                f"build {build.id}")
+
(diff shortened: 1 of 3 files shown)


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1297,7 +1315,8 @@ def clean_old_builds(cls):
 
             for build in delete_builds:
                 try:
-                    cls.delete_build(build.copr.user, build)
+                    cls.delete_build(None, build,
+                                     automation_user="Admin: clean-old-builds")
```
praiskup, line 1319: Note: This is the actual fix that is needed to fix #2523.

FrostyX:
I don't have anything against the PR as is.

But, we are complicating the `check_build_to_delete` function with permission-checking logic. In case we need to avoid some permission checking also somewhere else, we will have to add the same `automation_user=None` and its logic there as well. I think I'd like something like this more

```python
class AutomationUser:
    def __init__(username):
        self.username = username

    def can_edit(self, _copr):
        return True


# and then usage
cls.delete_build(AutomationUser("Admin: clean-old-builds"), build)
```

But I am not forcing the idea.

praiskup:
Updated, PTAL

PR #2551, opened 2023-02-21
build_aux: Add hack for unittesting in container

nikromen:
this is just my personal hack that I am using to unittest in containers. (bcs I don't want to install many packages in my system that I won't use)


FrostyX:
I won't use this but if you find it useful, then why not.

nikromen:
thanks!

PR #2552, opened 2023-02-21
doc: add backup recovery page

FrostyX:
SSIA

### DIFF
--- a/doc/backup_recovery.rst
+++ b/doc/backup_recovery.rst
@@ -1,0 +1,201 @@
+.. _backup_recovery:
+
+Recovery from backups
+=====================
+
+In case of emergency ... break glass.
+
+
+Backend
+-------
+
+Don't rush, take your time. It will take 5 days to sync 20TB of data so it is
+not worth micro-optimizing tasks to save seconds. The ``rsync`` from
+storinator is expected to run 110 MB/s while the disk can handle 130
+MB/s. Our instance has 5 Gbps so the bottleneck is probably on the
+network between data centers.
+
+Prepare a new RAID array
+........................
+
+In case of a real disaster, you will probably do the recovery from
+the real production instance. In case of a simulated disaster
+(i.e. testing the backups), spawn a new instance::
+
+    $ git clone git@github.com:fedora-copr/ansible-fedora-copr.git
+    # Follow the README.md steps for preparation
+    $ ./run-playbook pb-backup-recovery-01.yml
+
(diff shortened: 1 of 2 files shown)


FrostyX:
See also https://github.com/fedora-copr/ansible-fedora-copr/pull/3

doc/backup_recovery.rst
```diff
@@ -0,0 +1,109 @@
...
+
+    mkfs.ext4 /dev/md0 -L copr-data
+    mkdir /mnt/data
+    mount /dev/disk/by-label/copr-data /mnt/data/
+
+
+SSH key shenanigans
+...................
+
+Generate a new SSH key::
```
nikromen, line 61: To make it more clear I would explicitly mention that we want to do it on our local machine... you know, as soon as it's not an exercise, people may do mistakes :D

```suggestion
Generate a new SSH key on your local machine::
```

nikromen (commented):
otherwise looks good to me. But I rather leave the approval to Pavel, because I reviewed it mainly from the clarity point of view, not the functionality point of view :)

doc/backup_recovery.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    cat /proc/mdstat
+
+You can see the raid details using::
+
+    mdadm --detail /dev/md0
+
+Format and mount::
+
+    mkfs.ext4 /dev/md0 -L copr-data
```
xsuchy, line 60: I would add:
tune2fs -m0 /dev/md0
By default it is 5% and that is HUGE amount of data on large volumes.

FrostyX: Updated

FrostyX:
I updated the `rsync` command to drop the `--delete` and other unwanted parameters. It finally finished without `malloc` errors.

To sync 20TB of data it took us 7677 minutes (5 days and some change). I haven't yet identified if the bottleneck is the network or disk, I'll need some help with that.

But I suppose we will be addressing the bottleneck in https://github.com/fedora-copr/ansible-fedora-copr/pull/3 , not here.

doc/backup_recovery.rst
```diff
@@ -0,0 +1,130 @@
+.. _backup_recovery:
+
+Recovery from backups
+=====================
+
+In case of emergency ... break glass.
```
praiskup, line 6: Can we have a rough estimate "what is going to happen now" here?

FrostyX: Updated

PR #2556, opened 2023-02-26
doc: adding a pulp instance

FrostyX:
I started working on PULP playbooks for Copr


FrostyX:
So, basically ... this is not finished and won't be finished for a while (who knows when we will have a production instance for pulp). We can either merge as is, fill in the blanks once we know how, and skip irrelevant steps during releases and upgrades; or we can close the PR because it would be hanging here for months.

Dropping the WIP tag now.

praiskup (approved):
Thank you

PR #2561, opened 2023-03-02
Update from downstream branch 'rawhide'

majamassarini:
Downstream commit: 807ade96

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,59 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+packit_instances: ["stg"]
+upstream_project_url: https://github.com/fedora-copr/copr.git
+
+packages:
+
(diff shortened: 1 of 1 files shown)


praiskup:
> This should be closed...

I'm ignoring the pull-request then;  but having a link to some docs about "what you are testing" would be good :-)
We are using tito for automatic spec file synchronisation (at release time), but this Packit feature might
be very interesting to learn anyway!

majamassarini:
> I'm ignoring the pull-request then; but having a link to some docs about "what you are testing" would be good :-) We are using tito for automatic spec file synchronisation (at release time), but this Packit feature might be very interesting to learn anyway!

Sure, we will let you know! At the moment the docs are just the ones related with the [research](https://github.com/packit/research/tree/main/monorepo-support).

PR #2570, opened 2023-03-09
backend: run createrepo without --database

### LINKED ISSUE #1171: createrepo for dnf systems can be run without --database
Original issue: https://pagure.io/copr/copr/issue/1171
Opened: 2019-12-20 10:31:19

Do we need the sqlite databases in Fedora repositories?   The only use-case
that comes to my mind is running `mock -r fedora-*-x86_64` on el7, then yum
needs to be used at least to initialize bootstrap chroot. 

---

#### [ngompa](https://accounts.fedoraproject.org/user/ngompa) commented at 2020-01-02 02:32:57:
You should probably just not produce SQLite databases unless you want to wire up something like repoview to all repos. They're not needed for YUM on EL7 either...

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2020-01-02 16:58:32:
Hm, seems like repoview could be "fixed" to generate the DB automatically, ...

So the sqlite database is only needed on el6?

---

#### [ngompa](https://accounts.fedoraproject.org/user/ngompa) commented at 2020-01-02 16:59:15:
Not even there. YUM has always supported rpm-md XML, as that's the primary format.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2020-01-02 17:03:14:
I'm lost then.  46dcceee91eff9bfecdab67cea90bc1675279d27 says that 
--database was added to generate database, but not why we needed that.

Any idea?  AFAIK the --database option  was not related to repoview,
I'm not aware we ever considered it.

---

#### [ngompa](https://accounts.fedoraproject.org/user/ngompa) commented at 2020-01-02 17:05:35:
We never needed it unless you wanted to do remote query things or use repoview. YUM has supported using rpm-md XML since slightly before RHEL 5. The format replaced the yumhdr metadata format in 2005. The SQLite version was introduced slightly later to provide "optimized" variants for super large repositories, but the "fast" resolution of SQLite repositories requires an extension for YUM that isn't always installed.

So anyway, in the end, you don't need it and you can just drop SQLite databases if you want.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2020-01-02 17:27:25:
Thanks!  This originally came from @dmach's idea.  Anything more to add?

---

#### [dmach](https://accounts.fedoraproject.org/user/dmach) commented at 2020-01-03 09:35:44:
@praiskup DNF doesn't use sqlite repodata. YUM3 uses it, but if it's not available it generates it locally from downloaded xml repodata. The sqlite repodata used to be provided to skip this step.

If you disable generating sqlite repodata, you'll save CPU cycles and also disk space on all Fedoras and RHEL8. Builds on RHEL <= 7 will take a bit longer because yum needs to generate the sqlite files from xml. The difference should be quite negligible because RHEL/CentOS/EPEL repos have the sqlite files and the COPR repos that will need the conversion are usually quite small.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2020-12-11 12:46:02:
I think we can have some dummy if for EL <= 7 chroots, and keep the sqlite DB
pre-compiled  there.  Just in case someone maintained rather larger repos
in copr.  Epel 7 is going to be a minor thing on backend in terms of storage
in near future.  Marking as RFE.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-08-09 17:14:29:
SEO: --no-database

nikromen:
dnf and yum system don't need to create sqlite database since dnf and yum no longer utilize them.

Although there are some systems that still require --database option. Yum3 used sqlite data so system like RHEL <= 7 and EPEL <= 7 still require --database option.

for more detailed reasoning of the --database exception please refer to: https://github.com/fedora-copr/copr/issues/1171

Fixes #1171

### DIFF
--- a/backend/run/copr-repo
+++ b/backend/run/copr-repo
@@ -21,13 +21,14 @@ import sys
 
 import filelock
 
+from copr_backend.constants import CHROOTS_USING_SQLITE_REPODATA
+from copr_backend.createrepo import BatchedCreaterepo
 from copr_backend.helpers import (
     BackendConfigReader,
     CommandException,
     run_cmd,
     get_redis_logger,
 )
-from copr_backend.createrepo import BatchedCreaterepo
 
 
 def printable_cmd(cmd):
@@ -127,8 +128,16 @@ def filter_existing(opts, subdirs):
     return new_subdirs
 
 
+def _database_option(chroot: str) -> str:
+    for os_family, old_versions in CHROOTS_USING_SQLITE_REPODATA.items():
+        if any(f"{os_family}-{old_version}" == chroot for old_version in old_versions):
+            return "--database"
+
+    return "--no-database"
+
+
 def run_createrepo(opts):
-    createrepo_cmd = ['/usr/bin/createrepo_c', opts.directory, '--database', '--ignore-lock',
+    createrepo_cmd = ['/usr/bin/createrepo_c', opts.directory, _database_option(opts.chroot), '--ignore-lock',
                       '--local-sqlite', '--cachedir', '/tmp/', '--workers', '8']
 
     if "epel-5" in opts.directory or "rhel-5" in opts.directory:
--- a/backend/copr_backend/constants.py
+++ b/backend/copr_backend/constants.py
@@ -1,5 +1,6 @@
 import os
 from logging import Formatter
+from typing import Dict, List
 
 mockchain = "/usr/bin/mockchain"
 # rsync path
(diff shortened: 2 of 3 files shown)


nikromen:
note that I haven't tested this on stg yet

backend/run/copr-repo
```diff
@@ -127,8 +127,22 @@ def filter_existing(opts, subdirs):
     return new_subdirs
 
 
+def _database_option(chroot: str) -> str:
```
praiskup, line 130: I don't think it makes much sense to keep dead code here.  Would you mind moving the `system_that_needs_db` into config.py (say `CHROOTS_USING_SQLITE_REPODATA`), documenting it, keeping that dict empty by default, and keeping this method used?

praiskup (approved):

PR #2578, opened 2023-03-14
Keep whoosh cache small

### LINKED ISSUE #2534: searching with whoosh in Fedora Copr is too slow
Follow-up from [the red hat bug 2168394](https://bugzilla.redhat.com/show_bug.cgi?id=2168394).

The Fedora Copr search is too slow.  The Python Whoosh library behind the
search box is probably slow by design for reasonably large databases and
[doesn't scale](https://github.com/whoosh-community/whoosh/issues/554) [well](https://groups.google.com/g/whoosh/c/QM_P8cGi4v4/m/pZv-zxkFBQAJ).

    $ du --max-depth=0 -h /var/lib/copr/data/whooshee/copr_whoosheer
    568M    /var/lib/copr/data/whooshee/copr_whoosheer

It seems reasonably big, even though we use just one whoosheer.  And when being
used, the process memory consumption goes to unreasonable sizes.  Single query
means ~3GB of eaten RAM - and the memory is not used wisely actually (multiple
requests can end up with >= 8GB and so on).

We can either consider optimizing the searching (smaller DB, dedicate a running daemon
with preloaded DB in memory, fix leaks etc.) or migrate somewhere else (our own
hosted lucene-based stuff, use the AWS cloudsearch service, revamp elasticsearch).

We should keep whoosh as opt-in, the instance we have in Red Hat is reasonably fast
with Whoosh (smaller DB).

FrostyX:
Fix #2534
See #2574

Please see the commit messages for more info.


frontend/coprs_frontend/coprs/whoosheers.py
```diff
@@ -54,13 +54,18 @@ def update_package(cls, writer, package):
 
     @classmethod
     def insert_copr(cls, writer, copr):
+        # Avoid indexing the largest projects to reduce the whooshee index size,
+        # and thus memory usage
+        packages=cls.get_package_names(copr)
+        if len(packages) > 10000:
+            packages = []
```
praiskup, line 61: How much does this help?  I'm not against it, just please document it.

praiskup:
Thanks, interesting.  This is good to have no matter what, but I'd bet it doesn't Fix #2534 unless we significantly decrease the memory consumption.

frontend/coprs_frontend/coprs/whoosheers.py
FrostyX: Documented in the commit description, on my laptop it went from 99M to 30M. Not sure if it is worth it but I'd say yes.

praiskup: Ah, thanks.  What makes me curious now is how **just the set of package names** can make the cache 3 times bigger?  Tested this:
```
coprdb=# select sum(ARRAY_LENGTH(REGEXP_SPLIT_TO_ARRAY(description || ' ' || instructions, '\s'), 1)) from copr;
┌─────────┐
│   sum   │
├─────────┤
│ 2502387 │
└─────────┘
(1 row)

coprdb=# select count(*) from package;
┌─────────┐
│  count  │
├─────────┤
│ 1619863 │
└─────────┘
(1 row)
```

xsuchy: This means that packages from PyPI and Rubygems project cannot be found. :(

FrostyX: I removed that commit, I think we don't need it.

FrostyX:
I wanted to remove the `update-indexes-quick` command but we call it from `@coprs_ns.route("/update_search_index/", methods=["POST"])` so instead I added `optimize=True` and kept the command.

I am not really sure how the is the searching going to behave in our day-to-day production but at this point, I think this is our best shot before moving away from whoosh.

PR #2581, opened 2023-03-20
fronted: follow the self.result_dir name instead of hardcoded id

### LINKED ISSUE #2567: Incorrect results dirs
Hi team, 
I notice a strange error:
- when i submit a build though cli with: `copr build-package  --name python-abimap test`，the builder log will live in `{buildid}-{package_name}/builder-live.log.tar.gz` dir
image
 but the log `url` in frontend page is just `{buildid}/builder-live.log.tar.gz` 
image
- when i rebuild directly through rebuild page, it works as expected
- when i tried the same command using the offical copr instance, it works as expected

It can be reproduced in latest `docker-compose` setup and have no idea why, any help appreciated!

nikromen:
Fixes #2567


frontend/coprs_frontend/tests/test_models.py
```diff
@@ -100,7 +100,7 @@ def test_build_logs(self):
 
         def _pfxd(basename):
             pfx = ("http://copr-be-dev.cloud.fedoraproject.org/results/"
-                   "user1/foocopr/srpm-builds/00000001")
+                   "user1/foocopr/srpm-builds/00000001-foopackage")
```
praiskup, line 103: For srpm-builds -> I think we want to keep the old pattern.  The only problem we had with the old logic is non-determinism (it picks one or the other pattern depending on if the package name is known in advance, while it should _always_ use the `00000001` pattern).

praiskup (approved):
Thank you!

PR #2582, opened 2023-03-20
cli: Add `get` command

### LINKED ISSUE #2557: Copr-cli variant for project_proxy.get()
We are missing a copr-cli alternative to get the project detail (we have just `copr list`).

nikromen:
Fix #2557

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -33,7 +33,7 @@ from copr.v3 import (
     CoprConfigException, CoprNoResultException, CoprAuthException,
 )
 from copr.v3.pagination import next_page
-from copr_cli.helpers import cli_use_output_format
+from copr_cli.helpers import cli_use_output_format, print_project_info
 from copr_cli.monitor import cli_monitor_parser
 from copr_cli.printers import cli_get_output_printer as get_printer
 from copr_cli.util import get_progress_callback, serializable
@@ -590,7 +590,6 @@ class Commands(object):
         build_config.rootdir = "{0}-{1}_{2}".format(ownername.replace("@", "group_"), projectname, args.chroot)
         print(MockProfile(build_config))
 
-
     def action_list(self, args):
         """ Method called when the 'list' action has been selected by the
         user.
@@ -605,15 +604,7 @@ class Commands(object):
             return
 
         for project in projects:
-            print("Name: {0}".format(project.name))
-            print("  Description: {0}".format(project.description))
-            if project.chroot_repos:
-                print("  Repo(s):")
-                for name, url in project.chroot_repos.items():
-                    print("    {0}: {1}".format(name, url))
-            if project.additional_repos:
-                print("  Additional repo: {0}".format(" ".join(project.additional_repos)))
-            print("")
+            print_project_info(project)
 
     def action_status(self, args):
         build = self.client.build_proxy.get(args.build_id)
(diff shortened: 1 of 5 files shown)


cli/copr_cli/main.py
```diff
@@ -590,6 +590,18 @@ def action_mock_config(self, args):
         build_config.rootdir = "{0}-{1}_{2}".format(ownername.replace("@", "group_"), projectname, args.chroot)
         print(MockProfile(build_config))
 
+    @staticmethod
+    def _print_project_info(project):
+        print(f"Name: {project.name}")
```
FrostyX, line 595: There are no f-strings on EPEL7, so this is what makes the CI fail.
You'll need to use `.format`

cli/tests/resources/get_project_response.json
```diff
@@ -0,0 +1,13 @@
...
+  },
+  "additional_repos": [
+    "http://copr-be.cloud.fedoraproject.org/results/msuchy/scl-utils/epel-6-$basearch/",
+    "http://copr-be.cloud.fedoraproject.org/results/rhscl/httpd24/epel-6-$basearch/",
+    "http://copr-be.cloud.fedoraproject.org/results/rhscl/v8314/epel-6-$basearch/"
+  ],
+  "description": "A recent stable release of Ruby with Rails 3.2.8 and a large collection of Ruby gems. This Software Collection gives developers on Red Hat Enterprise Linux 6 access to Ruby 1.9, which provides a number of new features and enhancements, including improved Unicode support, enhanced threading, and faster load times.",
+  "name": "ruby193",
+  "instructions": ""
+}
```
FrostyX, line 13: You can see the error sign here in the diff saying "No newline at the end of file".
AFAIK you use Vim, so I think you want to `set eol` option.

cli/copr_cli/main.py
```diff
@@ -590,6 +590,18 @@ def action_mock_config(self, args):
         build_config.rootdir = "{0}-{1}_{2}".format(ownername.replace("@", "group_"), projectname, args.chroot)
         print(MockProfile(build_config))
 
+    @staticmethod
```
FrostyX, line 593: We are making an instance of this class in the `main` function, so I think we can use a normal (not static) method. Or is there some reason why we need it to be static?

cli/copr_cli/main.py
```diff
@@ -1243,6 +1256,11 @@ def setup_parser():
     cli_use_output_format(parser_builds, default=None)
     parser_builds.set_defaults(func="action_list_builds")
 
+    parser_get = subparsers.add_parser("get", help="Get information about project")
+    parser_get.add_argument("owner", help="Owner of a package. Groups begins with `@` e.g. @copr")
+    parser_get.add_argument("project", help="Name of the project.")
```
FrostyX, line 1261: I think this will require us to call the command as `copr-cli get @copr copr` instead of `copr-cli get @copr/copr`? That would be inconsistent with other commands.

FrostyX (commented):
Thank you for the PR,
overall I like the change and big thanks for the tests.

I made some comments within the code.

cli/copr_cli/main.py
praiskup: +1, we want to keep EPEL7 compatibility in `python-copr` and `copr-cli`.

cli/tests/resources/get_project_response.json
nikromen: thanks for the tip!

cli/copr_cli/main.py
nikromen: Hmm I didn't think of calling it as `@copr/copr`... It's more intuitive this way

praiskup (approved):

PR #2584, opened 2023-03-20
cli: add --review parameter for download-build  

### LINKED ISSUE #2459: Limit download-build to fedora-review files
We have `copr download-build <id>` command to download all build results. It is now possible, to filter the results to download only RPM packages using `--rpms` parameter and/or spec file using the `--spec` parameter.

I would like to add a `--fedora-review` or `--review` option to download only the `fedora-review` results.

FrostyX:
Fix #2459


praiskup (approved):
Nice, thanks!

Issue #2589, opened 2023-03-23
Make copr_messaging optional


pkking:
In openEuler community, we do not shipped any message bus like fedmsg and so we can't get pkgs like fedora-messaging and fedmsg.
Is there any workaround to make these pkg optional?

praiskup:
There are ways to make package optional in Rpm (conditional deps, soft deps, etc.).
But there's a low motivation in the Copr team to make messaging optional (even if unused, it typically costs very small amount of storage to have the libraries installed).

pkking:
I'd like submit a PR to make backend work without copr_messaging :)

praiskup:
Sure, patches are welcome!  The "optional dependencies" is a non-trivial topic, so be prepared we'll review it very carefully. :)

praiskup:
See also: https://github.com/fedora-copr/copr/pull/2596

PR #2590, opened 2023-03-23
make copr_messaging optional

pkking:
#2589 My thoughts stay straightforward, just make the copr_messaging and fedmsg into weak dependency...

### DIFF
--- a/backend/copr_backend/msgbus.py
+++ b/backend/copr_backend/msgbus.py
@@ -7,7 +7,11 @@ import logging
 import copy
 import json
 
-from copr_messaging import schema
+try:
+    from copr_messaging import schema
+except ImportError:
+    # copr_messaging is optional
+    schema = None
 
 from .constants import BuildStatus
 
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -77,10 +77,10 @@ Requires:   openssh-clients
 Requires:   prunerepo >= %prunerepo_version
 Requires:   python3-copr
 Requires:   python3-copr-common >= %copr_common_version
-Requires:   python3-copr-messaging
+Recommends: python3-copr-messaging
 Requires:   python3-daemon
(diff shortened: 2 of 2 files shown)


praiskup:
@jmacku, any idea if we could avoid the initial `You have successfully added a new vcs-diff-lint configuration ...` message?  That message spams every single PR.

backend/copr_backend/msgbus.py
```diff
@@ -33,6 +39,8 @@ def message_from_worker_job(style, topic, job, who, ip, pid):
     once we can exepect that people are using copr-messaging module (and thus
     we can change message format without affecting the API).
     """
+    if not copr_messaging_enabled:
+        return None
```
praiskup, line 43: It looks like this is not needed, what is the traceback you face if you uninstall python-copr-messaging?

MessageSender.announce() calls `bus.announce_job()` only if some bus is configured (which shouldn't be your case) therefore neither `message_from_worker_job` shouldn't be called.

pkking: Yes, in today's codebase, there's no need to add these lines.
But i want to make this method more robust, maybe it was call by another function in future, and the caller may not know our assumptions.
The `schema` in the context is undefined and may cause an exception.

praiskup: This just feels weird, maybe the position in code, I'm not sure.

If things work for you just like that, can we just (and resolve the rest later?):
```
ry:
    from copr_messaging import schema
except ImportError:
    # copr_messaging is optional
    schema = None
```

jamacku:
@praiskup It's GitHub's "new" feature, but it is messy. It requires you to upload SARIF on the `push` event as well.

I was dealing with this issue in:
- https://github.com/redhat-plumbers-in-action/differential-shellcheck/issues/215

praiskup (approved):

PR #2591, opened 2023-03-23
Monorepo packit.yaml draft

majamassarini:
I opened this PR to work on packit-service monorepo support.
Do not merge!


.packit.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - python-copr.spec
+
+  cli:
+    downstream_package_name: copr-cli
+    upstream_package_name: copr-cli
+    paths: 
+      - ./cli
+    specfile_path: copr-cli.spec
+    files_to_sync: 
+      - copr-cli.spec
```
praiskup, line 24: Could we have `after: [ python ]` here, and perform `copr-cli` build after `python-copr` (that's good to resolve the dependencies).

.packit.yaml
```diff
@@ -0,0 +1,59 @@
+# See the documentation for more information:
+# https://packit.dev/docs/configuration/
+packit_instances: ["stg"]
+upstream_project_url: https://github.com/fedora-copr/copr.git
+
+packages:
+
+  python:
```
praiskup, line 8: `python:` seems acceptable, but does the spelling matter for automation?  Having `python-copr` here would seem nicer (it's the actual package name, if we speak about `packages:`).

.packit.yaml
```diff
@@ -0,0 +1,59 @@
...
+    downstream_package_name: python-copr
+    upstream_package_name: copr
+    paths: 
+      - ./python
```
praiskup, line 12: My feedback on this
- good this is an array, I suppose that change on these paths trigger the actual build, right
- being able to "deny-list" some sub-directories (somehow) would be nice

.packit.yaml
majamassarini: We have not yet a solution for this, at least I think (@lachmanfrantisek @csomh am I wrong?) .
But while working on it I will pay attention if any simple solution comes to my mind!

.packit.yaml
majamassarini: Sure, I will rename it!

.packit.yaml
majamassarini: Actually this is a list but only the first entry, at the moment, is used as root for building the sub-package. I am waiting for an example where a list is really needed, if you have one please let me know :)

Regarding the deny list, I might be wrong, but I think @csomh already has something for it.

.packit.yaml
praiskup: Note there's the `--after-build-id` feature in `copr-cli` in case the problem is how to "chain" the builds Copr side.

nikromen: I really like the approach with `after: [ ... ]` like @praiskup suggested. People will typically want to specify the order of builds in monorepos.

Similar approach uses docker-compose with `depends_on` field.

As pavel mentioned, you can use build batches to solve this.

csomh: @majamassarini is right, we haven't thought about this before. I'll put it on our list :) I would call it `build_after`, though, b/c that would be more explicit.

Issue #2566, opened 2023-03-06
Build stuck in clone loop if committish is ambiguous with filename


jn64:
I was building a package, where the spec file was in a subdir named `solfege`. The branch name was also `solfege`, which I entered as the "committish".

After the initial git clone, it tried to `git checkout solfege` which failed because its ambiguous:

```
Running: git checkout solfege

cmd: ['git', 'checkout', 'solfege']
cwd: /var/lib/copr-rpmbuild/workspace/workdir-q7kzlna9/fedora-spec
rc: 128
stdout: 
stderr: fatal: 'solfege' could be both a local file and a tracking branch.
Please use -- (and optionally --no-guess) to disambiguate
```

And then it proceed to try to clone again repeatedly.

Though git-checkout(1) says that it should default to treating the ambiguous argument as a branch name.

> ARGUMENT DISAMBIGUATION
>        When there is only one argument given and it is not -- (e.g. git
>        checkout abc), and when the argument is both a valid <tree-ish> (e.g. a
>        branch abc exists) and a valid <pathspec> (e.g. a file or a directory
>        whose name is "abc" exists), Git would usually ask you to disambiguate.
>        Because checking out a branch is so common an operation, however, git
>        checkout abc takes "abc" as a <tree-ish> in such a situation. Use git
>        checkout -- <pathspec> if you want to checkout these paths out of the
>        index.

I am not so familiar with Copr so maybe there is something different in the git version or configuration.

---

Full log below. I accidentally deleted the build, but I saved the `builder-live.log` that I had open at the time:

```
Warning: Permanently added '2620:52:3:1:dead:beef:cafe:c1cf' (ED25519) to the list of known hosts.
Running: /usr/bin/copr-rpmbuild --verbose --drop-resultdir --srpm --build-id 5594128 --detached
Version: 0.64
PID: 5383
Logging PID: 5384
Task:
{'appstream': True,
 'background': False,
 'build_id': 5594128,
 'chroot': None,
 'project_dirname': 'playground',
 'project_name': 'playground',
 'project_owner': 'jn64',
 'repos': [],
 'sandbox': 'jn64/playground--jn64',
 'source_json': {'clone_url': 'https://github.com/jn64/fedora-spec.git',
                 'committish': 'solfege',
                 'spec': 'solfege.spec',
                 'srpm_build_method': 'rpkg',
                 'subdirectory': 'solfege',
                 'type': 'git'},
 'source_type': 8,
 'submitter': 'jn64',
[...]

PR #2594, opened 2023-03-24
rpmbuild: use 'git switch', not 'git checkout'

praiskup:
There's no ambiguity when committish we are switching to matches both filename/direname or branch name.

Fixes: #2566


FrostyX:
The change from `git checkout` to `git switch` very much makes sense. But I am not sure what is the purpose of introducing `--detach`. It may be worth mentioning in the commit message.

praiskup:
Ok, updated.

nikromen (approved):

PR #2596, opened 2023-03-28
also remove fedmsg and copr_messaging from buildreq

pkking:
Fix: #2589


praiskup (approved):
Makes sense, thanks!  Small nit on commit message, please try to concentrate on why instead of what;  why you are doing this, why you think it is OK, why do you need that, etc.

praiskup:
Typically we also say `Relates: #2589` in the commit message.

PR #2600, opened 2023-03-30
dist-git: downgrade python-rpkg to version 1.65

nikromen:
new version of python-rpkg introduced unexpected behavior of commit
where it commits empty commit by default. This causes a bug if multiple
importing of a package happens in copr and the latter import of a
package is without any change, the following `commit` will make a new
commit hash because it allows empty commits by default.

I created an issue about it in rpkg. See
https://pagure.io/rpkg/issue/677

in the meantime the temporary solution of this is to downgrade
rpk package so we can release copr. Once this issue is resolved
remove the version specification from python3-rpkg package.

### DIFF
--- a/dist-git/tests/test_crazy_merging.py
+++ b/dist-git/tests/test_crazy_merging.py
@@ -1,13 +1,12 @@
 # coding: utf-8
 
 import tempfile
+from unittest import mock, skip
 import pytest
 import shutil
 import time
 import os
 import logging
-
-from unittest import mock
 
 from munch import Munch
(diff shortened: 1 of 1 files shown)


praiskup:
I'm OK with this (temporarily) but we desperately need to dump somewhere (in-code, git-commit-message or both) the bug number we are working-around.

nikromen:
Ok I figured it out. I updated the commit message and filed an issue https://pagure.io/fedpkg/issue/514

PR #2604, opened 2023-04-01
frontend: don't 500 for KRB5_USER_DENYLIST_PATTERNS=None

praiskup:
The default KRB5_USER_DENYLIST_PATTERNS=None is not iterable.

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
@@ -47,7 +47,7 @@ def krb_straighten_username(krb_remote_user):
     if not re.match(r"^[\w.-]+$", username):
         return None
 
-    for pattern in app.config.get("KRB5_USER_DENYLIST_PATTERNS", []):
+    for pattern in app.config.get("KRB5_USER_DENYLIST_PATTERNS") or []:
         if fnmatch(username, pattern):
             return None
     return username


frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
```diff
@@ -47,7 +47,7 @@ def krb_straighten_username(krb_remote_user):
     if not re.match(r"^[\w.-]+$", username):
         return None
 
-    for pattern in app.config.get("KRB5_USER_DENYLIST_PATTERNS", []):
+    for pattern in app.config.get("KRB5_USER_DENYLIST_PATTERNS") or []:
```
FrostyX, line 50: You could probably do `app.config.get("KRB5_USER_DENYLIST_PATTERNS", [])` but it's not even worth changing.

FrostyX (approved):

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
praiskup: Nah, that's what we had before..?

FrostyX: /me dumb 
I understand the problem now.

praiskup: np, thank you for the review!

Issue #2606, opened 2023-04-03
copr-frontend FTBFS (CI failure)


praiskup:
Every pull-request CI run against frontend gets failed:

```
ERROR tests/test_anitya.py::TestAnitya::test_pypi - AttributeError: create_sc...
ERROR tests/test_auth.py::TestGroupAuth::test_group_names_ldap - AttributeErr...
ERROR tests/test_comps.py::TestComps::test_edit_project_chroot_comps[api] - A...
ERROR tests/test_comps.py::TestComps::test_edit_project_chroot_comps[webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case0-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case0-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case1-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case1-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case2-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case2-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case3-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case3-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case4-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_valid_configuration_cases[case4-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case0-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case0-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case1-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case1-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case2-api]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case2-webui]
ERROR tests/test_config_overrides.py::TestConfigOverrides::test_isolation_override[case3-api]
...
    def setup_method(self, method):
[...]

PR #2607, opened 2023-04-03
frontend: test compat with flask-sqlalchemy 3+

praiskup:
Drop the @new_app_context, we are creating a new app context for each test-case, and "pushing" it to the stack.

Also, recording SQL queries is now not ON by default, therefore setting SQLALCHEMY_RECORD_QUERIES to True.

Fixes: #2606


praiskup:
Please hide white-space changes while reviewing (this is a rather simple change).

FrostyX (approved):

PR #2651, opened 2023-04-04
frontend: show API config even for not logged users

xsuchy:
Users are often confused and does not expect that when they log in this page change. Despite it is clearly written here. Let show the config and mask the values.

### DIFF
--- a/frontend/coprs_frontend/coprs/templates/api.html
+++ b/frontend/coprs_frontend/coprs/templates/api.html
@@ -27,8 +27,8 @@
     <p>The API token is valid for {{ config['API_TOKEN_EXPIRATION'] }} days after it has been generated.
     </p>
 
-    {% if g.user %}
     {% include 'additional_token_info.html' %}
+    {% if g.user %}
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/templates/api.html
```diff
@@ -45,6 +45,17 @@ <h2>API Token</h2>
         <input type="button" value="Generate a new token" />
     </a>
     {% else %}
+    <pre style="font-size:120%">
+[copr-cli]
+login = VALUE-IS-HIDDEN
```
praiskup, line 50: Nice idea, thanks!  Can you `s/VALUE-IS-HIDDEN/LOGIN_TO_UNHIDE/`?  I observed the page after the change, and the info that you have to log-in is pretty invisible now.

xsuchy: UNHIDE is not a word. I think. I used REVEAL. 
Updated.

Issue #2655, opened 2023-04-04
Git switch detach error


FrostyX:
I am seeing an error in our beaker tests. Reproducer:

```
$ copr-cli buildscm @copr/TEST1680648707856267158Project10 \
    --clone-url https://pagure.io/copr/copr-hello.git --commit rpkg-util
```

There seems to be some issue with 

```
stderr: Cloning into '/var/lib/copr-rpmbuild/workspace/workdir-m49llvcs/copr-hello'...

Running: git switch --detach rpkg-util

cmd: ['git', 'switch', '--detach', 'rpkg-util']
cwd: /var/lib/copr-rpmbuild/workspace/workdir-m49llvcs/copr-hello
rc: 128
stdout: 
stderr: fatal: '--detach' cannot be used with '-b/-B/--orphan'
```

praiskup:
Uhh, what a pain.  I'm not sure how to do this easily, can we tweak `checkout.guess` config and revert  26ea0dfaafb8d49607e53cefc92e442258b5c836?  Perhaps this issue relates to the "unstable" note in Git manual we discussed.

PR #2656, opened 2023-04-05
Revert "rpmbuild: use 'git switch', not 'git checkout'"

praiskup:
This causes other problems, see #2655.

This reverts commit 26ea0dfaafb8d49607e53cefc92e442258b5c836.


praiskup:
Not sure if this is needed, we should try the updated git first, I think (I can not reproduce the misbehaviour locally on my up2date box).

PR #2667, opened 2023-04-13
Derive all message classes from _CoprMessage

nphilipp:
Previously, `BuildChrootStartedV1StompDontUse` didn’t and thus failed to set `app_name`, which made FMN fall back to the toplevel namespace of the message topic (lowercase `copr`), which in turn caused it to use both variations both in the user interface and when filtering messages (which warrants a fix of its own, see fedora-infra/fmn#854).


praiskup:
LGTM, thank you!  Also thanks for the test.

PR #2668, opened 2023-04-13
Fix outdated deps

nphilipp:
Some of `requirements.txt` files listed dependencies which haven’t been maintained in a while, replace by what Copr actually uses.

### DIFF
--- a/backend/docs/requirements.txt
+++ b/backend/docs/requirements.txt
@@ -6,7 +6,7 @@ redis
 mock
 requests
 python-daemon
-bunch
+munch
 lockfile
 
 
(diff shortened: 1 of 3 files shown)


xsuchy (approved):
+1

praiskup:
Thank you for the PR, how has this affected you?  Can you actually drop the file instead?  (shouldn't be used at all).

nphilipp:
Heh, I wanted to setup a virtualenv to test #2667 and the `requirements.txt` files where the first thing I found to get to a state where I could run the test suite.

praiskup:
Nah, it is confusing.  The way to run the testsuite is to use `dnf builddep *spec`.  Can I close this PR, and provide a new one removing the requirements.txt?

PR #2671, opened 2023-04-15
rpmbuild: retry only git clone without checkouting

### LINKED ISSUE #2657: git clone: fatal: destination path 'XXX' already exists and is not an empty
we want to retry `git clone` in `git_clone_and_checkout()` in rpmbuild in case of network error. But if clone is done successfully and then `git checkout` in the same function fails, the backoff decorator will retry the function with `git clone` which leads to error because repo is already cloned.

nikromen:
We want to retry only git clone when exception happens because when `git clone succeeds but later `git checkout` fails then it will try to clone repo again which will fail in 100% times.

Fixes #2657


praiskup (approved):
Thank you!

PR #2676, opened 2023-04-19
frontend: add command for generating usage heatmap

### LINKED ISSUE #2673: Statistics of who is using a service
I want to have a heatmap statistics on who is using our service. I have this code
  https://gist.github.com/xsuchy/eb5d8394cafd428176138ac4fbb881e2
but this needs to be incorporated in copr code.
It will be great if the users could be grouped in groups (e.g. Python team, Cloud SIG,...) and this can be configured in a configuration file.

FrostyX:
Fix #2673

### DIFF
--- a/docker/frontend/files/etc/copr/copr.conf
+++ b/docker/frontend/files/etc/copr/copr.conf
@@ -136,3 +136,9 @@ MAX_BUILD_TIMEOUT = 108000
 
 # Disable template caching
 CACHE_TYPE = "NullCache"
+
+USAGE_TREEMAP_TEAMS = {
+    "CPT": ["@copr", "@mock", "msuchy", "praiskup", "frostyx", "nikromen"],
+    "Packit": ["packit", "packit-stg", "ttomecek"],
+    "Python": ["@python", "thrnciar"],
+}
--- a/frontend/conf/cron.daily/copr-frontend-optional
+++ b/frontend/conf/cron.daily/copr-frontend-optional
@@ -5,3 +5,4 @@
 #runuser -c 'copr-frontend notify-outdated-chroots' - copr-fe
 #runuser -c 'copr-frontend delete-outdated-chroots' - copr-fe
 #/usr/libexec/copr_dump_db.sh /var/lib/copr/data/db_dumps/
+#runuser -c 'copr-frontend usage-treemap -D /var/lib/copr/data/usage' - copr-fe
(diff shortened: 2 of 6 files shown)


FrostyX:
I started working on the heatmap. As it turns out `python3-pandas` and `python3-plotly` has a lot of dependencies. In the `frontend` container, it installed 1G of dependencies and copr-fe-dev isn't much better. It doesn't want to print me the total install size but the total download size would be 511M. Makes me wonder if we really want to incorporate this script into copr-frontend.

Otherwise, the script runs around one minute, and the grouping feature needs to be re-written using https://plotly.com/python/treemaps/#nested-layers-in-treemap

praiskup:
Indeed, seems like material for a separate container image, and we could run it on FE with `podman`?
Don't we have some foss solution in JS?

praiskup:
Seems this is called treemap, and we could use [d3](https://observablehq.com/@d3/treemap) e.g.?

FrostyX:
> Indeed, seems like material for a separate container image, and we could run it on FE with podman?

We don't have podman on FE yet, so it will be `Installed size: 294 M` and some qemu dependencies but I think its a better option.

FrostyX:
I found [pygal](https://www.pygal.org/en/stable/) which can generate treemaps as well and has minimal dependencies. That will probably be easiest.

praiskup:
I'd still probably prefer having zero deps in python, and just generating the dataset for some JS library.  I mean, it's better to delegate the computational power related to charts to end-users from our httpd server.  At least as long as the work on the feature is similarly man-power-expensive.

xsuchy:
+1 to using pygal. This is indeed easier to maintain.


Can you add a route that displays the latest file?

This is what I get on dev instance. 
2023-04-24-usage_treemap_grouped
It LGTM here, but when I display it on my workstation I see:
Screenshot_20230424_125010
I do not see the labels there.

The legend in Fedora's Copr can be huge. Can you print the label inside of the square?

docker/frontend/files/etc/copr/copr.conf
```diff
@@ -136,3 +136,9 @@ MAX_BUILD_TIMEOUT = 108000
 
 # Disable template caching
 CACHE_TYPE = "NullCache"
+
+USAGE_TREEMAP_TEAMS = {
+    "CPT": ["@copr", "@mock", "msuchy", "praiskup", "frostyx", "nikromen"],
```
praiskup, line 141: Just discrepancy in the way we handle @copr/PyPI, which is mosthe tly a place where Python team works. Not a big issue, but it might look like we are the most consuming consumers of our service.  So the grouping could be done on a per-project level (at least in the future).

FrostyX: Or maybe we can move `@copr/PyPI` outside of the `@copr` namespace? We could then create an empty `@copr/PyPI` with no builds and specify a runtime dependency for the outside-of-copr project, to ensure that all the articles written about it still work?

praiskup (approved):

PR #2686, opened 2023-04-24
Fix our dist-git CI

### LINKED ISSUE #2317: dist-git - avoid the downstream-only hacks we are using for autospec
There are now two problems to fix:

- [ ] `spec_file_processed_by_rpmautospec` check needs to be disabled for copr dist git deployments
- [ ] the `--allow-empty` option shouldn't be used while importing SRPMs and unskip test_no_op_1 (see #2600)

---

Original issue: https://pagure.io/copr/copr/issue/2317
Opened: 2022-09-22 13:23:01

We need a way around this (a new opt-out in rpkg, I guess, and proper configuration in copr-dist-git or dist-git):
https://pagure.io/rpkg/c/3087dd72e933568896b2f167bab75194482a6cd1

The reason is that in Copr we **want to import the pre-generated SRPMs** into our "proxy" DistGit instance.  This way we keep the changelog and Release info baked into the Fedora DistGit "git log".

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2022-09-22 23:03:11:
For now, we can apply this patch (ansible patching), and things should be fine...

```diff
--- /usr/lib/python3.10/site-packages/pyrpkg/utils.py   2022-09-06 20:05:35.000000000 +0000            [===                                                                                               ] ---  B/s |   0  B     --:-- ETA
+++ /usr/lib/python3.10/site-packages/pyrpkg/utils.py   2022-09-22 19:08:16.328405267 +0000
@@ -323,6 +323,7 @@
 
 
 def spec_file_processed_by_rpmautospec(file_name, dir_path=None):
+    return False
     file_path = os.path.join(dir_path or "", file_name)
 
     try:
```

### LINKED ISSUE #2608: copr-dist-git fails to build on rawhide
https://download.copr.fedorainfracloud.org/results/@copr/copr-pull-requests:pr:2600/fedora-rawhide-x86_64/05738991-copr-dist-git/builder-live.log.gz

my wild guess is that some dependent package is updated in rawhide with new unexpected behavior and causes trouble. I think it will be some package inside rpkg but it requires further investigation.

nikromen:
this PR is an attempt to shut down our CI problems with pyrpkg

`dist-git: refresh pyrpkg.Commands after import_srpm` commit - WDYT? this shouldn't be a problem for us from the efficiency side

- [x] wait until [rpkg updates](https://bodhi.fedoraproject.org/updates/?packages=rpkg) will be in stable

after merging this, don't forget to merge https://pagure.io/fedora-infra/ansible/pull-request/1386

Fixes #2608 
Fixes #2317

### DIFF
--- a/dist-git/copr_dist_git/package_import.py
+++ b/dist-git/copr_dist_git/package_import.py
@@ -148,6 +148,30 @@ def cleanup_repo(repo_path):
             ['git', 'rm', '-r'] + to_remove)
 
 
+def _load_commands(opts, repo_name, repo_dir):
+    # use rpkg lib to import the source rpm
+    commands = Commands(path=repo_dir,
+                        lookaside="",
+                        lookasidehash="md5",
+                        lookaside_cgi="",
+                        gitbaseurl=opts.git_base_url,
+                        anongiturl="",
+                        branchre="",
+                        kojiprofile="",
+                        build_client="")
+
+    # rpkg gets module_name as a basename of git url
+    # we use module_name as "username/projectname/package_name"
+    # basename is not working here - so I'm setting it manually
+    commands.repo_name = repo_name
+
+    # rpkg calls upload.cgi script on the dist git server
+    # here, I just copy the source files manually with custom function
+    # I also add one parameter "repo_dir" to that function with this hack
+    commands.lookasidecache.upload = types.MethodType(my_upload_fabric(opts), repo_dir)
+    return commands
+
+
 def import_package(opts, namespace, branches, srpm_path, pkg_name):
     """
     Import package into a DistGit repo for the given branches.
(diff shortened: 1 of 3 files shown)


nikromen:
this PR on rpkg https://pagure.io/rpkg/pull-request/691 solves our issue so this PR is not needed

dist-git/copr_dist_git/package_import.py
```diff
@@ -218,6 +224,15 @@ def import_package(opts, namespace, branches, srpm_path, pkg_name):
...
                 upload_files = commands.import_srpm(srpm_path)
+                # in case of importing, the content of directory in `reponame`
+                # changes. To update the state of `Commands` class isn't an easy
+                # process - look how much logic pyrpkg.cli.cliClient.load_cmd has
+                # and the logic is written for context in cliClient, not for Commands
+                # which we are using. Considering this wouldn't be an easy task for
+                # pyrpkg to implement, we can refresh the Commands class after every
+                # import - even if it is not elegant
+                # note: if https://pagure.io/rpkg/issue/690 is resolved, you may delete this
+                commands = _load_command(opts, reponame, repo_dir)
```
praiskup, line 235: @onosek do you think this change makes sense?

praiskup:
Should we have a particular versioned `Require:` on python3-rpkg?  Or is this going to work with both old and new versions?

dist-git/copr_dist_git/package_import.py
praiskup: Ping @onosek :-) seems like we could rather fix the library.  Can you give us guidance?

onosek: I prepared a fix that potentially solves the issue:
https://bodhi.fedoraproject.org/updates/?search=rpkg-1.66-8

PR #2689, opened 2023-04-26
frontend, python, cli: general check if it makes sense to upload SRPM

### LINKED ISSUE #2658: Generally check if we can build in a project before uploading
This RFE is about adding a check that allows us to build with given arguments first, before we actually start uploading large srpms.  See the example:
```
$ copr build ping --chroot centos-stream+epel-next-9-x86_64 /tmp/quick-package/dummy-pkg-20230406_1042-1.src.rpm
Uploading package /tmp/quick-package/dummy-pkg-20230406_1042-1.src.rpm
 |################################| 8.0 kB 19.4 kB/s eta 0:00:00

Something went wrong:
Error: Bad request parameters: {'chroots': ["'centos-stream+epel-next-9-x86_64' is not a valid choice for this field."]}
```

FrostyX:
Fix #2658
See #2524

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -1329,6 +1329,14 @@ def _get_build_form(active_chroots, form, package=None):
     return F
 
 
+class BuildFormCheckFactory:
+    """
+    For checking before SRPM upload
+    """
+    def __new__(cls, active_chroots, package=None):
+        return _get_build_form(active_chroots, BaseForm, package)
+
+
 class BuildFormScmFactory(object):
     def __new__(cls, active_chroots, package=None):
         return _get_build_form(active_chroots, PackageFormScm, package)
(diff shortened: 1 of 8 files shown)


frontend/coprs_frontend/coprs/forms.py
```diff
@@ -1329,6 +1329,14 @@ def _validate_batch_opts(form, field):
     return F
 
 
+class BuildFormCheckFactory:
+    """
+    For checking before SRPM upload
+    """
+    def __new__(cls, active_chroots, package=None):
+        return _get_build_form(active_chroots, BasePackageForm, package)
```
praiskup, line 1337: This makes sense, nice.

praiskup: Even though, shouldn't we use `BaseForm` instead of `BasePackageForm`?  I'm - again after some time - lost in the Form abstraction :-)  forms.py deserves some simplification IMO.

praiskup:
Very nice one, thanks!

frontend/coprs_frontend/coprs/forms.py
FrostyX: > Even though, shouldn't we use BaseForm instead of BasePackageForm? 

You are right, `BaseForm` should work as well. I chose `BasePackageForm` because some (and I thought that all) other build forms inherit from it. But funnily enough, the only exception is `BuildFormUploadFactory`. I'll change this PR to use `BaseForm`.

> I'm - again after some time - lost in the Form abstraction :-) 

As it turns out, so do I :-)

> forms.py deserves some simplification IMO.

I am hoping it will be touched upon during the OpenAPI implementation.

FrostyX: Updated

praiskup (approved):

PR #2694, opened 2023-04-27
doc: explain that SRPM is downloaded from DistGit only once per build

### LINKED ISSUE #2528: Clarify that "No branch selected by default" is the expected behavior
It is unclear to me whether this behavior is expected or not.

When starting a new build through distgit, COPR doesn't select the branch corresponding to the chroot that is built on.

Example: if you have a fedora-36-x86_64  chroot, and start a distgit build for Satyr package, while copr will properly select the distgit repository, the default branch selected is "wrong": COPR will try to build rawhide HEAD instead of f36 HEAD.

Since the committish information is optional, and since the correct repository is selected, I would have assumed that the proper branch is also selected.
Thus, looking at the "import.log" it can be clearly seen that f36 branch is selected in the COPR git, while the source "build-live.log" doesn't select any specific branch (hence the default one is taken). It feels like the logic is available to select the correct branch, but not always used.

If that is the expected behaviour, may you make it more explicit in the UI/documentation ?

Assuming you have multiple chroots in a project, I understand that it may be complicated to have multiple sources simultaneously if the branches diverged.

FrostyX:
Fix #2528

### DIFF
--- a/doc/user_documentation.rst
+++ b/doc/user_documentation.rst
@@ -144,6 +144,28 @@ It's even easier for a Fedora Rawhide package::
 because 'fedora' distgit is the default, and we automatically pick the default
 branch.
 
+.. note::
+    Please note that an SRPM is downloaded from the specified DistGit instance
+    only once per Copr build, regardless of the number of chroots you
+    build for.
+
+    It **is not** the case that within one Copr build, e.g.
+    ``fedora-37-x86_64`` chroot would be built from the ``f37``
+    branch, ``fedora-38-x86_64`` from the ``f38`` branch, and
+    ``fedora-39-x86_64`` from the ``rawhide`` branch. You can,
+    however, use the **Committish** field to specify what DistGit
+    branch should be used (by default, it is ``rawhide`` for
+    `Fedora DistGit <https://src.fedoraproject.org/>`_).
+
+    The sources from your DistGit branch (e.g. ``rawhide``) can be
+    incompatible with some of the target chroots (e.g. ``epel-8-x86_64``)
+    because of different dependencies, build tooling, etc. A typical
+    workaround is to submit multiple builds, e.g.::
+
+      copr build-distgit ping --name <package> --chroot fedora-rawhide-x86_64 --commit rawhide
+      copr build-distgit ping --name <package> --chroot epel-8-x86_64 --commit epel8
+
+
 PyPI
 ^^^^
 


doc/user_documentation.rst
```diff
@@ -144,6 +144,18 @@ It's even easier for a Fedora Rawhide package::
 because 'fedora' distgit is the default, and we automatically pick the default
 branch.
 
+.. note::
+    Please note that an SRPM is downloaded from the specified DistGit
+    only once per Copr build, regardless of the number of enabled
```
praiskup, line 149: I tend to say "DistGit instance", not just DistGit, but?? dunno

FrostyX: I'll update

praiskup: s/enabled/chroots you build for/

praiskup: or against

doc/user_documentation.rst
```diff
@@ -144,6 +144,18 @@ It's even easier for a Fedora Rawhide package::
...
+    chroots. It **is not** the case that within one Copr build,
+    ``fedora-37-x86_64`` chroot would be built from the ``f37``
+    branch, ``fedora-38-x86_64`` from the ``f38`` branch, and
+    ``fedora-39-x86_64`` from the ``rawhide`` branch. You can,
```
praiskup, line 153: I know what you try to say by the `It is not ...` part, but it might be hard for those who don't know the behavior.
What about:
```
.. enabled chroots.  The effect is that the sources selected (e.g. Fedora `rawhide` DistGit branch) can easily be incompatible with the target chroot (e.g. `EPEL`, different dependencies, build tooling, etc.).
```

doc/user_documentation.rst
```diff
@@ -144,6 +144,18 @@ It's even easier for a Fedora Rawhide package::
...
+    however, use the **Committish** field to specify what DistGit
+    branch should be used (by default, it is ``rawhide`` for
+    `Fedora DistGit <https://src.fedoraproject.org/>`_).
```
praiskup, line 156: And typical work-around (to use different sources for different chroots) is to
submit multiple builds like:
```
$ copr build-distgit ping --name <package> --chroot fedora-rawhide-x86_64 --distgit fedora --commit rawhide
$ copr build-distgit ping --name <package> --chroot epel-8-x86_64         --distgit fedora --commit epel8
```

FrostyX: Good idea, added.

doc/user_documentation.rst
FrostyX: Fixed

doc/user_documentation.rst
FrostyX: I kinda think what you are saying and what I am saying, are different but both important things. I kept my explanation but added one more paragraph.

What do you think?

praiskup: If this isn't a duplicate to the part `The sources from your DistGit branch...` below, then I don't get what you are saying!  :-)  We can discuss this on monday..  might be an issue on my radio, I admit.

praiskup (approved):

Issue #2696, opened 2023-05-03
Can't login on frontend (Error 405)


jn64:
When I click [login](https://copr.fedorainfracloud.org/login/), after entering credentials on id.fp.o, it redirects to a URL like this:

`https://copr.fedorainfracloud.org/login/?next=https://copr.fedorainfracloud.org/coprs/&openid_complete=yes&janrain_nonce=2023-05-03T06%3A57%3A15ZUyM510`

with this error message:

> Error 405: Unknown error
> The method is not allowed for the requested URL.

If I click login again, it goes to:

`https://copr.fedorainfracloud.org/login/?next=https://copr.fedorainfracloud.org/login/%3Fnext%3Dhttps://copr.fedorainfracloud.org/coprs/%26openid_complete%3Dyes%26janrain_nonce%3D2023-05-03T06%253A57%253A15ZUyM510&openid_complete=yes&janrain_nonce=2023-05-03T06%3A57%3A42Z4Iux5n`

with the same error message (can repeat to mangle the URL more and more)

Tried clearing all cookies from copr.fedorainfracloud.org, id.fp.o, and accounts.fp.o.

I can login to accounts.fp.o.

praiskup:
Confirmed, I can reproduce this.  The way around this is to use Kerberos/GSSAPI for now:
- take `fkinit` ticket
- configure your browser to accept tickets from fedoraproject.org (perhaps already done)
- use gssapi-login link at the top of the copr frontend page

PR #2697, opened 2023-05-03
frontend: accept the OpenID response as a POST request

praiskup:
Seems like something changed in the Fedora OpenID service, and that the second /login/ route request coming from the service to us is newly done as POST (not GET).

Closes: #2696


praiskup:
Hot-fixed in staging, can anyone take a look so I can patch in production?

FrostyX:
According to the https://pythonhosted.org/Flask-OpenID/, the `@loginhandler` function should be `methods=['GET', 'POST']`, so +1

FrostyX (approved):

praiskup:
Ok, merging to close the issue.  Hot-fixed in production.

Issue #2698, opened 2023-05-03
llvm-snapshots copr is missing gpg key


mh21:
https://copr.fedorainfracloud.org/coprs/g/fedora-llvm-team/llvm-snapshots/ is forked from the latest working nightly repo.

Since today, when trying to install a package from it

```
sudo dnf copr enable @fedora-llvm-team/llvm-snapshots
sudo dnf update
```

we get the error:

```
Status code: 404 for https://download.copr.fedorainfracloud.org/results/@fedora-llvm-team/llvm-snapshots/pubkey.gpg
```

mh21:
Some debugging on Slack:

- All the pubkeys listed are not reachable
- The pubkey is present in the project it's forked from: https://download.copr.fedorainfracloud.org/results/%40fedora-llvm-team/llvm-snapshots-incubator-20230502/
- can work-around by doing at least one build in the forked project

praiskup:
Appears we should install the GPG keys: https://github.com/fedora-copr/copr/blob/dda9591b8b9e7b2541e6926b65a28579008a8d09/backend/copr_backend/actions.py#L177-L180

praiskup:
The corresponding logs:

```
[2023-05-03 01:19:14,992][  INFO][PID:4154575][/usr/bin/copr-backend-process-action.managed.pid-4154575][copr-backend-process-action:handle_action:24] Handling action 1199442
[...]

PR #2700, opened 2023-05-04
backend: forking: better handle FileExistsError

praiskup:
Relates: #2698

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -606,11 +606,21 @@ def get_backend_opts():
 
 
 def ensure_dir_exists(path, log):
-    if not os.path.isdir(path):
-        try:
-            os.makedirs(path)
-        except OSError as e:
-            log.exception(str(e))
+    """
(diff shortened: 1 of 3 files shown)


backend/copr_backend/helpers.py
```diff
@@ -606,11 +606,21 @@ def get_backend_opts():
...
+    Create the directory if it doesn't exist yet.  But still "raise" if it can
+    not be created.
+    """
+    try:
+        os.makedirs(path)
+        log.info("Directory %s created", path)
+    except FileExistsError:
+        if os.path.isdir(path):
+            return
+        log.exception("Can't create %s directory, file exists but not a file", path)
```
FrostyX, line 619: > file exists but not a file

Shouldn't this be "path exists but is not a directory"?

praiskup: Yep

praiskup:
Fixed, PTAL

FrostyX (approved):

PR #2709, opened 2023-05-10
frontend: allow submitting EPEL8 module builds

### LINKED ISSUE #2703: Enable modules for el8
It seems we cannot create modules for EL8.

From my inbox:

I had read through that and ran into:

```
copr-cli build-module --yaml modulemd/modulemd-foreman-el8.yaml @theforeman/foreman-nightly

Something went wrong:
Error: Unexpected platform 'el8', it should be e.g. f28 or -f30
```

On Fri, May 5, 2023 at 9:25 AM Miroslav Suchý wrote:

    Dne 05. 05. 23 v 15:19 Eric Helms napsal(a):
    > We have to create a module for our repositories to properly function. I am thinking we are going to just need as part
    > of our process copying the Copr repositories to a stage environment to inject the modular metadata and then for
    > testing before we promote them to our release location.

    http://frostyx.cz/posts/how-to-build-modules-in-copr

    Miroslav

FrostyX:
Fix #2703

Previously we validated the modulemd `buildrequires.platform` and allowed only Fedora. I am updating it to EPEL as well. In ideal world this would be in a config file but it is IMHO not worth it.


frontend/coprs_frontend/coprs/logic/modules_logic.py
```diff
@@ -157,16 +157,18 @@ def platform_chroots(self):
...
         for chroot in self.copr.active_chroots:
             chroot_archs.setdefault(chroot.name_release, []).append(chroot.arch)
 
         def abbrev2chroots(abbrev):
-            name_release = abbrev.replace("-", "").replace("f", "fedora-")
+            abbrev = abbrev.replace("-", "")
+            if abbrev.startswith("f"):
+                name_release = abbrev.replace("f", "fedora-")
+            elif abbrev.startswith("el"):
+                name_release = abbrev.replace("el", "epel-")
```
praiskup, line 171: Typically I translate "el" into "rhel" (or just Enterprise Linux).  Can you please document why epel?  Would this work against the RHEL chroots?

FrostyX: I updated the PR, to match "el" to both EPEL and RHEL. This way whatever of those chroots will be enabled in the project, the module will be built for it.

praiskup (approved):
Thank you!

PR #2713, opened 2023-05-12
add support for OIDC

### LINKED ISSUE #2422: OpenID Connect supported
Hi again,
I notice that copr can be integrated with some openID group or kerberos infra
For now, our infra using a [OpenID Connect](https://openid.net/connect/) facility to support authentication, is it possiable for copr to support this spec?

pkking:
The first PR for #2422 
1. add some config items for OpenID Connect
2. OpenID Connect supported is a opt-in feature
3. Integrated with authlib

Fixes #2422

### DIFF
--- a/frontend/coprs_frontend/coprs/views/misc.py
+++ b/frontend/coprs_frontend/coprs/views/misc.py
@@ -16,8 +16,9 @@ from coprs.logic.complex_logic import ComplexLogic
 from coprs.logic.users_logic import UsersLogic
 from coprs.exceptions import ObjectNotFound
 from coprs.measure import checkpoint_start
-from coprs.auth import FedoraAccounts, UserAuth
-
+from coprs.auth import FedoraAccounts, UserAuth, OpenIDConnect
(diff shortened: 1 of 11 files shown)


praiskup:
Thank you for the pull-request, @pkking !

frontend/coprs_frontend/coprs/views/misc.py
```diff
@@ -89,6 +89,64 @@ def oid_login():
     return FedoraAccounts.login()
 
 
+if OpenIDConnect.enabled(app.config):
```
praiskup, line 92: I'm not sure about this "global" module-level code flow.  We should do this minimally, and we typically do this in `frontend/coprs_frontend/coprs/__init__.py`  (with lots of ugly hacks).  Would you mind moving it there?  Perhaps wrapping the logic with something like  `enable_oidc_authentication(app)` (implemented in auth.py?) and just calling it in `__init__.py`?

praiskup (approved):

PR #2721, opened 2023-05-15
get info about build from SRPM's spec file

### LINKED ISSUE #1701: The name-version-release in messages doesn't match the reality
Original issue: https://pagure.io/copr/copr/issue/1701
Opened: 2021-02-19 12:02:18

https://apps.fedoraproject.org/datagrepper/id?id=2021-57120f7f-7540-491e-ba41-511db72f1c15&is_raw=true&size=extra-large

That is message for epel-8-x86_64, but NVR contains fc33 dist tag.

```
    "chroot": "epel-8-x86_64", 
    "ip": "54.167.254.18", 
    "user": "https://api.github.com/users/subpop", 
    "who": "backend.worker-rpm_build_worker:1994239-epel-8-x86_64", 
    "pid": 768371, 
    "copr": "yggdrasil", 
    "version": "0.0.git.248.6bd0e5d-1.fc33", 
```

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2021-12-08 13:18:07:
This is not only a messaging problem.  We store the Build (not BuildChroot) NVR
into the database once we have the source RPM in hand - and that happens when
we finish the source RPM build.  When our builders are on F34, then the src RPM
is built with %_dist .fc34, and then this is part of src.rpm Release.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2021-12-08 13:20:16:
FTR, this confuses users (I got the question "why is there the wrong dist tag" several times).
We should store just the "numeric" release value per-Build (not per BuildChroot).

---

True

nikromen:
I am wondering whether we want to use information from spec file to name the SRPM? This means that SRPM name would be `example-1.0.14-1.src.rpm` instead of `example-1.0.14-1.fc37.src.rpm`

Fixes #1701

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -27,6 +27,7 @@ import pytz
 
 import munch
 from munch import Munch
+from specfile import Specfile
 
 from copr_common.redis_helpers import get_redis_connection
 from copr.v3 import Client
@@ -632,7 +633,47 @@ def get_chroot_arch(chroot):
     return chroot.rsplit("-", 2)[2]
 
 
-def pkg_name_evr(srpm_path, logger):
+def find_spec_file(folder, logger):
+    """
+    Returns spec file path or None
+    """
+    spec_pattern = os.path.join(folder, "*.spec")
+    result = glob.glob(spec_pattern)
+    if not result:
+        return None
+
+    if len(result) > 1:
+        logger.error("More than one spec file found! Using the first %s from %s",
+                     str(result[0]), str(result))
+
+    return result[0]
+
+
+def format_evr(epoch, version, release):
+    """
+    Return evr in format (epoch:)version-release
+    """
+    if epoch is not None and epoch.isdigit():
+        return f"{epoch}:{version}-{release}"
+
+    return f"{version}-{release}"
+
+
+def pkg_name_evr(spec_file_path, srpm_path, logger):
+    """
+    Queries a SRPM or spec file for its name and evr (epoch:version-release)
+    """
+    try:
+        spec = Specfile(spec_file_path)
+    except Exception:  # pylint: disable=broad-exception-caught
+        # Specfile library raises too many exception to name the in except block
+        logger.exception("Exception appeared during handling spec file: %s", spec_file_path)
+        return pkg_name_evr_from_srpm(srpm_path, logger)
+
+    return spec.expanded_name, format_evr(spec.expanded_epoch, spec.expanded_version, spec.expanded_release)
+
+
+def pkg_name_evr_from_srpm(srpm_path, logger):
     """
     Queries a package for its name and evr (epoch:version-release)
     """
@@ -642,7 +683,7 @@ def pkg_name_evr(srpm_path, logger):
     try:
         result = run_cmd(cmd, logger=logger)
     except OSError as e:
-        raise CoprBackendSrpmError(str(e))
+        raise CoprBackendSrpmError(str(e)) from e
 
     if result.returncode != 0:
         raise CoprBackendSrpmError('Error querying SRPM')
@@ -650,15 +691,9 @@ def pkg_name_evr(srpm_path, logger):
     try:
         name, epoch, version, release = result.stdout.split(" ")
     except ValueError as e:
-        raise CoprBackendSrpmError(str(e))
+        raise CoprBackendSrpmError(str(e)) from e
 
-    # Epoch is an integer or '(none)' if not set
-    if epoch.isdigit():
-        evr = "{}:{}-{}".format(epoch, version, release)
-    else:
-        evr = "{}-{}".format(version, release)
-
-    return name, evr
+    return name, format_evr(epoch, version, release)
 
 
 def format_filename(name, version, release, epoch, arch, zero_epoch=False):
--- a/rpmbuild/main.py
+++ b/rpmbuild/main.py
@@ -3,6 +3,7 @@
 import re
 import os
 import fcntl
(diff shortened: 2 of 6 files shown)


rpmbuild/main.py
```diff
@@ -225,6 +226,10 @@
     log.info("Output: {0}".format(
         os.listdir(resultdir)))
 
+    srpm_path = locate_srpm(resultdir)
+    if os.stat(srpm_path).st_size < 4 * (10 ** 9):
+        subprocess.run("bsdtar -xf *.src.rpm *.spec", shell=True)
```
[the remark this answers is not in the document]
praiskup: I think we could use `rpm2archive -n < mock-2.0-2.git.17.e0a6e45.fc31.src.rpm | tar xf - '*.spec'`.

@pmatilai is there a better way to extract a spec file from SRPM?

backend/copr_backend/helpers.py
```diff
@@ -627,7 +628,40 @@ def get_chroot_arch(chroot):
...
 
 
-def pkg_name_evr(srpm_path, logger):
+def find_spec_file(folder):
+    """
+    Returns spec file path or None
+    """
+    spec_pattern = os.path.join(folder, "*.spec")
+    result = glob.glob(spec_pattern)
+    if not result:
```
praiskup, line 637: We should do some action if there are multiple spec files.

rpmbuild/main.py
```diff
@@ -225,6 +226,9 @@ def build_srpm(args, config):
     log.info("Output: {0}".format(
         os.listdir(resultdir)))
 
+    # extract spec file from SRPM
+    subprocess.run(f"rpm2archive -n < {locate_srpm(resultdir)} | tar xf - '*.spec'", shell=True, check=False)
```
praiskup, line 230: Here we should decide what happens if there are multiple source RPM files.

backend/copr_backend/helpers.py
```diff
@@ -627,7 +628,40 @@ def get_chroot_arch(chroot):
...
+    if epoch is not None and epoch.isdigit():
+        return f"{epoch}:{version}-{release}"
+
+    return f"{version}-{release}"
+
+
+def pkg_name_evr(spec_file_path, srpm_path, logger):
+    """
+    Queries a SRPM or spec file for its name and evr (epoch:version-release)
+    """
```
praiskup, line 656: I'd wrap the whole call of this method into a `try-except` block;  eventually raising a clear human-readable error to the user if something goes wrong.

praiskup:
Otherwise looks nice!

rpmbuild/main.py
nikromen: could there be a situation where there are more SRPMs in the resultdir?

praiskup: It's not a valid case, but anything can happen (we've seen this already).

backend/copr_backend/helpers.py
```diff
@@ -627,7 +628,49 @@ def get_chroot_arch(chroot):
...
+    if spec_file_path is None:
+        return pkg_name_evr_from_srpm(srpm_path, logger)
+
+    try:
+        spec = Specfile(spec_file_path)
+    except Exception as exc:
+        logger.error("Exception appeared during handling spec file: %s", str(exc))
+        raise CoprBackendSrpmError(str(exc)) from exc
```
praiskup, line 668: Small wish, can we reverse this logic a bit and - if Specfile parsing fails - fallback to `pkg_name_evr_from_srpm` as we did before this PR?

praiskup:
Otherwise lgtm!

backend/copr_backend/helpers.py
nikromen: I am afraid I don't get it here. Do you mean by reversing the logic - if Specfile parsing fails, then call the `pkg_name_evr_from_srpm`?

backend/copr_backend/helpers.py
```diff
@@ -629,7 +630,46 @@ def get_chroot_arch(chroot):
...
+
+
+def pkg_name_evr(spec_file_path, srpm_path, logger):
+    """
+    Queries a SRPM or spec file for its name and evr (epoch:version-release)
+    """
+    try:
+        spec = Specfile(spec_file_path)
+    except Exception as exc:
+        logger.error("Exception appeared during handling spec file: %s", str(exc))
```
praiskup, line 666: Don't you want to do `log.exception` here?  Just
```python
logger.exception("Exception appeared during handling spec file: %s", spec_file_path)
```

praiskup:
One nit, otherwise good to go

praiskup (approved):
Thanks!

praiskup:
@nforro JFYI we'll run the python3-specfile  code in every single Copr build now.

What's the state of exception hierarchy in the library?  Can we be 100% sure no other
exception than `SpecfileException`(-based) one bubbles out from the library calls?

nforro:
> What's the state of exception hierarchy in the library? Can we be 100% sure no other
> exception than `SpecfileException`(-based) one bubbles out from the library calls?

I believe you can be sure the constructor won't raise anything but `SpecfileException`-based exception. Other than that, you can expect `AttributeError` for example when trying to access a section or a tag that doesn't exist, and in some rare cases also `ValueError`.

praiskup:
Thank you, I actually submitted #2765 now - but this isn't ideal.  @nikromen also claimed it can raise OSError/FileNotFoundError or something alike.

The thing is that it is not really convenient to catch all of those, /me shrugs.

PR #2724, opened 2023-05-16
doc: describe Copr-specific RPM macros

### LINKED ISSUE #2695: Document Copr specific RPM macros
We should document our `%copr_username`, `%copr_projectname`, `%vendor`, etc macros somewhere. Ideally with specfile snippets, using the macros in conditions (Copr/Koji).

FrostyX:
Fix #2695


FrostyX:
There seems to be some bug in `main`, causing the `runtest-pkg-with-macros.sh` to fail. I am setting a `blocked` tag for this PR until it gets fixed and I can write a test for `vendor` and `buildtag`.

doc/user_documentation.rst
```diff
@@ -481,6 +481,38 @@ Quick HOWTO for the `Package Review`_ time::
...
+- ``%buildtag`` - This macro contains an ID of a Copr build,
+  e.g. ``.copr5925897``
+
+Macros for SRPM builds:
+
+- ``%dist`` - Regardless of the number of chroots that a build is
+  submitted for, the SRPM package is built only once. And its
+  ``%dist`` tag would be determined by the operating system on a
+  builder. To avoid such confusion, we redefine ``%dist`` to ``%nil``
+  for SRPM builds.
```
praiskup, line 509: can you point at the issue @nikromen is dealing with?  this is a bit off-topic though as %dist is not defined by Copr -- or maybe, we should point out that "we try to undefine it as much as possible" to be distro-agnostic.

praiskup:
Thank you for this PR, very good job!

PR #2728, opened 2023-05-18
fix: common build failed in openEuler

pkking:
ADD build supported in openEuler env, as openEuler use python3 as default

### DIFF
--- a/common/python-copr-common.spec
+++ b/common/python-copr-common.spec
@@ -5,12 +5,12 @@
 %global _pkgdocdir %{_docdir}/%{name}-%{version}
 %endif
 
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Maybe instead change the conditionals so Python 3 is done _unless_ specific distros need it to be Python 2. Or maybe we can rip out all the Python 2 stuff?

praiskup:
> Or maybe we can rip out all the Python 2 stuff?

I'd prefer keeping that one actually; there might be a need to depend on it in python-copr in the future.

praiskup (approved):
reversed conditions would be better, but it is a nit, +1

PR #2732, opened 2023-05-18
beaker-tests-sanity: don't test project searching pt.2

FrostyX:
See PR #2731

I made a mistake and forgot to comment the following code as well:

    cleanProject "${NAME_PREFIX}Project8"
    cleanProject "${NAME_PREFIX}Project9"

Now the project deletion fails because the projects are not being created in the first place. This PR fixes it. Alternatively I could comment the project deletion but then we will surely forget to uncoment them in the future because they are too far from the project search tests.


nikromen (approved):

praiskup:
Well, is this actually a release blocker, or could we merge post-release? :-)

FrostyX:
We can merge this whenever. It is a FAIL in tests, so I marked it a release-blocker, but nothing will happen if we merge afterward.

PR #2733, opened 2023-05-19
Allow setting up repo priority

### LINKED ISSUE #2568: Allow setting `priority=X` for the copr_base repo
This might be useful for mass rebuilds where users can't guarantee doing
version bumps (or frequent enough version bumps)  - and they prefer
installing content from copr repository instead of Fedora default (even if
the Fedora packages have higher NVRs).  E.g. rebuilding the distro with a
newer GCC stack or Clang might be a good example.

Original report by @fweimer.

We already have the priority= option support for external repositories;
the RFE is about having direct support for setting priority= option
for the copr default build repo itself.  Work-around is to re-add "self"
(copr_repo) again as external repo with the prirority= tag set.

I'm not really sure about all the possible consequences.

This is a build-time option, not affecting the repo files installed by
 `dnf copr enable`.

nikromen:
TODO:
- [x] tests
- [x] add it to copr-cli
- [x] some documentation

Fixes #2568

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -15,13 +15,13 @@ except ImportError:
     from flask_wtf import Form as FlaskForm
 
 from coprs import app
+from coprs import exceptions
 from coprs import helpers
 from coprs import models
 from coprs.logic.coprs_logic import CoprsLogic, MockChrootsLogic
 from coprs.logic.users_logic import UsersLogic
 from coprs.logic.dist_git_logic import DistGitLogic
 from coprs.logic.complex_logic import ComplexLogic
-from coprs import exceptions
 
 from wtforms import ValidationError
 
@@ -654,6 +654,18 @@ class CoprForm(BaseForm):
         "Packit allowed forge projects",
         filters=[StringListFilter(), StripUrlSchemaListFilter()],
         validators=[wtforms.validators.Optional()],)
+
+    repo_priority = wtforms.IntegerField(
+        "The priority value of this repository",
+        description="""The priority value of this repository, default is 99. If there is more than one candidate
+        package for a particular operation, the one from a repo with the lowest priority value is
+        picked, possibly despite being less convenient otherwise (e.g. by being a lower version).""",
+        render_kw={"placeholder": "Optional - integer, e.g. 22"},
+        validators=[
+            wtforms.validators.Optional(),
+            wtforms.validators.NumberRange(min=1)],
+        default=None,
+    )
 
     @property
     def errors(self):
--- a/frontend/coprs_frontend/coprs/logic/complex_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/complex_logic.py
@@ -14,6 +14,7 @@ from coprs import helpers
 from coprs import models
 from coprs import exceptions
 from coprs import cache
+from coprs.constants import DEFAULT_COPR_REPO_PRIORITY
 from coprs.exceptions import ObjectNotFound, ActionInProgressException
 from coprs.logic.builds_logic import BuildsLogic
 from coprs.logic.batches_logic import BatchesLogic
@@ -456,7 +457,6 @@ class ProjectForking(object):
 
 
 class BuildConfigLogic(object):
-
     @classmethod
     def generate_build_config(cls, copr, chroot_id):
         """ Return dict with proper build config contents """
@@ -485,6 +485,15 @@ class BuildConfigLogic(object):
                 "baseurl": copr.repo_url + "/{}/devel/".format(chroot_id),
                 "name": "Copr buildroot",
             })
+
+        # None value of the priority won't show in API
+        if copr.repo_priority in [None, DEFAULT_COPR_REPO_PRIORITY]:
+            repo_priority = None
+        else:
+            repo_priority = copr.repo_priority
+
+        for repo in repos:
+            repo["priority"] = repo_priority
 
         repos.extend(cls.get_additional_repo_views(copr.repos_list, chroot_id))
         repos.extend(cls.get_additional_repo_views(chroot.repos_list, chroot_id))
--- a/frontend/coprs_frontend/coprs/models.py
+++ b/frontend/coprs_frontend/coprs/models.py
@@ -388,6 +388,17 @@ class _CoprPublic(db.Model, helpers.Serializer):
     # allowed to build in this Copr via Packit
     packit_forge_projects_allowed = db.Column(db.Text)
 
+    # priority=NUMBER in repo configs
+    repo_priority = db.Column(db.Integer, nullable=True)
+
+    @validates("repo_priority")
+    def validate_repo_priority(self, _, value):
+        """Checks whether priority value is correct"""
+        if value < 1:
+            raise ValueError("Repo priority can't be lower than 1")
+
+        return value
+
 
 class _CoprPrivate(db.Model, helpers.Serializer):
     """
--- a/cli/copr_cli/build_config.py
+++ b/cli/copr_cli/build_config.py
@@ -47,7 +47,7 @@ config_opts['dnf.conf'] += \"\"\"
 [{{ repo.id }}]
 name="{{ repo.name }}"
 baseurl={{ repo.baseurl }}
-{%- if repo.priority %}
+{%- if repo.priority and repo.priority is not none %}
 priority={{ repo.priority }}
 {%- endif %}
 {%- if repo.module_hotfixes %}
@@ -63,6 +63,7 @@ best=1
 \"\"\"
 {%- endif %}
 """
+
 
 class MockProfile(object):
     def __init__(self, data):
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -481,6 +481,7 @@ class Commands(object):
             appstream=ON_OFF_MAP[args.appstream],
             runtime_dependencies=args.runtime_dependencies,
             packit_forge_projects_allowed=args.packit_forge_projects_allowed,
+            repo_priority=args.repo_priority,
         )
 
         owner_part = username.replace('@', "g/")
@@ -514,6 +515,7 @@ class Commands(object):
             appstream=ON_OFF_MAP[args.appstream],
             runtime_dependencies=args.runtime_dependencies,
             packit_forge_projects_allowed=args.packit_forge_projects_allowed,
+            repo_priority=args.repo_priority,
         )
 
     @requires_api_auth
@@ -1177,6 +1179,9 @@ def setup_parser():
               "created for you (as soon as they are available) as rawhide "
               "chroot forks."))
 
+    parser_create.add_argument("--repo-priority", default=None,
+                               help="Set the priority value of this repository")
+
     create_and_modify_common_opts(parser_create)
 
     parser_create.set_defaults(func="action_create")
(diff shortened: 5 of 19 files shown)


FrostyX (commented):
Thank you for PR @nikromen,

Overall it looks very good. There are some parts that I think can be simplified, so added some comments to the code.

Apart from that, we should probably squash all the commits together before merging.

build_aux/per-subdir-makefile
```diff
@@ -13,7 +16,12 @@ unittests:
...
+
+enter-test-image:
+	$(CONTAINER_ENGINE) run -v .:/home/copr_user/copr_bind:Z -ti $(COPR_TEST_IMAGE) /bin/bash
+
 unittest-in-container:
-	podman build -f ../build_aux/Containerfile.unittest . -t copr_test_image
-	podman run --rm -it copr_test_image /bin/sh -c "./run_tests.sh -vv -s"
-	podman image rm copr_test_image
+	$(CONTAINER_ENGINE) run --rm -v .:/home/copr_user/copr_bind:Z \
+		-it $(COPR_TEST_IMAGE) /bin/sh -c "./run_tests.sh -vvv -s --no-coverage"
```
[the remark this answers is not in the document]
nikromen: Oh, I should separate this to another PR, sorry :D These make targets are for my use. I run unit tests in container so I made make targets for me in the past. They are somewhat niche currently but I plan to enhance them a bit in the summer

backend/copr_backend/job.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, task_data, worker_opts):
         self.results = None
         self.appstream = None
 
+        self.priority = None
```
praiskup, line 76: I would probably call this "repo_priority", so we don't intuitively mix-up with the task priority we have all around the code.

rpmbuild/main.py
```diff
@@ -82,6 +82,9 @@ def get_parser():
     base_parser.add_argument("--task-url", help="Full URL to a json task definition")
     base_parser.add_argument("--task-file", help="Path to a local json file with task definition")
 
+    # TODO: better help message
+    base_parser.add_argument("--priority", type=int, help="Set priority value of this repository")
```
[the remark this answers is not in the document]
praiskup: Agreed, this should be `frontend` and `rpmbuild` change; optionally `python3-copr` + `copr-cli`.

frontend/coprs_frontend/coprs/models.py
```diff
@@ -388,6 +389,23 @@ class _CoprPublic(db.Model, helpers.Serializer):
     # allowed to build in this Copr via Packit
     packit_forge_projects_allowed = db.Column(db.Text)
 
+    # priority=NUMBER in repo configs
+    priority_raw = db.Column(
```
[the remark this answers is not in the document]
nikromen: Ah, this seems reasonable. I didn't know that I can do this with `nullable=True`, that's why I implemented this hack

rpmbuild/main.py
```diff
@@ -197,6 +197,9 @@ def get_task(args, config, build_config_url_path=None, task_id=None):
     if args.copr:
         task['task_id'] = copr_chroot_to_task_id(args.copr, args.chroot)
 
+    if args.repo_priority:
+        task["repo_priority"] = args.repo_priority
+
```
FrostyX, line 202: I think this is now a leftover that does nothing.

rpmbuild/man/copr-rpmbuild.1.asciidoc
```diff
@@ -42,6 +42,9 @@ OPTIONS
 --verbose::
     Print debugging information.
 
+--priority::
+    Set the priority of this copr repository.
```
FrostyX, line 46: Same here, the `--priority` parameter was removed.

backend/copr_backend/background_worker_build.py
```diff
@@ -450,6 +450,7 @@ def _start_remote_build(self):
             command += " --srpm --task-url {task_url} --detached"
         else:
             command += " --task-url {task_url} --chroot {chroot} --detached"
+
```
FrostyX, line 453: Unnecessary blank line change.

backend/copr_backend/job.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, task_data, worker_opts):
         self.results = None
         self.appstream = None
 
+        self.repo_priority = None
```
FrostyX, line 76: I think this is a leftover

rpmbuild/copr_rpmbuild/builders/mock.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, task, sourcedir, resultdir, config):
         self.buildroot_pkgs = task.get("buildroot_pkgs")
         self.enable_net = task.get("enable_net")
         self.repos = task.get("repos")
+        if task.get("repos") and task.get("repo_priority") is not None:
+            self.repos["priority"] = task.get("repo_priority")
```
FrostyX, line 31: Are you sure this works? 

In `mock.cfg.j2` we do

```
{% for repo in repos %}
...
priority={{ repo["priority"] }}
```

so it looks that `self.repos["priority"]` isn't the thing that needs to be set. But rather `"repo"` key for each repo.

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -9,19 +9,21 @@
 from flask_wtf.file import FileRequired, FileField
 from fnmatch import fnmatch
 
+from coprs.constants import DEFAULT_REPO_PRIORITY_VALUE
```
FrostyX, line 12: I think if you move this import closer to the rest of `coprs` imports, you will fix both linter errors.

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -655,6 +656,17 @@ class CoprForm(BaseForm):
         filters=[StringListFilter(), StripUrlSchemaListFilter()],
         validators=[wtforms.validators.Optional()],)
 
+    repo_priority = wtforms.IntegerField(
+        "The priority value of this repository",
+        description="""If there is more than one package to install from (e.g. vim from
+        Fedora repo and Copr repo), the repo with lower priority is picked.""",
```
FrostyX, line 662: The `man dnf.conf` says one important thing that might be surprising for some users

> repo with the lowest priority value is picked, possibly despite being less convenient otherwise (e.g. **by being a lower version**).

Can we say something like this also? I think maybe even copy-pasting the whole `priority` description from `man dnf.conf` might be a good idea.

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -655,6 +656,17 @@ class CoprForm(BaseForm):
...
+        render_kw={"placeholder": "Optional - integer, e.g. 22"},
+        validators=[
+            wtforms.validators.Optional(),
+            wtforms.validators.NumberRange(min=1, max=DEFAULT_REPO_PRIORITY_VALUE)],
```
FrostyX, line 666: I think the `min` value looks reasonable but IMHO there should be no `max`. Someone might want to de-prioritize their repo and prefer the default repositories. I don't why would they do it but we should make it possible to do. If we want to have `max`, let's make it at least 999.

rpmbuild/copr_rpmbuild/builders/mock.py
```diff
@@ -27,6 +27,11 @@ def __init__(self, task, sourcedir, resultdir, config):
         self.buildroot_pkgs = task.get("buildroot_pkgs")
         self.enable_net = task.get("enable_net")
         self.repos = task.get("repos")
+        if self.repos and task.get("repo_priority") is not None:
+            repo_priority = task.get("repo_priority")
+            for repo in self.repos:
+                repo["priority"] = repo_priority
```
FrostyX, line 33: This doesn't look right. The `self.repos` should contain all the repositories enabled for the Copr project (both the Copr repo itself but also external repos). We IMHO don't want to modify those.

I think we want to specify the priority here
https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/logic/complex_logic.py#L473-L487

FrostyX (commented):
The PR looks very good, just a couple of minor fixes are needed.

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -489,6 +488,11 @@ def generate_build_config(cls, copr, chroot_id):
         repos.extend(cls.get_additional_repo_views(copr.repos_list, chroot_id))
         repos.extend(cls.get_additional_repo_views(chroot.repos_list, chroot_id))
 
+        repo_priority = copr.repo_priority
+        if repo_priority is not None:
+            for repo in repos:
+                repo["priority"] = repo_priority
+
```
FrostyX, line 495: I am not sure if we want to set the priority only for `copr_base` or `copr_base_devel` as well. But IMHO this code block needs to be at least above the 

```python
repos.extend(cls.get_additional_repo_views(copr.repos_list, chroot_id))
repos.extend(cls.get_additional_repo_views(chroot.repos_list, chroot_id))
```

@praiskup, you might want check this, if I am understanding the feature correctly.

praiskup: I agree, `copr_base` and `copr_base_devel` only, and moving up.

nikromen:
I tested it locally in container and API call returns the priority correctly so it should work as expected now

Screenshot_20230725_164022

@FrostyX PTAL

FrostyX (approved):

praiskup:
I'd probably prefer having this less intrusive, I tested this and I basically see the number 99 everywhere, which will cause a wave of questions (the number 99 is suspicious, and people will have to study dnf.conf manual page that we do not even point them to).

Can this be done so that `None` is equivalent to `99` and if that happens, no `priority=` tag is placed into the repo files?

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -486,6 +486,10 @@ def generate_build_config(cls, copr, chroot_id):
                 "name": "Copr buildroot",
             })
 
+        repo_priority = copr.repo_priority or DEFAULT_COPR_REPO_PRIORITY
+        for repo in repos:
+            repo["priority"] = repo_priority
```
praiskup, line 491: I'd probably prefer just:
```python
for repo in repos:
    if copr.repo_priority in [None, DEFAULT_COPR_REPO_PRIORITY]:
        continue
    repo["priority"] = copr.repo_priority
```

nikromen: I didn't find a way how to make a field in flask-restx API model optional (discussed in today's planning) but it seems like it's possible to don't display them if the values are none via [skip_none=True](https://flask-restx.readthedocs.io/en/latest/marshalling.html#skip-fields-which-value-is-none) so I decided to set up `priority` to `None` in these cases which should hide the `99`s and `None`s
PTAL

FrostyX: > I didn't find a way how to make a field in flask-restx API model optional (discussed in today's planning) but it seems like it's possible to don't display them if the values are none via [skip_none=True](https://flask-restx.readthedocs.io/en/latest/marshalling.html#skip-fields-which-value-is-none) so I decided to set up priority to None in these cases which should hide the 99s and Nones

I think we might be over-complicating things. When use this PR and remove the `skip_none` thing

```diff
- @apiv3_project_chroots_ns.marshal_with(project_chroot_build_config_model, skip_none=True)
+ @apiv3_project_chroots_ns.marshal_with(project_chroot_build_config_model)
```

It seems to work anyway and when the priority is not set, its value is `null`.

Screenshot_2023-07-31_23-12-46

This is IMHO a behavior that we want.

praiskup (approved):
Thank you!

nikromen:
@praiskup @FrostyX added priority to repo file per standup mtg, PTAL

praiskup:
Can you please update `copr_cli/build_config.py`, too?

praiskup (approved):

PR #2734, opened 2023-05-21
doc: add list of known public Copr instances

FrostyX:
SSIA

### DIFF
--- a/doc/user_documentation.rst
+++ b/doc/user_documentation.rst
@@ -32,6 +32,42 @@ How to enable copr repository?
 
 :ref:`how_to_enable_repo`
 
+
+Public Copr instances
+---------------------
+
+Copr is a free software and anyone can maintain their own instance in
+case the Fedora Copr instance doesn't suit their needs. This is a list
+of known Copr instances:
+
+.. list-table::
+   :header-rows: 1
+
+   * - Instance
+     - Description
+     - Links
+
+   * - https://copr.fedorainfracloud.org
+     - |
+       | Fedora Copr instance which is also
+       | considered to be the default Copr
+       | instance by many users and tools.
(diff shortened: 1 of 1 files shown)


doc/user_documentation.rst
```diff
@@ -32,6 +32,22 @@ How to enable copr repository?
...
+
+Public Copr instances
+---------------------
+
+Copr is a free software and anyone can maintain their own instance in
+case the Fedora Copr instance doesn't suit their needs. This is a list
+of known Copr instances:
+
+- https://copr.fedorainfracloud.org - Fedora Copr instance which is
+  also considered to be the "main" Copr instance.
```
praiskup, line 44: I'd say *which is the oldest instance, and also considered to be the default Copr instance by many users and tools*

praiskup: I think it would be nice to cite a public issue tracker, or something alike ...

FrostyX: Good idea. So what about a table like this?

Screenshot_2023-06-01_00-18-42

doc/user_documentation.rst
```diff
@@ -32,6 +32,42 @@ How to enable copr repository?
...
+
+   * - https://copr.stg.fedoraproject.org
+     - |
+       | Fedora Copr staging instance is useful for
+       | testing the upcoming changes.
+       | We periodically delete all its data.
+     -
+
+   * - https://eur.openeuler.openatom.cn
+     - OpenEuler Copr instance
```
TommyLike, line 66: sorry it's openEuler not OpenEuler 😄

FrostyX: Oh, you are right. Thank you.
Updated.

PR #2735, opened 2023-05-21
cli: show CHROOT in help output instead of CHROOTS

FrostyX:
Per IRC discussion:

> <JensPetersen[m]>  I was trying to work out where `-r CHROOTS` "comes"
>                    from in `copr --help` say
> <JensPetersen[m]>  Does it accept comma separation?
> <JensPetersen[m]>  I wonder if -r CHROOT wouldn't be less confusing?

I agree, the -r/--chroot parameter does take only one value. The correct way to provide multiple values is to repeat the parameter.


juhp:
Thank you!

PR #2740, opened 2023-05-23
rpmbuild: fix typo in pyrhon binary python -> python3

nikromen:
python-backoff is available only as a `python3-backoff` binary

### DIFF
--- a/rpmbuild/copr-rpmbuild.spec
+++ b/rpmbuild/copr-rpmbuild.spec
@@ -47,7 +47,7 @@ BuildRequires: %{python_pfx}-munch
 BuildRequires: %{python}-requests
 BuildRequires: %{python_pfx}-jinja2
 BuildRequires: %{python_pfx}-simplejson
-BuildRequires: python-backoff >= 1.9.0
+BuildRequires: python3-backoff >= 1.9.0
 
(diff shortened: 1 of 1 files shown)


rpmbuild/copr-rpmbuild.spec
```diff
@@ -47,7 +47,7 @@ BuildRequires: %{python_pfx}-munch
 BuildRequires: %{python}-requests
 BuildRequires: %{python_pfx}-jinja2
 BuildRequires: %{python_pfx}-simplejson
-BuildRequires: python-backoff >= 1.9.0
+BuildRequires: python3-backoff >= 1.9.0
```
FrostyX, line 50: Looking at the lines above, should we use `%{python_pfx}-backoff`?

nikromen: backoff is not available as python-backoff in official repos (it has never been) and python_pfx is set only to `python` (missing the 3 there)

nikromen: but on the other side the package does not exist for el < 8 so I think we can use it. WDYT?

praiskup (approved):

PR #2748, opened 2023-05-29
frontend: don't list temporary projects on main page

### LINKED ISSUE #1867: Don't show temporary projects on COPR HP by default
Original issue: https://pagure.io/copr/copr/issue/1867
Opened: 2021-06-28 10:48:51
Opened by: [dmach](https://accounts.fedoraproject.org/user/dmach)

AFAIK `unlisted_on_hp` defaults to `False`.
It doesn't seem to be a good idea to list temporary projects on COPR HP and the value needs to be changed to `True` explicitly.

How about to set the default value based on if `delete_after_days` is set?
`unlisted_on_hp = bool(delete_after_days)`

---

#### [msuchy](https://accounts.fedoraproject.org/user/msuchy) commented at 2021-06-28 13:37:30:
Sounds reasonable.

---

#### [frostyx](https://accounts.fedoraproject.org/user/frostyx) commented at 2021-06-29 19:14:20:
Agreed

nikromen:
Fixes #1867

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/coprs_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/coprs_logic.py
@@ -128,7 +128,10 @@ class CoprsLogic(object):
             query = query.filter(models.Copr.deleted.is_(False))
 
         if not include_unlisted_on_hp:
-            query = query.filter(models.Copr.unlisted_on_hp.is_(False))
+            query = (
+                query.filter(models.Copr.unlisted_on_hp.is_(False))
+                .filter(models.Copr.delete_after.is_(None))
+            )
 
         return query
 


frontend/coprs_frontend/coprs/logic/coprs_logic.py
```diff
@@ -292,6 +292,10 @@ def add(cls, user, name, selected_chroots, repos=None, description=None,
         # form validation checks for duplicates
         cls.new(user, name, group, check_for_duplicates=check_for_duplicates)
 
+        delete_after_days = kwargs.get("delete_after_days")
+        if delete_after_days is not None:
+            kwargs["unlisted_on_hp"] = bool(delete_after_days) or kwargs.get("unlisted_on_hp", False)
+
```
FrostyX, line 298: Interesting approach but I think this won't work when:

1. I create a new project that is going to be deleted in 30 days
2. Update the project and make it not temporary anymore

With this PR, the project will be forever unlisted?

I was thinking more of an approach where we filter temporary projects when querying from the database. Would that work?

nikromen: > With this PR, the project will be forever unlisted?

then you need to uncheck the `unlisted_on_hp` in project settings. But yep, filtering tmp project from DB sounds better. PTAL

FrostyX: > then you need to uncheck the unlisted_on_hp in project settings.

Oh, right. I didn't think of that.

>  filtering tmp project from DB sounds better. PTAL

Thank you for the update. LGTM now.

FrostyX (approved):

PR #2749, opened 2023-05-31
Appstream builder in pruner

praiskup:


backend/run/copr_prune_results.py
```diff
@@ -170,15 +176,21 @@ def prune_project(self, project_path, username, projectdir):
         LOG.info("projectname = %s", projectname)
 
         try:
-            if uses_devel_repo(self.opts.frontend_base_url, username, projectname):
+            project_info = get_project_info(self.opts.frontend_base_url,
+                                            username, projectname)
+
+            if uses_devel_repo(self.opts.frontend_base_url, username,
+                               projectname, project_info):
```
FrostyX, line 187: Maybe we can go as far as removing the whole `uses_devel_repo` function, and using only

```
if bool(project.get("devel_mode", False)):
    ...
```

Or I overlooked something?

FrostyX (commented):
I agree, this is a big improvement.
Only one comment, otherwise LGTM

backend/run/copr_prune_results.py
praiskup: The method is used in multiple places I think.

praiskup:
It appears pruner is entirely broken.  Thus not hot-fixing, we need to resolve #2750 first.

PR #2753, opened 2023-06-02
rpmbuild: make sure we have the latest DNF5 on builders

### LINKED ISSUE #2739: copr-builder: don't depend on `dnf`
(I'm filing issues with all the packages that currently depend on `dnf`)

Starting in Fedora 39, the `dnf` command will be provided by the `dnf5` package rather than the `dnf` package, and `dnf5` will obsolete `dnf`. The old DNF 4 Python API will still be provided by `python3-dnf`, as well as the DNF 4 command, as `/usr/bin/dnf-3`.

To avoid breaking the Fedora upgrade, copr-builder should switch to depending on `python3-dnf` rather than `dnf` (or dnf-yum).

And since `/usr/bin/dnf` will soon point to DNF 5, which has a slightly different command-line interface, any use of the `/usr/bin/dnf` binary should be replaced with calls to `/usr/bin/dnf-3`.

At some point, this repository should adopt libdnf5, but the immediate issue is removing the dependency on `dnf`.

For more information about the switch to DNF 5, see https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5.

FrostyX:
Fix #2739


rpmbuild/copr-rpmbuild.spec
```diff
@@ -139,8 +139,14 @@ Requires: yum-utils
 %latest_requires ca-certificates
 %latest_requires distribution-gpg-keys
 %if 0%{?fedora}
+
+%if 0%{?fedora} >= 39
+%latest_requires dnf5
```
praiskup, line 144: I'm not sure if we are fully compatible with DNF5.  In Mock and prunerepo we switch to /bin/yum-3 and `python3-dnf`.  Would it make sense to do the same thing here?

FrostyX: I am not sure but, I think we don't care.
The `copr-rpmbuild` doesn't do anything with DNF, this is just `copr-builder` dependency to ensure we have the latest tooling on builders.

So if `mock` can run on a system that has DNF5 (can it?), we should be fine.

praiskup: Ah, you are right.  Then we probably want %latest_requires for both `dnf5` and `python3-dnf`.
The newer chroots will probably prefer DNF5, older DNF4 (`dnf` is misleading, because it may be any of those).

praiskup: The same way we'll probably need `dnf5-plugins` (for builddep, repoquery, etc...)

FrostyX: Note to me: Don't forget on `dnf5-plugins`

FrostyX: Updated, PTAL

rpmbuild/copr-rpmbuild.spec
```diff
@@ -139,6 +139,12 @@ Requires: yum-utils
 %latest_requires ca-certificates
 %latest_requires distribution-gpg-keys
 %if 0%{?fedora}
+
+%if 0%{?fedora} >= 38
+%latest_requires dnf5
+%latest_requires dnf5-plugins
+%endif
+
 %latest_requires dnf
```
praiskup, line 148: Here we want to use `python3-dnf` I think.

FrostyX: Updated

rpmbuild/copr-rpmbuild.spec
```diff
@@ -139,7 +139,15 @@ Requires: yum-utils
...
 %if 0%{?fedora}
+
+%if 0%{?fedora} >= 38
+%latest_requires dnf5
+%latest_requires dnf5-plugins
+%else
 %latest_requires dnf
+%endif
+
+%latest_requires python3-dnf
```
praiskup, line 150: I meant `dnf => python3-dnf`.  The thing is that `dnf` is overloaded (on F39 it is provided DNF5, on F38 it is provided by DNF4).  I don't think the current variant is a problem, but dropping the `%latest_reqiures dnf` seems cleaner.

FrostyX: > I meant dnf => python3-dnf

I know, but the way I changed it, we will have up-to-date `dnf` on F37, which is what we do now, and since F38 there will be only up-to-date `python3-copr` and the new DNF5 packages.

That IMHO sounds better than `dnf => python3-dnf` because that would unnecessarily remove `dnf` from the set of up-to-date packages that we use on our current builders.

praiskup: Well `/bin/dnf-3` is used by Mock 4.1+ anyway - and updating either `dnf5` or `python3-dnf` or both would trigger the automatic update of `dnf` (whatever provides that).

praiskup (approved):

praiskup:
Thank you very much for all the updates!

Issue #2743, opened 2023-05-24
Support redis with authenticate


pkking:
Our redis instance deployed with a password with `requirepass`, in `backend` and `frontend`, it will be great to support configure a password for the redis server.

praiskup:
Help with the patch is welcome!

PR #2762, opened 2023-06-06
add redis auth supported

pkking:
Fix: #2743 
Tested on our testing env, will marked as `Ready` when our prod envs are stable :)

### DIFF
--- a/backend/run/print_queues.py
+++ b/backend/run/print_queues.py
@@ -14,6 +14,7 @@ redis_config = {
     'host': opts['redis_host'],
     'port': opts['redis_port'],
     'db': opts['redis_db'],
+    'password': opts['redis_password'],
 }
 
 for i in range(0, NUM_QUEUES):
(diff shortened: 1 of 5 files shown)


backend/run/print_queues.py
```diff
@@ -14,6 +14,7 @@
     'host': opts['redis_host'],
     'port': opts['redis_port'],
     'db': opts['redis_db'],
+    'password': opts['redis_pwd'],
```
praiskup, line 17: redis_password?

pkking: Sorry for the mess, the option key in configfile is `redis_password`, but in `opts` object attr name is `redis_pwd`, see [code](https://github.com/fedora-copr/copr/pull/2762/files#diff-20fa3ce7e615864b4b1ed869c8f7547b60c693f52cf5be7cf6674b7b460e6ab6R291-R293)

pkking: Make sense :)

praiskup (approved):

PR #2764, opened 2023-06-09
rpmbuild: supported in openeuler

pkking:
This PR make tito build works in openEuler chroot:
1. some packages not exsted in openEuler so we exclude
2. some packages have a different name in openEuler
3. we donot have a dnf5 for now

### DIFF
--- a/rpmbuild/copr-rpmbuild.spec
+++ b/rpmbuild/copr-rpmbuild.spec
@@ -1,16 +1,9 @@
-%if 0%{?fedora} || 0%{?rhel} > 7
 %global __python        %__python3
 %global python          python3
 %global python_pfx      python3
 %global rpm_python      python3-rpm
 %global sitelib         %python3_sitelib
-%else
-%global __python        %__python2
-%global python          python2
-%global python_pfx      python
-%global rpm_python      rpm-python
-%global sitelib         %python_sitelib
-%endif
+
 %global copr_common_version 0.12.1.dev
 
 # do not build debuginfo sub-packages
@@ -49,9 +42,7 @@ BuildRequires: %{python_pfx}-jinja2
 BuildRequires: %{python_pfx}-simplejson
 BuildRequires: python3-backoff >= 1.9.0
 
-%if 0%{?fedora} || 0%{?rhel} > 7
-BuildRequires: argparse-manpage
-%endif
+BuildRequires: /usr/bin/argparse-manpage
 BuildRequires: python-rpm-macros
 
 %if "%{?python}" == "python2"
@@ -73,7 +64,9 @@ Requires: git
 Requires: git-svn
 # for the /bin/unbuffer binary
 Requires: expect
+%if !0%{?openEuler}
 Requires: qemu-user-static
+%endif
 Requires: sed
 
 %if 0%{?fedora} || 0%{?rhel} > 7
@@ -103,32 +96,27 @@ Requires: dnf-yum
 Requires: dnf-utils
 %endif
 # selinux toolset to allow running ansible against the builder
-%if 0%{?fedora}
 Requires: python3-libselinux
 Requires: python3-libsemanage
-%else
-Requires: libselinux-python
-Requires: libsemanage-python
-%endif
+%if 0%{?openEuler}
 # for mock to allow: config_opts['nosync'] = True
 Requires: nosync
-%ifarch x86_64
-# multilib counterpart to avoid: config_opts['nosync_force'] = True
-Requires: nosync(x86-32)
 %endif
 Requires: openssh-clients
 Requires: podman
+%if !0%{?openEuler}
 Requires: pyp2rpm
 Requires: pyp2spec
+Requires: rubygem-gem2rpm
+Requires: scl-utils-build
+Requires: fedora-review >= 0.8
+Requires: fedora-review-plugin-java
+%endif
 # We need %%pypi_source defined, which is in 3-29+
 Requires: python-srpm-macros >= 3-29
 Requires: rpkg
 Requires: rsync
-Requires: rubygem-gem2rpm
-Requires: scl-utils-build
 Requires: tito
-Requires: fedora-review >= 0.8
-Requires: fedora-review-plugin-java
 # yum* to allow mock to build against el* chroots without bootstrap_container
 %if 0%{?rhel}
 Requires: yum
@@ -138,8 +126,6 @@ Requires: yum-utils
 # We want those to be always up-2-date
 %latest_requires ca-certificates
 %latest_requires distribution-gpg-keys
-%if 0%{?fedora}
-
 %if 0%{?fedora} >= 38
 %latest_requires dnf5
 %latest_requires dnf5-plugins
@@ -150,14 +136,10 @@ Requires: yum-utils
 %latest_requires libdnf
 %latest_requires librepo
 %latest_requires libsolv
-%endif
 %latest_requires mock
 %latest_requires mock-core-configs
-%latest_requires redhat-rpm-config
+%latest_requires system-rpm-config
 %latest_requires rpm
-
-# Outdated gem2rpm version may have problems with fetching from rubygems.org
-%latest_requires rubygem-gem2rpm
 
 
 %description -n copr-builder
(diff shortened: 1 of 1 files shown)


praiskup:
EL 8 build fails with:
```
Start: rpmbuild -bs
sh: /usr/bin/python3.6: No such file or directory
error: line 54: Unknown tag: %elif 0
Building target platforms: x86_64
```

rpmbuild/copr-rpmbuild.spec
```diff
@@ -153,12 +161,13 @@ Requires: yum-utils
 %endif
 %latest_requires mock
 %latest_requires mock-core-configs
+%if !0%{?openEuler}
 %latest_requires redhat-rpm-config
+%else
+%latest_requires openEuler-rpm-config
```
praiskup, line 167: Instead of the ifdef, can't we simply require `system-rpm-config`?  On RHEL and Fedora:
```
$ rpm -q --whatprovides system-rpm-config
redhat-rpm-config-252-1.fc38.noarch
```

rpmbuild/copr-rpmbuild.spec
```diff
@@ -282,7 +291,7 @@ install -p -m 755 copr-update-builder %buildroot%_bindir
 )
 
 install -p -m 755 bin/copr-distgit-client %buildroot%_bindir
-%if 0%{?fedora} || 0%{?rhel} > 7
+%if 0%{?fedora} || 0%{?rhel} > 7 || 0%{?openEuler}
```
praiskup, line 294: I think we can bury the EL7 support here.  Can you please drop the condition entirely?

rpmbuild/copr-rpmbuild.spec
```diff
@@ -114,21 +118,25 @@ Requires: libsemanage-python
 Requires: nosync
 %ifarch x86_64
 # multilib counterpart to avoid: config_opts['nosync_force'] = True
+%if !0%{?openEuler}
```
praiskup, line 121: The nosync doesn't work nowadays, https://bugzilla.redhat.com/show_bug.cgi?id=2019329 - and it is perhaps a good time to remove it entirely from our deplist.

rpmbuild/copr-rpmbuild.spec
```diff
@@ -114,21 +118,25 @@ Requires: libsemanage-python
...
+%endif
 Requires: openssh-clients
 Requires: podman
+%if !0%{?openEuler}
 Requires: pyp2rpm
 Requires: pyp2spec
+Requires: rubygem-gem2rpm
+Requires: scl-utils-build
+Requires: fedora-review >= 0.8
+Requires: fedora-review-plugin-java
```
praiskup, line 133: Hm, this seems like resolving deployment-specific preferences.  I don't quite like doing this upstream.

Can you please invent something better; that would allow us to specify additional requirements for our deployment (same applies to your instance, how do you keep your deps maintained?).

rpmbuild/copr-rpmbuild.spec
```diff
@@ -103,7 +107,7 @@ Requires: dnf-yum
 Requires: dnf-utils
 %endif
 # selinux toolset to allow running ansible against the builder
-%if 0%{?fedora}
+%if 0%{?fedora} || 0%{?openEuler}
```
praiskup, line 110: I believe we can drop this `%if` entirely.

rpmbuild/copr-rpmbuild.spec
```diff
@@ -51,6 +51,8 @@ BuildRequires: python3-backoff >= 1.9.0
 
 %if 0%{?fedora} || 0%{?rhel} > 7
 BuildRequires: argparse-manpage
+%elif 0%{?openEuler}
+BuildRequires: python3-argparse-manpage
```
praiskup, line 55: Can we BuildRequire `/usr/bin/argparse-manpage`?

rpmbuild/copr-rpmbuild.spec
```diff
@@ -1,4 +1,4 @@
-%if 0%{?fedora} || 0%{?rhel} > 7
+%if 0%{?fedora} || 0%{?rhel} > 7 || 0%{?openEuler}
```
praiskup, line 1: I'd drop this as well;  let's drop Python2 support with EL7 support.

praiskup:
Thank you for the pull-request!  This is good opportunity to clean the spec file a lot.

rpmbuild/copr-rpmbuild.spec
pkking: Sure thing!
```
[lcr@lcrpc ~]$ rpm -q --provides openEuler-rpm-config|grep system-rpm-config
system-rpm-config = 30-35.oe2303
```

rpmbuild/copr-rpmbuild.spec
pkking: > The nosync doesn't work nowadays, https://bugzilla.redhat.com/show_bug.cgi?id=2019329 - and it is perhaps a good time to remove it entirely from our deplist.

In our case, we are still using `nosync` in our builder but only with `x86_64&aarch64`, so i will keep it for openEuler :)

rpmbuild/copr-rpmbuild.spec
pkking: > Hm, this seems like resolving deployment-specific preferences. I don't quite like doing this upstream.
> 
> Can you please invent something better; that would allow us to specify additional requirements for our deployment (same applies to your instance, how do you keep your deps maintained?).

Its a bit ugly for us to ship the ruby/python generator, [we just install the tools from our SCM directly](https://github.com/opensourceways/copr_docker/blob/c79d3e36b5c9633243ac544c395ead8a3442c3c8/docker/builder/Dockerfile#L63-L67) in Dockerfile

Look into the future can we define some virtual provides for ruby/python generator like `system-release`?

And for `fedora-review` and `scl-utils-build`, it should be optional(tools absensed in openEuler), maybe left as `Recommands` so it can be determined at deploy moment?

rpmbuild/copr-rpmbuild.spec
pkking: Done

rpmbuild/copr-rpmbuild.spec
pkking: Done

rpmbuild/copr-rpmbuild.spec
pkking: Good idea!

rpmbuild/copr-rpmbuild.spec
pkking: > I'd drop this as well; let's drop Python2 support with EL7 support.

Done

rpmbuild/copr-rpmbuild.spec
```diff
@@ -330,7 +315,7 @@ install -p -m 644 copr_distgit_client.py %{buildroot}%{expand:%%%{python}_siteli
 %files -n copr-distgit-client
 %license LICENSE
 %_bindir/copr-distgit-client
-%if 0%{?fedora} || 0%{?rhel} > 7
+%if 0%{?fedora} || 0%{?rhel} > 7 || 0%{?openEuler}
```
praiskup, line 318: This `%if` condition can be removed as well.

rpmbuild/copr-rpmbuild.spec
```diff
@@ -103,32 +96,27 @@ Requires: dnf-yum
...
 Requires: podman
-Requires: pyp2rpm
-Requires: pyp2spec
+%if !0%{?openEuler}
+Suggests: pyp2rpm
+Suggests: pyp2spec
+Suggests: rubygem-gem2rpm
+Suggests: scl-utils-build
+Suggests: fedora-review >= 0.8
+Suggests: fedora-review-plugin-java
```
praiskup, line 113: Heh :) this is a nice misunderstanding.  We really need `Requires` here - but some suggestions on how to solve this kind of deployment-specific dependency requirements in general.

rpmbuild/copr-rpmbuild.spec
```diff
@@ -138,7 +126,7 @@ Requires: yum-utils
 # We want those to be always up-2-date
 %latest_requires ca-certificates
 %latest_requires distribution-gpg-keys
-%if 0%{?fedora}
+%if 0%{?fedora} || 0%{?openEuler}
```
praiskup, line 129: I think this `%if` could be dropped as well.

PR #2765, opened 2023-06-09
backend: move all python-specfile stuff into try-except

praiskup:
The expanded_{epoch,version,reklease} attributes are all dynamic properties that can potentially raise exceptions.


nikromen (approved):
Thanks!

PR #2766, opened 2023-06-09
backend, frontend: add lock mechanism around copr projects for delete action

### LINKED ISSUE #2698: llvm-snapshots copr is missing gpg key

nikromen:
This implements a lock mechanism around the whole project so the delete action can lock its resources and block other actions to work on particular projects until deletion is done. I don't like that the information about locking is implemented via DB thus this means a lot of new requests for DB but I can't think of a better mechanism to pass this info between actions. Suggestions are welcomed 

Fixes #2698

### DIFF
--- a/backend/copr_backend/actions.py
+++ b/backend/copr_backend/actions.py
@@ -8,9 +8,10 @@ import base64
 
 from distutils.dir_util import copy_tree
 from distutils.errors import DistutilsFileError
+from functools import wraps
+from typing import Type
 from urllib.request import urlretrieve
 from copr.exceptions import CoprRequestException
-from requests import RequestException
 from munch import Munch
 
 import modulemd_tools.yaml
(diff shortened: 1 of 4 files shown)


praiskup:
I'm not sure, thinking about this for a week again and again;  can you please re-consider the idea that Frontend is responsible for this?  I.e., not generating the action for backend until the previous action is finished?

PR #2774, opened 2023-06-17
doc: describe how to enlarge btrfs volumes

praiskup:
This might be handy;  while on it, I'm removing the part that relates to resizing backend volume (we now have RAID10 volume, and a separate doc).


FrostyX (approved):

FrostyX:
> This might be handy

Indeed. Thank you.

PR #2775, opened 2023-06-17
frontend: better clone_url matching on webhooks

praiskup:
Previously we just dropped ".git" suffix from git_cloneurl before comparison, but not the trailing slashes.  We also did just 'rstrip(".git")', which would also remove strings like "tig.git", potentially leading false-positive matches.

Fixes: #2722


praiskup:
The CI failure seems to be a different thing: https://github.com/fedora-copr/copr/issues/2776

FrostyX (approved):

Issue #2410, opened 2022-11-30
Every single cgit http request to copr-dist-git easts CPU for several seconds


praiskup:
ssia

praiskup:
My bet is `/etc/cgitrc` which does `include=/var/cache/cgit/repo-configuration.rc`.
In the production that file has 340MB.

praiskup:
I think we should avoid starting 'cgit' binary for every single request.  We can keep the server processes for several minutes, for example ... (for newly imported packages, cgit will probably not work till restarted, but that sounds acceptable).

praiskup:
Looking at some old [CZ school work](https://rtime.felk.cvut.cz/osp/student/mildedan/vysledek.pdf) presentation, it seems cgit isn't (or wasn't by that time) ready for FastCGI use-cases.  Namely it looks like it suffers from memory leaks.

Therefore I think the only option is to use the `MACRO EXPANSION` feature (see man 5 cgitrc).  We can e.g. have one include file per copr owner, or similar.

The problem is how to use it properly.  Not sure if we don't have to have our own cgit CGI wrapper.
Some env vars available for such a script:
```
GIT_PROJECT_ROOT=/var/lib/dist-git/git
GIT_HTTP_EXPORT_ALL=
HTTPS=on
HTTP_HOST=copr-dist-git.fedorainfracloud.org
HTTP_USER_AGENT=Mozilla/5.0 (X11; Linux x86_64; rv:107.0) Gecko/20100101 Firefox/107.0
HTTP_ACCEPT=text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
HTTP_ACCEPT_LANGUAGE=en-US,en;q=0.5
HTTP_ACCEPT_ENCODING=gzip, deflate, br
HTTP_REFERER=https://copr-dist-git.fedorainfracloud.org/cgit/?ofs=715450
HTTP_DNT=1
HTTP_CONNECTION=keep-alive
HTTP_UPGRADE_INSECURE_REQUESTS=1
HTTP_SEC_FETCH_DEST=document
HTTP_SEC_FETCH_MODE=navigate
HTTP_SEC_FETCH_SITE=same-origin
HTTP_SEC_FETCH_USER=?1
HTTP_IF_MODIFIED_SINCE=Mon, 05 Dec 2022 14:03:59 GMT
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin
[...]

PR #2793, opened 2023-06-27
dist-git: make the cgitrc include file smaller

praiskup:
In Fedora Copr, it goes from 550M to 94M.  This minimizes the issues with cgit.  The additional metadata were default, or just stub, so we can live with that.

Also, don't use `cat <<EOF` pattern to avoid unnecessary fork() calls by shell, use echo build-tin.

Fixes: #2410


FrostyX (approved):

sergiomb2:
cool , cgit is working, thanks

PR #2796, opened 2023-06-28
fix: python build failed in openeuler

pkking:


python/python-copr.spec
```diff
@@ -147,6 +147,7 @@ BuildRequires: python3-pytest
 BuildRequires: python3-requests-gssapi
 BuildRequires: python3-future
 BuildRequires: python3-filelock
+BuildRequires: pyproject-rpm-macros
```
praiskup, line 150: This seems redundant.  Per guidelines, we need python3-devel:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/

python/python-copr.spec
```diff
@@ -137,7 +137,7 @@ Requires: python3-future
 
 %{?python_provide:%python_provide python3-copr}
 
-%if 0%{?fedora} > 30 || 0%{?rhel} > 8
+%if 0%{?fedora} > 30 || 0%{?rhel} > 8 || 0%{?openEuler}
```
praiskup, line 140: This seems like an ideal candidate for '%else' branch above to `%if 0%{?rhel} && 0%{?rhel} <= 8`

python/python-copr.spec
praiskup: Hmm, I am not sure why we actually need all the other requirements that should be automatically generated.

python/python-copr.spec
pkking: > This seems like an ideal candidate for '%else' branch above to `%if 0%{?rhel} && 0%{?rhel} <= 8`

Done/

python/python-copr.spec
praiskup: ACK.  I think we can keep it (seems like a distro glitch).

python/python-copr.spec
```diff
@@ -1,6 +1,6 @@
 %global srcname copr
 
-%if 0%{?fedora} || 0%{?rhel} > 7
+%if 0%{?fedora} || 0%{?rhel} > 7 || 0%{?openEuler}
```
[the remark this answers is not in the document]
pkking: > Would you mind using `!0%{?rhel} || 0%{?rhel} > 7` ? Any new distro will directly use python3.

Done

praiskup (approved):

PR #2798, opened 2023-06-29
frontend: add failed-to-succeeded-stats command

### LINKED ISSUE #2779: measure the time between failed and suceeded build
It will be helpful, if I can have a data how much time pass between failed job and first succeeded of the same NEVRA. Quick'n'dirty one shot is fine.

FrostyX:
Fix #2779

### DIFF
--- a/frontend/coprs_frontend/commands/failed_to_succeeded_stats.py
+++ b/frontend/coprs_frontend/commands/failed_to_succeeded_stats.py
@@ -1,0 +1,154 @@
+"""
+Generate failed to succeeded stats
+"""
+
+import os
+from datetime import datetime
+
+import click
+import pygal
+from coprs import models
+
+
+@click.command()
+@click.option("--dest", "-D", required=True, help="Result directory")
+def failed_to_succeeded_stats(dest):
+    """
+    Generate failed to succeeded stats
+    """
+    print("Please wait, this will take at least 20 minutes.")
+    categories = {
+        "immediately": 0,
+        "seconds": 0,
+        "minutes": 0,
+        "hours": 0,
+        "days": 0,
+        "weeks": 0,
+    }
+    tuples = failed_to_succeeded_tuples()
+    for failed, succeeded in tuples:
+        delta = datetime.fromtimestamp(succeeded) \
+            - datetime.fromtimestamp(failed)
+        categories[delta_to_category(delta)] += 1
+
+    os.makedirs(dest, exist_ok=True)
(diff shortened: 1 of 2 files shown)


FrostyX:
> how much time pass between failed job and first succeeded **of the same NEVRA**.

I think the same NEVRA probably isn't what you are interested in? Because I can bump the package version/release while fixing it. 

So the script goes through every package and its builds. If any of them fails, it will find the next successful build of the same package (It doesn't care for NEVRA, it only cares for `package.id`) and compares their `build.ended_on`.

@xsuchy, I didn't exactly remember what you are going to use this for, but I know we talked about average/median time. I think this output might be more helpful. But I can change it any way you need.

I run this against the current production database dump, and the result looks like this:

failed-to-succeeded-stats

Which can be read as "In 367 cases, a new successful build was submitted before the failing one could even finish. 59 failures were fixed in seconds, 9210 builds were fixed within hours after they failed, etc. The time ranges are very much understandable from the script itself. I can add more of them if you need.

praiskup (approved):

xsuchy:
For the record here are few artifacts that Jakub created:

Statistics for Packit user only:
failed-to-succeeded-stats-packit (1)

Statistics for all users for 2021
failed-to-succeeded-stats-2021

Statistics for all users  for 2022
failed-to-succeeded-stats-2021

frontend/coprs_frontend/commands/failed_to_succeeded_stats.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Only recent builds
+    # start = datetime(2022, 1, 1).timestamp()
+    # end = datetime(2023, 1, 1).timestamp()
+    # query = (query
+    #          .filter(models.Build.submitted_on >= start)
+    #          .filter(models.Build.submitted_on <= end))
+
+    # Packit user
+    # query = query.filter(models.Copr.user_id==5576)
```
lbarcziova, line 59: with this, are the builds filtered via the Copr project owner? (or via the submitter of the build?)

nikromen: afaik the `user_id` in Copr model is only project owner since this model represents the project as a whole, not the individual builds. To filter builds according to submitter:

```py
query = query.filter(models.Build.user_id=PACKIT_USER_ID)
```

FrostyX: Exactly as @nikromen is saying.

PR #2800, opened 2023-07-01
Move package NEVR parsing to builder

### LINKED ISSUE #2777: Benchmark python specfile a bit
We are going to process all spec files with python-specfile now. We already count with parse errors, though parsing might be really expensivefor some packages. We should test it on texlive.spec e.g.

### LINKED ISSUE #2790: Move python-specfile dep from backend to builder

FrostyX:
Fix #2790
Fix #2777

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -27,13 +27,12 @@ import pytz
 
 import munch
 from munch import Munch
-from specfile import Specfile
 
 from copr_common.redis_helpers import get_redis_connection
 from copr.v3 import Client
 from copr_backend.constants import DEF_BUILD_USER, DEF_BUILD_TIMEOUT, DEF_CONSECUTIVE_FAILURE_THRESHOLD, \
     CONSECUTIVE_FAILURE_REDIS_KEY, default_log_format
-from copr_backend.exceptions import CoprBackendError, CoprBackendSrpmError
+from copr_backend.exceptions import CoprBackendError
 
 from . import constants
 
@@ -633,22 +632,6 @@ def get_chroot_arch(chroot):
     return chroot.rsplit("-", 2)[2]
 
 
-def find_spec_file(folder, logger):
-    """
-    Returns spec file path or None
-    """
-    spec_pattern = os.path.join(folder, "*.spec")
-    result = glob.glob(spec_pattern)
-    if not result:
-        return None
-
-    if len(result) > 1:
-        logger.error("More than one spec file found! Using the first %s from %s",
-                     str(result[0]), str(result))
-
-    return result[0]
-
-
 def format_evr(epoch, version, release):
     """
     Return evr in format (epoch:)version-release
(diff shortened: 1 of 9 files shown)


FrostyX:
Some of the backend tests are failing. That's because we need to add `results.json` file here
https://github.com/fedora-copr/test-data-copr-backend/tree/main/build_results/00848963-example

```json
{
    "name": "example",
    "epoch": null,
    "version": "1.0.14",
    "release": "1",
    "exclusivearch": [],
    "excludearch": []
}
```

The problem is, this test directory is for RPM build, not SRPM. We are only abusing it for testing SRPM builds as well. So this will require a bit more changes. But the code itself works fine.

backend/copr_backend/helpers.py
```diff
@@ -659,45 +642,6 @@ def format_evr(epoch, version, release):
...
-        logger.exception("Exception appeared during handling spec file: %s", spec_file_path)
-        return pkg_name_evr_from_srpm(srpm_path, logger)
-
-
-
-def pkg_name_evr_from_srpm(srpm_path, logger):
-    """
-    Queries a package for its name and evr (epoch:version-release)
-    """
-    cmd = ['rpm', '-qp', '--nosignature', '--qf',
```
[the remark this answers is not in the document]
FrostyX: I don't know :-).
But we should cover it in tests for sure.

praiskup:
+1, but these two are worth fixing
```
Error: PYLINT_WARNING:
copr_backend/background_worker_build.py:634:13: W1514[unspecified-encoding]: BuildBackgroundWorker._get_srpm_build_details: Using open without explicitly specifying an encoding

Error: PYLINT_WARNING:
copr_backend/helpers.py:35: W0611[unused-import]: Unused CoprBackendSrpmError imported from copr_backend.exceptions
```

FrostyX:
Fixed linter errors and tests, PTAL

backend/copr_backend/helpers.py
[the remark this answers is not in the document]
praiskup: That was kind of a safety belt (I still don't trust python-specfile that much, also - newly it will be run on multiple architectures).  But I won't argue too much on this :-)

rpmbuild/copr-rpmbuild.spec
```diff
@@ -40,9 +40,14 @@ BuildRequires: %{python_pfx}-munch
 BuildRequires: %{python}-requests
 BuildRequires: %{python_pfx}-jinja2
 BuildRequires: %{python_pfx}-simplejson
-BuildRequires: %{python_pfx}-specfile
 BuildRequires: python3-backoff >= 1.9.0
 
+%if 0%{?rhel} == 8
+BuildRequires: %{python_pfx}-specfile
```
praiskup, line 46: I submited the issue https://github.com/packit/specfile/issues/255 for this.

praiskup: I tested a bit specfile v0.20 on epel8 and wasn't able to make it work. So I don't think there's an easy way to update specfile into EPEL8 (missing setup.py, incompatible (too new) typing, I stopped there...).

The question is - is the python-specfile in epel 8 "new enough" for us (100% working, but slow), or does this actually mean we have to drop the support for EPEL8?

praiskup: Getting to EPEL8 https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-8a36e6e848 so we can probably drop the %if/%else

praiskup:
@FrostyX @nikromen can you give +1 to https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2023-8a36e6e848?

FrostyX:
I added a simple beaker test and the `rpm -qip` fallback.
Also, CI fails for rawhide but it is not related to this PR. Even `main` branch fails.

praiskup (approved):

PR #2812, opened 2023-07-17
python: remove raise_from from future package

### LINKED ISSUE #2808: Drop dependency on `future`
Python `future` package is seriously outdated and broken on Python 3.12. We managed to fix it in Fedora for 3.12 but this won't be possible for 3.13 because it depends heavily on lib2to3 which will be removed from 3.13. It'd be best to remove it from dependencies rather sooner than later.

AFAIK, the only usage of future in this repo is the `raise_from` function there:
https://github.com/fedora-copr/copr/blob/46ae36775bc6bfe5f13894f8659446b1a9fdc8e5/python/copr/v3/auth/gssapi.py#L43

If you really need it, it might be a good idea to copy the implementation and drop the dependency, see https://github.com/PythonCharmers/python-future/blob/3dc7acc8e5a4266df6c823b9350a731dd7df6124/src/future/utils/__init__.py#L390

nikromen:
Fixes #2808

### DIFF
--- a/python/copr/v3/auth/gssapi.py
+++ b/python/copr/v3/auth/gssapi.py
@@ -9,7 +9,6 @@ try:
 except ImportError:
     requests_gssapi = None
 
-from future.utils import raise_from
 from copr.v3.exceptions import CoprAuthException
 from copr.v3.requests import munchify, handle_errors
 from copr.v3.auth.base import BaseAuth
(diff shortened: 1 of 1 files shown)


praiskup:
> future package is outdated and won't work in python 3.13

Hm, I thought we already have a problem with 3.12 in Rawhide?

praiskup (approved):
Reading better the original ticket, and you are right. LGTM

PR #2813, opened 2023-07-17
backend: log task dict returned from redis to better debugging

nikromen:
To debug cases like #2315 in the future (if they will happen again)


praiskup:
Is this still happening in logs?  If not, I think we can mark the originl issue as "works for me" :-) because logging the large dictionaries would produce ugly and large logs.

praiskup (approved):

PR #2804, opened 2023-07-10
Wait for delete actions before forking

nikromen:
Fixes #2698

PR #2818, opened 2023-07-18
packit: wait for python-copr-common to build first

nikromen:
e.g. in #2804 python-copr-common has to be built before FE so FE build can pass

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -13,6 +13,9 @@ packages:
     files_to_sync:
       - python-copr.spec
     upstream_tag_template: python-copr-{version}
+    actions:
+      post-upstream-clone:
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -98,6 +99,7 @@ packages:
     upstream_tag_template: copr-cli-{version}
     actions:
       post-upstream-clone:
+        - sh -c "wait-for-copr --owner $COPR_OWNER --project $COPR_PROJECT python-copr-common `git rev-parse --short HEAD`"
```
praiskup, line 102: Shouldn't python-copr wait for python3-copr-common instead?

nikromen: yep, you are right

praiskup (approved):
LGTM

PR #2821, opened 2023-07-21
cli: make build works in openEuler

pkking:


cli/copr-cli.spec
```diff
@@ -1,4 +1,4 @@
-%if 0%{?rhel} > 7 || 0%{?fedora}
+%if 0%{?rhel} > 7 || 0%{?fedora} || 0%{?openEuler}
```
praiskup, line 1: I think we can revert the if-else again like `%if 0%{?rhel} && 0%{?rhel} <= 7`. Every other distro will be python3 compliant.

pkking: @praiskup Done

praiskup (approved):
thank you

praiskup (approved):

PR #2822, opened 2023-07-23
backend: update tests to work with new test-data tarball

FrostyX:
https://github.com/fedora-copr/test-data-copr-backend/pull/1


praiskup:
This seems to be CI blocker, but not sure why CI still fails. Do we need a new release for the data tarball?

praiskup:
Thank you!

Issue #2825, opened 2023-07-28
Do we support multi group alias for one group or not?


pkking:
Opened the issue because im confused by these [code](https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py#L32C1-L49C67):
```
        if fas_group not in flask.g.user.user_teams:
            raise InsufficientRightsException(
                "User '{}' doesn't have access to fas group {}"
                .format(flask.g.user.username, fas_group))

        alias = form.name.data
        group = UsersLogic.get_group_by_fas_name_or_create(
            fas_group, alias)

        db.session.add(group)
        db.session.commit()

        flask.flash(
            "FAS group {} is activated in the Copr under the alias {} "
            .format(fas_group, alias)
        )
        return flask.redirect(url_for(
            "groups_ns.list_projects_by_group", group_name=alias))
```
The code above will allow user submit more than one alias for a group, but if the group already has a alias, it just will always return it instead of create another alias, so the code behave is wired:
1. if we allow multi alias for one group, the maybe we should make a new method `get_group_by_fas_name_and_alias_or_create` to both match the alias and group name
2. if just one alias is allowed, the redirect should be using `group.name` instead of  `alias`

praiskup:
There's just 1:1 mapping, and `if form.validate_on_submit()` does some checking.  There's though a race condition that we should fix, perhaps by installing unique key https://github.com/fedora-copr/copr/blob/7442ec4303f2e0ba2e31a6555c4da34b96de0ee1/frontend/coprs_frontend/coprs/models.py#L2210

pkking:
Otherwise, the `form.validate_on_submit()` just check if the alias already existed, but if we try to add another alias for a group, it will try to redirect to `/group/g/<new alias>/coprs` which will return 404, i think we should just redirect to  `/group/g/<existed alias>/coprs`.

Issue #2563, opened 2023-03-04
LDAP is used to support Copr groups, but UI links to FAS


praiskup:
![Screenshot_20230304_163611](https://user-images.githubusercontent.com/3277839/222915038-420d7964-f60b-4c64-8985-7d4d0ad7f0c0.png)

Thse links point to FAS, like https://accounts.fedoraproject.org/group/cpt-team

PR #2826, opened 2023-07-29
group alias need 1:1 mapping to group

pkking:
This PR try to fix: #2825 also fortunately fix: #2563
1. add unique contraint to `fas_name` and add migration script generated by ` alembic revision --autogenerate `
2. I think  macro `fas_group_href` canbe replaced by `onfig.GROUP_INFO.link.forma` which is more configurable
3. When creating a new alias for a group, if the group already has one alias, popup a flash and redirect to the already existed group

### DIFF
--- a/frontend/coprs_frontend/coprs/templates/groups/user_fas_groups.html
+++ b/frontend/coprs_frontend/coprs/templates/groups/user_fas_groups.html
@@ -1,5 +1,5 @@
 {% extends "layout.html" %}
-{% from "_helpers.html" import fas_group_href, initialize_datatables %}
+{% from "_helpers.html" import initialize_datatables %}
 {% block title %}Add a Group{% endblock %}
 {% block header %}Add a Group{% endblock %}
 
@@ -37,7 +37,7 @@
     <tbody>
     {% for team in teams %}
       <tr>
-        <td><a href="{{ fas_group_href(team) }}">
+        <td><a href="{{ config.GROUP_INFO.link.format(name=team) }}">
           {{ team }}
         </a></td>
         <td>
--- a/frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
+++ b/frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
@@ -10,6 +10,7 @@ from coprs.logic.complex_logic import ComplexLogic
 from coprs.logic.coprs_logic import CoprsLogic, PinnedCoprsLogic
 from coprs.logic.users_logic import UsersLogic
 from coprs import app
+from coprs.oidc import oidc_enabled
 
 from ... import db
 from ..misc import login_required
@@ -21,6 +22,15 @@ from . import groups_ns
 @groups_ns.route("/activate/<fas_group>", methods=["GET", "POST"])
 @login_required
 def activate_group(fas_group):
+    msg_fmt = "Group {} is activated in the system under the alias {}"
+
+    group = UsersLogic.get_group_by_fas_name(fas_group).first()
+    if group:
+         # can't have more than one alias for a group
+        flask.flash(msg_fmt.format(group.fas_name, group.name), "success")
+        return flask.redirect(url_for(
+            "groups_ns.list_projects_by_group", group_name=group.name))
+
     form = ActivateFasGroupForm()
 
     if form.validate_on_submit():
@@ -31,22 +41,17 @@ def activate_group(fas_group):
 
         if fas_group not in flask.g.user.user_teams:
             raise InsufficientRightsException(
-                "User '{}' doesn't have access to fas group {}"
+                "User '{}' doesn't have access to group {}"
                 .format(flask.g.user.username, fas_group))
 
         alias = form.name.data
-        group = UsersLogic.get_group_by_fas_name_or_create(
-            fas_group, alias)
+        group = UsersLogic.create_group_by_fas_name(fas_group, alias)
 
-        db.session.add(group)
         db.session.commit()
 
-        flask.flash(
-            "FAS group {} is activated in the Copr under the alias {} "
-            .format(fas_group, alias)
-        )
+        flask.flash(msg_fmt.format(group.fas_name, group.name), "success")
         return flask.redirect(url_for(
-            "groups_ns.list_projects_by_group", group_name=alias))
+            "groups_ns.list_projects_by_group", group_name=group.name))
 
     else:
         return flask.render_template(
@@ -85,7 +90,7 @@ def list_projects_by_group(group_name, page=1):
 @groups_ns.route("/list/my")
 @login_required
 def list_user_groups():
-    if not (app.config['FAS_LOGIN'] or app.config['LDAP_URL']):
+    if not (app.config['FAS_LOGIN'] or app.config['LDAP_URL'] or oidc_enabled(app.config)):
         raise ObjectNotFound("Fedora Accounts or LDAP groups not enabled")
 
     teams = flask.g.user.user_teams
--- a/frontend/coprs_frontend/alembic/versions/daa62cd0743d_add_unique_constraint_to_fas_group.py
+++ b/frontend/coprs_frontend/alembic/versions/daa62cd0743d_add_unique_constraint_to_fas_group.py
@@ -1,0 +1,24 @@
+"""
+add unique constraint to fas_group
+
+Revision ID: daa62cd0743d
+Revises: ba6ac0936bfb
+Create Date: 2023-08-01 09:52:01.522171
+"""
+
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = 'daa62cd0743d'
+down_revision = 'ba6ac0936bfb'
+branch_labels = None
+depends_on = None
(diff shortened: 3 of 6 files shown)


frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
```diff
@@ -41,12 +42,16 @@ def activate_group(fas_group):
...
-            "FAS group {} is activated in the Copr under the alias {} "
-            .format(fas_group, alias)
-        )
+        # can't have more than one alias for a group
+        if group.name != alias:
+            message = "Group {} already activated as alias {} ".format(
+                fas_group, group.name)
+        else:
+            message = "Group {} is activated in the System under the alias {} ".format(
+                fas_group, alias)
```
praiskup, line 51: :-) how do you trigger this one actually?  Is there a reproducer?

frontend/coprs_frontend/coprs/templates/groups/user_fas_groups.html
```diff
@@ -37,7 +37,7 @@ <h1>
     <tbody>
     {% for team in teams %}
       <tr>
-        <td><a href="{{ fas_group_href(team) }}">
+        <td><a href="{{ config.GROUP_INFO.link.format(name=team) }}">
```
praiskup, line 40: Is this #2563?

praiskup:
Nit that I can fix myself for the PR - but would be helpful to write  commit messages like:
```
<component>: <what you are doing>

_Why_ (rather how, it's the commit itself) are doing that. This should be readable and
describe the background story.

<tag>: value (e.g. Fixes: #XYZ)
<tag>: value
```

frontend/coprs_frontend/coprs/templates/groups/user_fas_groups.html
pkking: > Is this #2563?

Don‘t know this ticket before :) but in our case, we should be redirect to openEuler group info page instead of `fas_group` stuff, so i think fas_group_href can be replaced with `GROUP_INFO.link` 😄

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
pkking: > :-) how do you trigger this one actually? Is there a reproducer?

Yeah, i just go `/groups/activate/<group_name>` twice with two different alias
The second submit will lead to a 404 page

praiskup: AAAh, this. Yeah, I totally misunderstood the thing you want to fix (I though you want to map 2 external groups to a single Copr group, and it is vice versa).

praiskup:
> alembic revision --autogenerate

This is correct action to add the index to our existing databases;  can you add the migration script into this PR?

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
praiskup: Could you please return "400 Bad Request" if user visits `/groups/activate/<existing_group>` for the second time?  There's no such link, so that must be an error anyway (processing the request seems just redundant).

frontend/coprs_frontend/coprs/templates/groups/user_fas_groups.html
praiskup: Yeah, I think that's exactly #2563 (forgotten hard-coded link).  Can you use `Fixes: #2563`, too?

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
pkking: Will simplely return 400 be a good experience?
Or just redirect user to the `existed group`?

pkking:
> > alembic revision --autogenerate
> 
> This was a correct action (adding the index via migration); can you add the migration script into this PR?

Ah i forget to push this file 🐷, Done

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
praiskup: Not sure.  Do you think users can get to the route once it is activated?   If yes, redirecting is indeed a better option.

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
```diff
@@ -41,12 +42,16 @@ def activate_group(fas_group):
...
         db.session.commit()
 
-        flask.flash(
-            "FAS group {} is activated in the Copr under the alias {} "
-            .format(fas_group, alias)
-        )
+        # can't have more than one alias for a group
+        if group.name != alias:
+            message = "Group {} already activated as alias {} ".format(
+                fas_group, group.name)
```
praiskup, line 48: If we go with error 400 or redirect, this `if/else` condition is redundant.  While on this, would you mind making the flash message "green" with `flash("message", "success")` :-) so it is more visible?  The grey color doesn't look like something actually happened...

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
pkking: > Not sure. Do you think users can get to the route once it is activated? If yes, redirecting is indeed a better option.

Seems no possible unless they know the route and play with it like me :)
But return 400 or redirect will not waste anything, so why not?

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
pkking: Done

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
```diff
@@ -41,12 +42,12 @@ def activate_group(fas_group):
...
-            "FAS group {} is activated in the Copr under the alias {} "
-            .format(fas_group, alias)
-        )
+        # can't have more than one alias for a group
+        message = "Group {} is activated in the System under the alias {} ".format(
+            fas_group, alias)
+        flask.flash(message, "success")
         return flask.redirect(url_for(
-            "groups_ns.list_projects_by_group", group_name=alias))
+            "groups_ns.list_projects_by_group", group_name=group.name))
```
praiskup, line 50: This is certainly OK and better behavior than before, thank you!

What I thought meant before was that we should redirect the user (or error 400) much earlier, before the user starts even thinking about how to name the group (it is wasting of time, the group is already activated so nothing will happen).  Something like:

```python
def activate_group(fas_group):
    if "fas group is activated":
        return flask.redirect(...)
```

praiskup:
+1 (note the missing thing above, but still good enough)

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
pkking: That's sounds a good idea, will tweak the code :)

pkking: > This is certainly OK and better behavior than before, thank you!
> 
> What I thought meant before was that we should redirect the user (or error 400) much earlier, before the user starts even thinking about how to name the group (it is wasting of time, the group is already activated so nothing will happen). Something like:
> 
> ```python
> def activate_group(fas_group):
>     if "fas group is activated":
>         return flask.redirect(...)
> ```

Done

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
```diff
@@ -31,22 +41,18 @@ def activate_group(fas_group):
...
                 .format(flask.g.user.username, fas_group))
 
         alias = form.name.data
-        group = UsersLogic.get_group_by_fas_name_or_create(
-            fas_group, alias)
-
-        db.session.add(group)
+        group = UsersLogic.create_group_by_fas_name(fas_group, alias)
+        # db.session.add(group) is done by UsersLogic.create_group_by_fas_name()
+        db.session.flush()
```
praiskup, line 50: I bet `flush()`  is not needed.  I'd drop the commented-out line with `db.session.add()` (NB adding the same object twice shouldn't hurt).

pkking: You are right, `commit()` will unconditionally call `flush()` before commit

praiskup: Yeah, `add()` is redundant, I have no objections to removing it.

pkking: > Yeah, `add()` is redundant, I have no objections to removing it.

I totally misunderstand you :)
Done

praiskup (approved):
thank you

praiskup:
NOTE: Needs manual merge.

praiskup:
Merged as: bddee02fb014898075b69e058f1d8fe8a0d13dbf

Thank you.

PR #2830, opened 2023-08-04
packit: yamllint

praiskup:


FrostyX (approved):
AFAIK YAML considers both `true` and `True` to be true. And AFAIK `---` is optional, so none of the changes should IMHO be required. That being said, there is no problem with them.

PR #2835, opened 2023-08-06
Specify disttag when running fedora-review

FrostyX:
Fix https://github.com/FrostyX/fedora-review-service/issues/28
Fix https://pagure.io/FedoraReview/issue/486

The `fedora-review` commands fails with the following error:

    ERROR: 'No disttag in package and no DISTTAG flag.
    Use --define DISTTAG to set proper dist e. g., --define DISTTAG=fc21.'

It is strange to me why it cannot find any disttag in the package but
it is easy for us to specify it, so why not.


rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -10,6 +10,7 @@
 
 import os
 import shutil
+from copr_common.helpers import chroot_to_branch
```
praiskup, line 13: +1 for moving the method

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -39,10 +40,12 @@ def run(self):
         """
         Run `fedora-review` tool inside the `resultdir`
         """
+        branch = chroot_to_branch(self.task["chroot"])
         cmd = [
             "fedora-review", "--no-colors", "--prebuilt", "--rpm-spec",
             "--name", self.package_name,
             "--mock-config", self.mock_config_file,
+            "--define", "DISTTAG={0}".format(branch),
```
praiskup, line 48: This must be some bug in fedora-review, what am I missing?  Is the package actually having the dist tag or not? (this must be bothering normal fedora-review users, not just Copr)

FrostyX:
I made this way harder than it needed to be.

- The `ERROR: 'No disttag in package and no DISTTAG flag.` was not the real issue. It only happened to me locally because of a mess in my `/var/lib/copr-rpmbuild/results/`. This is not why it fails in produciton.
- However, the error handling is incorrect and the `self._filter_results_directory()` with some modification needs to be called even if the `fedora-review` command fails
- The real issue from production is `Cache-only enabled but no cache for repository "copr_base"` which is related to DNF5. Therefore it should get fixed by using the up-to-date fedora-review package.

rpmbuild/copr_rpmbuild/automation/fedora_review.py
FrostyX: This change is no longer needed because we won't use `chroot_to_branch` in `copr-rpmbuild` after all. But since you +1'd it, I left the commit there.

praiskup (approved):
Thank you!

praiskup (approved):
LGTM

Issue #2788, opened 2023-06-26
Support groups info through OIDC


pkking:
As we have a basic support for OIDC, there's another feature we need so that we do not lost any functions when switching to OIDC.
Copr should have ability to fetch user's group info from Idp.

As `group info` is not covered by OIDC spec, but most of OIDC providers(etc google/auth0) support [custom claim](https://community.auth0.com/t/group-scope-in-oidc/21317/5)

I also talk to openEuler infra team, they can add the a group scope in the OIDC provider but i have no idea what fedora OIDC provider can do.

So here's my idea: add a config item: OIDC_type(such as: google/openEuler/fedora. etc), for each type, we can have a specified impl to grab the group info

Now i'd like hear from you and what you say about this proposal :)

praiskup:
@FrostyX has took a look on what group support in Fedora OIDC, and so far no success, work in progress.

pkking:
Hi, @FrostyX @praiskup , we have some experience on `OIDC group info` before because we try to get it into our jenkins cluster, in jenkins story, our OIDC `userinfo` endpoint will return a list of groups if a `group` scope is provided such as:
```json
{
  "sub": "user123",
  "name": "John Doe",
  "email": "john.doe@example.com",
  "groups": [
    "dev-utils",
    "eBPF",
    "kernel"
  ]
}
```
Im trying to find a way to let `authlib` to work in this way :)

pkking:
seems https://github.com/lepture/authlib/issues/549 will works like a charm

pkking:
I think we should wait fedora guys to finish Fedora OIDC adaptation, the #2836 works well in openEuler now :)

praiskup:
@nikromen fyi, related to #3483

nikromen:
luckily, fedora folks implemented custom UserInfo fields with the same field name `groups`, so groups works with fedora oidc. Tested on stg (it's up and running there so you can try for yourself at https://copr.stg.fedoraproject.org)

PR #2836, opened 2023-08-07
frontend: support grab group info from OIDC provider

pkking:
As our former OIDC PR #2422 didn't have ability to grab any group info from Idp,
This PR add the missing functionalities:
1. unitfied the update group API, so it can be use by LDAP, openID and OIDC
2. suppose our OIDC Idp can return sort of group info like
    ```
    {
      "sub": "user123",
      "name": "John Doe",
      "email": "john.doe@example.com",
      "groups": [
        "dev-utils",
        "eBPF",
        "kernel"
      ]
    }
    ```
    PS: to achieve this, our Idp implement a non-standard but defacto standard: `group` scope and we add it in the `OIDC_SCOPES` config item
3. some refactor for oidc module: make `oidc_enabled` more useful, which will guarantee the OIDC related configs is configured and valid

try to fix: #2788

### DIFF
--- a/frontend/coprs_frontend/coprs/auth.py
+++ b/frontend/coprs_frontend/coprs/auth.py
@@ -92,37 +92,26 @@ class GroupAuth:
     `app.config["FAS_LOGIN"]` and `app.config["KRB5_LOGIN"]` should be
     encapsulated within this class.
     """
-
     @classmethod
-    def update_user_groups(cls, user, oid_resp=None):
+    def update_user_groups(cls, user, groups=None):
         """
         Upon a successful login, try to (a) load the list of groups from
         authoritative source, and (b) (re)set the user.openid_groups.
         """
-
         def _do_update(user, grouplist):
             user.openid_groups = {
                 "fas_groups": grouplist,
             }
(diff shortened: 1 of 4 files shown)


praiskup:
This is sooo hard to review :-) I'm not sure about the ordering now at all.  I need to try all the auth methods first before +1.

How exactly do you read the "groups" field from the OIDC dict?

pkking:
> This is sooo hard to review :-) I'm not sure about the ordering now at all. I need to try all the auth methods first before +1.
> 
> How exactly do you read the "groups" field from the OIDC dict?

The groups field should (which as is in our OIDC Idp) be a list object like:
```
"groups": [
    "dev-utils",
    "eBPF",
    "kernel"
  ]
```

praiskup:
Ok, I think I'm +1 here because it doesn't break any of our use-cases. But it is breaking the abstraction in GroupAuth, which should be "generic" - and should allow us to use LDAP groups even with OIDC.  I'm going to fix this in the future, but for now good.

praiskup (approved):

Issue #2816, opened 2023-07-17
`task_data` passed to `BuildJob` is really often `None` type


nikromen:
worker.log on backend has a ton of these:

```
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 793, in handle_task
    self.handle_build()
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 785, in handle_build
    self._get_build_job()
  File "/usr/lib/python3.11/site-packages/copr_backend/background_worker_build.py", line 350, in _get_build_job
    self.job = BuildJob(resp.json(), self.opts)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/copr_backend/job.py", line 77, in __init__
    for key, val in task_data.items():
                    ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'

```

nikromen:
note for me so I don't forget

these exceptions happen roughly at the same time on frontend
```
2023-08-08 12:42:29,487 [ERROR][/usr/share/copr/coprs_frontend/coprs/views/backend_ns/backend_general.py:187|backend_general:get_build_record][backend: 2600:1f18:8ee:ae00:d553:8ed5:d8b6:9f83] 'NoneType' object has no attribute 'module_setup_commands'
Traceback (most recent call last):
  File "/usr/share/copr/coprs_frontend/coprs/views/backend_ns/backend_general.py", line 169, in get_build_record
    modules = copr_chroot.module_setup_commands
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 ```
 
this exceptions happen in `get_build_record` which provides info about `/get-build-task/` endpoint, but `get_build_record` throws an exception, logs it and then returns None -> `/get-build-task/` returns None -> `resp.json()` in `self.job = BuildJob(resp.json(), self.opts)` returns "null"

FrostyX:
I am sorry, I forgot about this issue and canceled the broken build, see #2904. So I accidentally got rid of your reproducer. But AFAIK we know how to reproduce it locally, so hopefully it should be fine.

praiskup:
praiskup to help jkyjovsk with reproducing

PR #2842, opened 2023-08-08
frontend: return empty dict instead of null in API calls to BE

nikromen:
Fixes: #2816

### DIFF
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -5,8 +5,7 @@ from coprs import models
 from coprs.logic import actions_logic
 from coprs.logic.builds_logic import BuildsLogic
 from coprs.logic.complex_logic import ComplexLogic, BuildConfigLogic
-from coprs.logic.packages_logic import PackagesLogic
-from coprs.logic.coprs_logic import MockChrootsLogic, CoprChrootsLogic
+from coprs.logic.coprs_logic import MockChrootsLogic
 from coprs.exceptions import CoprHttpException, ObjectNotFound
 from coprs.helpers import streamed_json
 
@@ -171,7 +170,7 @@ def get_build_record(task, for_backend=False):
             "repo_priority": task.build.copr.repo_priority
         })
 
-        copr_chroot = CoprChrootsLogic.get_by_name_safe(task.build.copr, task.mock_chroot.name)
+        copr_chroot = ComplexLogic.get_copr_chroot_safe(task.build.copr, task.mock_chroot.name)
         modules = copr_chroot.module_setup_commands
         if modules:
             build_record["modules"] = {'toggle': modules}
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -136,61 +136,55 @@ def get_build_record(task, for_backend=False):
     if not task:
         return None
 
-    build_record = None
-    try:
-        build_record = {
-            "task_id": task.task_id,
-            "build_id": task.build.id,
-            "project_owner": task.build.copr.owner_name,
-            "sandbox": task.build.sandbox,
-            "background": bool(task.build.is_background),
-            "chroot": task.mock_chroot.name,
-            "tags": task.mock_chroot.tags + task.tags,
-        }
+    build_record = {
+        "task_id": task.task_id,
+        "build_id": task.build.id,
+        "project_owner": task.build.copr.owner_name,
+        "sandbox": task.build.sandbox,
+        "background": bool(task.build.is_background),
+        "chroot": task.mock_chroot.name,
+        "tags": task.mock_chroot.tags + task.tags,
+    }
 
-        if for_backend:
-            return build_record
+    if for_backend:
+        return build_record
 
-        build_record.update({
-            "project_name": task.build.copr_name,
-            "project_dirname": task.build.copr_dirname,
-            "submitter": task.build.submitter[0],
-            "repos": task.build.repos,
-            "memory_reqs": task.build.memory_reqs,
-            "timeout": task.build.timeout,
-            "enable_net": task.build.enable_net,
-            "git_repo": task.distgit_clone_url,
-            "git_hash": task.git_hash,
-            "package_name": task.build.package.name,
-            "package_version": task.build.pkg_version,
-            "uses_devel_repo": task.build.copr.devel_mode,
-            "isolation": task.build.isolation,
-            "fedora_review": task.build.copr.fedora_review,
-            "appstream": bool(task.build.appstream),
-            "repo_priority": task.build.copr.repo_priority
-        })
+    build_record.update({
+        "project_name": task.build.copr_name,
+        "project_dirname": task.build.copr_dirname,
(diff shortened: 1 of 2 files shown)


frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -354,6 +354,9 @@ def get_build_task(task_id):
     try:
         task = BuildsLogic.get_build_task(task_id)
         build_record = get_build_record(task)
+        if build_record is None:
```
nikromen, line 357: I can see that `get_build_record` and `get_srpm_build_record` are only used in API endpoint. Can we perhaps change return type of these to `dict` from `Optional[dict]`? Backend thinks these endpoints won't return `null` but they do

praiskup: I can't reproduce the `null` output, I can't decide.  It seems completely OK to return 404 if no such build task exists, 500 upon error, 400 for invalid requests.  And return 200 if and only if the request was processed.

praiskup:
I'm for fixing the `Backend <-> Frontend` protocol, though this rather seems like a `status 200` by mistake, instead of `status 404`.  These additional checks seem pretty complicated to validate.

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
nikromen: the build that causes this error is https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5942294/

namely, the f36-s390x chroot, as you can see there the chroot for this build does not exist... it looks like it was never created in DB for this build - I can see that this build was created at the same time I made f36 chroots EOL so some rare unlucky race condition perhaps :D

Until that time the build is still in a pending state and it is flooding our logs with errors (cca 2 months).

nikromen:
I fix the `null` output on the frontend API endpoints side - but the race condition with EOL and new builds remains... it seems like a really rare accident so I think it's not worth spending time on it, let's see if it will occur in next EOL session

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
praiskup: Hmmm, that would make sense, good catch.

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -354,6 +354,14 @@ def get_build_task(task_id):
...
+            jsonout = flask.jsonify(
+                {"msg": "Internal server error happened during searching"
+                        f" record for task {task_id} was not found"}
+            )
+            jsonout.status_code = 500
```
praiskup, line 362: Can we use error 400?  And I'm still not quite sure this is ideal, can't we raise this error directly in `get_build_task()`?

nikromen: OK, I used the already existing `get_copr_chroot_safe` method which returns 404, and changed the `except: return None` to re-raise - hopefully backend can handle this :D

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -171,7 +170,7 @@ def get_build_record(task, for_backend=False):
             "repo_priority": task.build.copr.repo_priority
         })
 
-        copr_chroot = CoprChrootsLogic.get_by_name_safe(task.build.copr, task.mock_chroot.name)
+        copr_chroot = ComplexLogic.get_copr_chroot_safe(task.build.copr, task.mock_chroot.name)
```
praiskup, line 173: This deserves a note in the commit message. :-)

praiskup:
Hm, while it looks like it should work, it doesn't seem to.

I just tested locally, and I'm still getting:
```
ERROR:coprs:'NoneType' object has no attribute 'module_setup_commands'
  File "/copr/coprs_frontend/coprs/views/backend_ns/backend_general.py", line 175, in get_build_record
    modules = copr_chroot.module_setup_commands
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'module_setup_commands'
```

And backend gets:
```
$ curl localhost:55555/backend/get-build-task/95628-rhel-9-x86_64
null
```

nikromen:
I did an eol event during SRPM build (to simulate https://github.com/fedora-copr/copr/pull/2842#discussion_r1291348851) to test the `Backend <-> Frontend` protocol and here's what I got:

FE:
```
2023-09-19 15:26:40,070 [INFO][/opt/copr/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py:357|backend_general:_stream][backend: 10.89.0.13] Generating RPM builds
2023-09-19 15:26:40,502 [ERROR][/opt/copr/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py:192|backend_general:get_build_record][backend: 10.89.0.13] Chroot name fedora-38-x86_64 does not exist in Copr nikromen/dalsi-novy.
  File "/opt/copr/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py", line 173, in get_build_record
    copr_chroot = ComplexLogic.get_copr_chroot_safe(task.build.copr, task.mock_chroot.name)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/copr/frontend/coprs_frontend/coprs/logic/complex_logic.py", line 289, in get_copr_chroot_safe
    raise ObjectNotFound(
coprs.exceptions.ObjectNotFound: Chroot name fedora-38-x86_64 does not exist in Copr nikromen/dalsi-novy.
```

BE:
```
[2023-09-19 15:31:40,518][ ERROR][PID:353][/usr/bin/copr-backend-process-build.managed.pid-353][background_worker_build.py:_get_build_job:342] Failed to download build info: Request client error on http://frontend:5000/backend/get-build-task/27-fedora-38-x86_64/: 404 NOT FOUND
[2023-09-19 15:31:40,518][ ERROR][PID:353][/usr/bin/copr-backend-process-build.managed.pid-353][background_worker_build.py:handle_task:793] Backend process error: Failed to get the build task get-build-task/27-fedora-38-x86_64
[2023-09-19 15:31:40,519][ ERROR][PID:353][/usr/bin/copr-backend-process-build.managed.pid-353][background_worker_build.py:handle_task:804] No job object from Frontend
```

so now as expected, frontend returns 404 and backend can handle it thus logs look meaningful now. But yep, the race condition with eol, etc. events still remain (I would rather do a separate issue/PR about it)

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -191,7 +190,7 @@ def get_build_record(task, for_backend=False):
 
     except Exception as err:
         app.logger.exception(err)
-        return None
+        raise err
```
praiskup, line 193: This is weird.  app.logger.exception dumps the traceback into logs for the first time, and if we re-raise - some other exception handler will log it once more?

praiskup (approved):
Seems reasonable to me. Thank you!

Issue #2848, opened 2023-08-09
epoch taken by python-specfile is null, in srpm header is 0


praiskup:
As pointed out in https://github.com/fedora-copr/copr/issues/2839

```
Running SRPMResults tool
Package info: {
    "name": "python-copr",
    "epoch": null,
    "version": "1.129",
    "release": "1.git.68.9e71174",
    "exclusivearch": [],
    "excludearch": []
}
SRPMResults finished
```

```Running RPMResults tool
Package info:
{
    "packages": [
        {
            "name": "python-copr-doc",
            "epoch": 0,
            "version": "1.129",
            "release": "1.git.68.9e71174.fc39",
            "arch": "noarch"
        },
...
```

Those RPM data are taken from RPM headers; but if taken from spec-file, Epoch is not `null` but `0`. Is this a problem actually? Shouldn't we migrate Epoch from `null` to `0`?

praiskup:
I mean - in case of `specfile` failure, we fallback to reading this data from srpm header, which likely would provide `0`.

praiskup:
Also note that specfile returns string, while rpm header returns number:

```
Running SRPMResults tool
Package info: {
    "name": "tar",
    "epoch": "2",
...
```
Running RPMResults tool
Package info:
{
    "packages": [
        {
            "name": "tar-debugsource",
            "epoch": 2,
```

praiskup:
It seems that we should stay as close to RPM behavior as possible, and work with integers or None.  WDYT?

@nforro JFYI, don't you want to return integer value for "epoch" in specfile automatically?

nforro:
> @nforro JFYI, don't you want to return integer value for "epoch" in specfile automatically?

I see your point, but `epoch` represents the value of the `Epoch` tag, which can contain macros for example, so it has to be a string. If it is `None`, it means there is no `Epoch` tag in the spec file. `expanded_epoch` could be an integer, but still, I think it should follow the same logic and return `None` if there is no `Epoch` tag.

On a related note, be aware that there could be multiple `Epoch` tags in a spec file, and you may not always get what you want. For example here: https://src.fedoraproject.org/rpms/samba/blob/9602a07a36b5fbe33eca0c55c883b1a163ac2f64/f/samba.spec#_207
[...]

PR #2849, opened 2023-08-09
rpmbuild: make sure detected epoch is int() or None

praiskup:
Resolves: #2848

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -634,11 +634,12 @@ def get_chroot_arch(chroot):
 
 def format_evr(epoch, version, release):
     """
-    Return evr in format (epoch:)version-release
+    Return evr in format (epoch:)version-release.  The argument 'epoch' should
+    be integer value or null (but we rather also consider "strings" values).
     """
-    if epoch is not None and epoch.isdigit():
-        return f"{epoch}:{version}-{release}"
-
+    if epoch is not None:
+        if isinstance(epoch, int) or epoch.isdigit():
+            return f"{epoch}:{version}-{release}"
     return f"{version}-{release}"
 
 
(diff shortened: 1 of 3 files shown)


backend/copr_backend/helpers.py
```diff
@@ -634,12 +634,20 @@ def get_chroot_arch(chroot):
...
+    Return evr in format (epoch:)version-release.  The argument 'epoch' should
+    be integer value or null (but we rather consider "strings" here, too).
     """
-    if epoch is not None and epoch.isdigit():
-        return f"{epoch}:{version}-{release}"
+    if isinstance(epoch, str):
+        if epoch == "":
+            epoch = None
+        elif epoch.isdigit():
+            epoch = int(epoch)
```
FrostyX, line 644: I think we don't need to convert the epoch to `int` here. The `return f"{epoch}:{version}-{release}"` value will be same in both cases. I think this function can remain as it was before? Maybe only with this small change to work for integers as well?

```diff
-    if epoch is not None and epoch.isdigit():
+    if epoch is not None and (isinstance(epoch, int) or epoch.isdigit()):
         return f"{epoch}:{version}-{release}"
```

rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -406,7 +406,11 @@ def __init__(self, path):
         self.tags = self.spec.tags(self.spec.parsed_sections.package).content
 
     def __getattr__(self, name):
-        return getattr(self.spec, name)
+        value = getattr(self.spec, name)
+        if name == "epoch":
+            if value is not None:
```
FrostyX, line 411: Nit: I'd go with

```python
if name == "epoch" and value is not None:
```

FrostyX:
Thank you for the PR and sorry for introducing the 0 vs null inconsistency.

backend/copr_backend/helpers.py
praiskup: I had this variant, but it seemed harder to understand to me. But ... :-) I think I like it more now.

praiskup:
> Thank you for the PR and sorry for introducing the 0 vs null inconsistency.

No problem! I'm sure I at least reviewed this before without complaints ;-) because it wasn't the first time I was thinking about the implicit Epoch value in RPM.

FrostyX (approved):

PR #2852, opened 2023-08-09
rpmbuild: retry downloads from lookaside-cache

praiskup:
It is a pity to fail whole build when some random temporary networking issue happens.

Let's use "--connect-timeout", "60" to avoid indefinite hangs on connecting.

Use "--retry", "3", "--retry-all-errors", "--retry-delay", "10" to repeat any download failure.

Observed by a random failure against the texlive package: INFO: Calling: curl -H Pragma: -o logical-markup-utils.doc.tar.xz --location --remote-time --show-error --fail https://src.fedoraproject.org/repo/pkgs/rpms/texlive/logical-markup-utils.doc.tar.xz/sha512/80c44068a6e474071e40a0bf3dc67f0201b21eb84a041fad8423c59c9e34ce138833c54ead638979357dc3026eea55da034daa05e71a56ecabe701ebf53faf24/logical-markup-utils.doc.tar.xz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (56) Recv failure: Connection reset by peer

### DIFF
--- a/rpmbuild/copr_distgit_client.py
+++ b/rpmbuild/copr_distgit_client.py
@@ -99,18 +99,29 @@ def _load_config(directory):
 
 def download(url, filename):
     """ Download URL as FILENAME using curl command """
+
+    if not hasattr(download, "curl_has_retry_all_errors"):
+        # Drop this once EL8 is not a thing to support
+        output = check_output(["curl", "--help", "all"])
+        # method's static variable to avoid using 'global'
+        download.curl_has_retry_all_errors = b"--retry-all-errors" in output
+
     command = [
         "curl",
         "-H", "Pragma:",
         "-o", filename,
         "--location",
+        "--connect-timeout", "60",
+        "--retry", "3", "--retry-delay", "10",
         "--remote-time",
         "--show-error",
         "--fail",
-        url,
     ]
 
-    if call(command):
+    if download.curl_has_retry_all_errors:
+        command += ["--retry-all-errors"]
+
+    if call(command + [url]):
         raise RuntimeError("Can't download file {0}".format(filename))
 
 


nikromen (approved):

rpmbuild/copr_distgit_client.py
```diff
@@ -99,18 +99,28 @@ def _load_config(directory):
 
 def download(url, filename):
     """ Download URL as FILENAME using curl command """
+
+    if not hasattr(download, "curl_has_retry_all_errors"):
```
FrostyX, line 103: I am very much confused by this. IMHO `download` should be reference to the function itself. So using it this way doesn't make sense to me (same for `download.curl_has_retry_all_errors = ...` below). Am I missing something?

FrostyX:
But otherwise the PR looks good to me.

rpmbuild/copr_distgit_client.py
praiskup: This is the way around not using `global` keyword.  A function is object and can have attributes. I'd swear we already have something like this somewhere in our code. Nevermind, if you want me to change this, lemme know.

praiskup: https://stackoverflow.com/questions/279561/what-is-the-python-equivalent-of-static-variables-inside-a-function

FrostyX: Oh I see. Thank you for the link.

> Nevermind, if you want me to change this, lemme know.

No, I think it's okay. Can you please just add a short comment saying, that we are checking and setting a static variable for the function?

FrostyX (approved):

PR #2857, opened 2023-08-10
rpmbuild: save review.json from fedora-review

FrostyX:
See https://pagure.io/FedoraReview/pull-request/463 
See https://pagure.io/FedoraReview/issue/310

I am planning to use this file from the fedora-review-service and provide users with better feedback based on that.

Also, we don't need to depend on a specific fedora-review version. If the file is not present, we skip it without failing.


praiskup (approved):

praiskup:
CI should be fixed, I retriggered the build.

PR #2859, opened 2023-08-10
Some release-blocker fixes for copr-rpmbuild

FrostyX:
Sorry for multiple unrelated commits in one PR. They are too close to each other and would cause merge conflicts.

### DIFF
--- a/rpmbuild/copr_rpmbuild/helpers.py
+++ b/rpmbuild/copr_rpmbuild/helpers.py
@@ -407,9 +407,17 @@ class Spec:
 
     def __getattr__(self, name):
         value = getattr(self.spec, name)
-        if name == "epoch" and value is not None:
-            return int(value)
         return value
+
+    @property
+    def epoch(self):
+        """
+        Evaluated %{epoch} as integer or None
+        """
+        value = self.safe_attr("epoch")
+        if value in [None, ""]:
+            return None
+        return int(value)
 
     @property
     def exclusivearch(self):
--- a/rpmbuild/copr_rpmbuild/helpers.py
+++ b/rpmbuild/copr_rpmbuild/helpers.py
@@ -395,7 +395,7 @@ class Spec:
             # TODO We want to loop over all name-version chroots and parse the
             # spec values for each of them. For that, we will set dist macros.
             # It expect tuples like this: [("fedora", "39"), ("foo", "bar")]
-            macros = []
+            macros = [("dist", None)]
             self.spec = Specfile(path, macros=macros)
         except TypeError as ex:
             raise RuntimeError("No .spec file") from ex
--- a/rpmbuild/copr_rpmbuild/helpers.py
+++ b/rpmbuild/copr_rpmbuild/helpers.py
@@ -406,8 +406,7 @@ class Spec:
         self.tags = self.spec.tags(self.spec.parsed_sections.package).content
 
     def __getattr__(self, name):
-        value = getattr(self.spec, name)
-        return value
+        return self.safe_attr(name)
 
     @property
     def epoch(self):
--- a/rpmbuild/mock.cfg.j2
+++ b/rpmbuild/mock.cfg.j2
@@ -52,7 +52,7 @@ metadata_expire=0
 cost=1
 best=1
 
-{%- if "priority" in repo and priority is not none %}
+{%- if repo.get("priority") is not none %}
 priority={{ repo["priority"] }}
 {%- endif %}
 


rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -407,10 +407,18 @@ def __init__(self, path):
...
-        if name == "epoch" and value is not None:
-            return int(value)
         return value
 
+    @property
+    def epoch(self):
+        """
+        Evaluated %{epoch} as integer or None
+        """
+        value = self.safe_attr("epoch")
```
praiskup, line 416: This doesn't solve the "multiple" tag value behind %if/else condition, or does it?  Shouldn't we use the `spec.tags(spec.parsed_sections.package).content.epoch.expanded_value` variant?

FrostyX: IMHO this should be exactly it :-)


`self.safe_attr` returns `getattr(self.tags, name).expanded_value` while `self.tags` are `self.spec.tags(self.spec.parsed_sections.package).content`

So basically:

```python
getattr(self.spec.tags(self.spec.parsed_sections.package).content, name).expanded_value
```

praiskup: Nice then, I think. Thank you for the explanation. @nforro would you mind having a quick look?

praiskup:
Hmm, what about Release and Version? From #2848:

    "epoch": 11,
    "version": "%{tl_version}",

rpmbuild/mock.cfg.j2
```diff
@@ -52,7 +52,7 @@ metadata_expire=0
 cost=1
 best=1
 
-{%- if "priority" in repo and priority is not none %}
+{%- if repo.get("priority") is not none %}
```
praiskup, line 55: How it comes that the previous `iss not none` check doesn't work? :-/

rpmbuild/copr_rpmbuild/helpers.py
nforro: LGTM :+1:

nforro:
> Hmm, what about Release and Version? From #2848:
> 
> ```
> "epoch": 11,
> "version": "%{tl_version}",
> ```

That should be fixed by https://github.com/fedora-copr/copr/pull/2859/commits/104a7dde57e3099cef5f6ba382753a7235fe8571, right?

praiskup:
Indeed!

rpmbuild/mock.cfg.j2
FrostyX: We checked `priority is not none` instead of `repo["priority"] is not none`. But it must have been wrong, there is no `priority`.

praiskup:
So I'm afraid it brings us back to the original reason why we started using specfile :-)

```
Output: ['hello-2.8-1.src.rpm']
Running SRPMResults tool
Package info: {
    "name": "hello",
    "epoch": null,
    "version": "2.8",
    "release": "1.fc38",
    "exclusivearch": [],
    "excludearch": []
}
SRPMResults finished
```

Can you please override dist tag with an empty string?

rpmbuild/mock.cfg.j2
nikromen: Agh... I tested the output of mock config only via frontend's API and [copr-cli](https://github.com/fedora-copr/copr/blob/main/cli/copr_cli/build_config.py#L50) to retrieve `priority` so I totally overlooked this mistake... thanks for noticing it!

BTW I did the same in `mock-custom-build.cfg.j2`. Can you please fix it there too?

FrostyX:
> So I'm afraid it brings us back to the original reason why we started using specfile :-)

We have come full circle on this one, didn't we :smile: 

Anyway, the main branch doesn't work properly either.

Main branch:

```
Package info: {
    "name": "samba",
    "epoch": 2,
    "version": "%{samba_version}",
    "release": "%{samba_release}",
    "exclusivearch": [],
    "excludearch": []
}
```

This PR:

```
Package info: {
    "name": "samba",
    "epoch": 2,
    "version": "4.19.0",
    "release": "0.1.rc2.fc37",
    "exclusivearch": [],
    "excludearch": []
}
```

I am going to push a new commit which should IMHO work the best:

```
Package info: {
    "name": "samba",
    "epoch": 2,
    "version": "4.19.0",
    "release": "0.1.rc2.fc37",
    "exclusivearch": [],
    "excludearch": []
}
```

praiskup (approved):

PR #2861, opened 2023-08-10
frontend: skip ExcludeArch and ExclusiveArch for resubmitted builds

### LINKED ISSUE #2833: ExclusiveArch / Exclude Arch for resubmitted builds?
https://github.com/fedora-copr/copr/blob/bddee02fb014898075b69e058f1d8fe8a0d13dbf/frontend/coprs_frontend/coprs/views/coprs_ns/coprs_builds.py#L472-L483

We should make sure that we skip those also for resubmitted builds.

FrostyX:
Fix #2833

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -458,6 +458,18 @@ class BuildsLogic(object):
         build.pkg_version = source_build.pkg_version
         build.resubmitted_from_id = source_build.id
 
+        # Skip previously skipped chroots (ExcludeArch, ExclusiveArch, etc)
+        # We need this only for "SRPM upload" builds because they completely
+        # skip the SRPM build phase. Other build methods build SRPM pacakge
+        # again and send the parsed results to backend and frontend. Therefore
+        # from this point of view, there is no difference between resubmitted
+        # and not-resubmitted builds.
+        skipped = [ch.name for ch in source_build.build_chroots
+                   if ch.status == StatusEnum("skipped")]
+        for chroot in build.build_chroots:
+            if chroot.name in skipped:
+                chroot.status = StatusEnum("skipped")
+
         return build
 
     @classmethod


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -458,6 +458,13 @@ def create_new_from_other_build(cls, user, copr, source_build,
         build.pkg_version = source_build.pkg_version
         build.resubmitted_from_id = source_build.id
 
+        # Skip previously skipped chroots (ExcludeArch, ExclusiveArch, etc)
+        skipped = [ch.name for ch in source_build.build_chroots
+                   if ch.status == StatusEnum("skipped")]
+        for chroot in build.build_chroots:
+            if chroot.name in skipped:
+                chroot.status = StatusEnum("skipped")
```
praiskup, line 466: Couldn't we place this into backend_enqueue_buildchroots() and have the logic concentrated on one place?

FrostyX: The `backend_enqueue_buildchroots` method is not called when resubmitting. Should I call it from this place in the code?

FrostyX: I forgot how the resubmit works. It doesn't use the previous SRPM but rather builds it again. That means the whole python3-specfile parsing happens again and it is send to backend as `results.json` and then sent to frontend as for any other not-resubmitted build. Soooo, #2833 is invalid? When I run frontend from the `main` branch and resubmit package with exclusivearch, it skips the chroots.

praiskup: Yes, but  *resubnitted srpm ploads*. This is tricky...

praiskup (approved):
We discussed this 1:1 and we don't have a better (or easier) solution than setting the "skipped" state in two places.  So this is good enough, and better to have than not to have. Thank you!

Issue #2854, opened 2023-08-10
Can't set fully-qualified bootstrap image per-chroot


praiskup:
$ dev-copr edit-chroot @copr/TEST1691655449854450884-bootstrap-project/epel-8-x86_64 --bootstrap-image=quay.io/almalinuxorg/almalinux:8
No login/token configured in /home/praiskup/.config/dev-copr,
GSSAPI/Kerberos will be used when needed.

Something went wrong:
Error: bootstrap_image: Enter valid bootstrap image id (<name>[:<tag>], e.g. fedora:33).

praiskup:
Moving in progress; this will help us to fix the beaker tests.

PR #2865, opened 2023-08-11
frontend: allow specifying fully-qualified bootstrap images

praiskup:
Fixes: #2854

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -26,6 +26,7 @@ from coprs.logic.complex_logic import ComplexLogic
 from wtforms import ValidationError
 
 FALSE_VALUES = {False, "false", ""}
+REGEX_BOOTSTRAP_IMAGE = r"^[-\./\w]+(:\w+)?$"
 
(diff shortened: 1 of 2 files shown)


frontend/coprs_frontend/coprs/forms.py
```diff
@@ -26,6 +26,7 @@
 from wtforms import ValidationError
 
 FALSE_VALUES = {False, "false", ""}
+REGEX_BOOTSTRAP_IMAGE = r"^[-\./\w]+(:\w+)?$"
```
nikromen, line 29:
```suggestion
REGEX_BOOTSTRAP_IMAGE = r"^[-\.\/\w]+(:\w+)?$"
```
Should this backslash be escaped just in case or it doen't matter?

praiskup: that was meant to be an escape character for the following forward slash :-)

FrostyX (approved):
Thank you very much for the tests.
They are priceless whenever regexes are involved. That way I don't have to try parsing them in my head :smile:

PR #2866, opened 2023-08-11
backend: don't eat the "build detail collecting" traceback

praiskup:
Use "log.exception()" which passes the traceback down to backend.log. This helps me to debug some release-blocker issue right now, normally we (when infra is OK) we shouldn't ever catch an exception there.

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -1034,11 +1034,12 @@ class BuildsLogic(object):
                 for chroot in build.build_chroots:
                     arch = chroot.mock_chroot.arch
 
-                    exclusivearch = upd_dict["results"]["exclusivearch"]
+                    exclusivearch = upd_dict["results"].get("exclusivearch")
                     if exclusivearch and arch not in exclusivearch:
                         finish(chroot, StatusEnum("skipped"))
 
-                    if arch in upd_dict["results"]["excludearch"]:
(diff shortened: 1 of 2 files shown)


nikromen (approved):

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1034,11 +1034,12 @@ def finish(chroot, status):
                 for chroot in build.build_chroots:
                     arch = chroot.mock_chroot.arch
 
-                    exclusivearch = upd_dict["results"]["exclusivearch"]
+                    exclusivearch = upd_dict["results"].get("exclusivearch")
                     if exclusivearch and arch not in exclusivearch:
                         finish(chroot, StatusEnum("skipped"))
 
-                    if arch in upd_dict["results"]["excludearch"]:
+                    excludearch = upd_dict["results"].get("excludearch")
```
FrostyX, line 1041: The change is okay and will fix the traceback, so +1.
But IMHO the `"exclusivearch"` and `"excludearch"` should always be present, so there is probably a bug earlier in the pipeline. Per @praiskup on Slack, it seems that we are using `RPMResults` instead of `SRPMResults` for custom method.

FrostyX (approved):

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: Should I drop the commit?  Or what do you prefer?

PR #2867, opened 2023-08-11
rpmbuild: more reliable SRPM vs RPM build check

### LINKED ISSUE #2841: Unstoppable "running" job
I even tried "cancel", and the source state still stays in "running" state:
https://copr.stg.fedoraproject.org/coprs/frostyx/custom-1-TEST1687524244666456862/build/2914596/
https://copr.stg.fedoraproject.org/status/running/

FrostyX:
Fix #2841

The previous check worked for all SRPM methods except for custom method. It uses `self.chroot` to specify "what chroot to run the script in". Checking `source_type` should be more reliable.


praiskup (approved):
thank you

PR #2874, opened 2023-08-15
backend: dump the /update/ payload to worker.log

praiskup:
This should simplify debugging of the Backend <-> Frontend protocol.


praiskup:
Merging into the release that is going to be tagged right now.

Issue #2870, opened 2023-08-14
Excludearch mapping doesn't seem to respect `powerpc64le` arch


praiskup:
https://copr.stg.fedoraproject.org/coprs/g/copr/TEST1692016260110954285ExclusiveArch/build/2915090/

```
Start: rpmbuild procyon-0.6.0-0.3.20220221.git88a95fa.fc38.src.rpm
error: Architecture is not included: powerpc64le
Building target platforms: ppc64le
Building for target ppc64le
```

It appears like that build chroot should be skipped, too?

FrostyX:
The `results.json` from SRPM build looks like this:

```json
{
    "name": "procyon",
    "epoch": null,
    "version": "0.6.0",
    "release": "0.3.20220221.git88a95fa",
    "exclusivearch": [
        "aarch64",
        "ppc64le",
        "s390x",
        "x86_64",
        "noarch"
    ],
    "excludearch": []
}
```

So I would expect the ppc64le chroot not be skipped. Unless we queried the exclusivearch incorrectly.

FrostyX:
Hmm, there is a discrepancy.

```python
import specfile
spec = specfile.Specfile("procyon.spec")
print(spec.tags(spec.parsed_sections.package).content.exclusivearch)
aarch64 ppc64le s390x x86_64 noarch
```

vs

```bash
rpm -qp procyon-0.6.0-0.3.20220221.git88a95fa.src.rpm --queryformat '%{EXCLUSIVEARCH}\n' --nosignature
aarch64
```

praiskup:
In procyon package, there's [ExclusiveArch:  %{java_arches} noarch](https://src.fedoraproject.org/rpms/procyon/blob/rawhide/f/procyon.spec#_23).  This expands on my host like:

```
$ rpm --eval "%java_arches noarch"
aarch64 ppc64le s390x x86_64 noarch
```

So it really appears to me that we really **should build it**. But then, when buliding, rpmbuild fails with `error: Architecture is not included: powerpc64le`. There must be some bug, somewhere and I'm really curious where. Is the `%java_arches` macro broken? Is `rpmbuild` broken?

praiskup:
Hm, Koji seems to handle procyon builds just fine on ppc64le: https://koji.fedoraproject.org/koji/buildinfo?buildID=2226538

praiskup:
This must be a different bug then.  Let's resolve this post-release.

praiskup:
[...]

PR #2877, opened 2023-08-15
beaker: align tests with the August 2023 release

praiskup:
Mock 5.0 is enforced by the new Copr release, and there's the bootstrap_image feature that can not use "fedora image" for EPEL target chroot

The unspecified `Epoch` tag in spec file is newly evaluated as "null" value in the BuildChrootResults table, and returned as None to end-user (null JSON value).

The default PyPI method is pyp2spec, not pyp2rpm now.  Let's use a package that is easily buildable with pyp2spec (motionpaint doesn't comply with the new licensing format, we could submit PR but that would require a release into PyPI, and ...).

Skip one test related to a ExclusiveArch parsing for now; our ppc64le builders are somewhat broken or non-standard, see #2870.

Several cleanup actions are now wrapped with cleanupAction, so we can keep the intermediate files/projects hanging around with COPR_CLEANUP=false when debugging testsuite failures.

### DIFF
--- a/beaker-tests/Sanity/copr-cli-basic-operations/runtest-exclusivearch.sh
+++ b/beaker-tests/Sanity/copr-cli-basic-operations/runtest-exclusivearch.sh
@@ -8,6 +8,7 @@ HERE=$(dirname "$(realpath "$0")")
 source "$HERE/config"
 source "$HERE/helpers"
 
+exclusive_arch_package=https://pagure.io/copr/copr-test-sources/raw/master/f/exclusivearch-test-1-1.src.rpm
 
 rlJournalStart
     rlPhaseStartSetup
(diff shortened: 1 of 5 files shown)


beaker-tests/Sanity/copr-cli-basic-operations/runtest-exclusivearch.sh
```diff
@@ -33,7 +33,8 @@ rlJournalStart
 
         # This is a more complicated package with `BuildArch: noarch` and
         # ExclusiveArch for subpackages. Test that we don't fail while parsing it
-        rlRun "copr-cli build-distgit ${NAME_PREFIX}ExclusiveArch --name procyon"
+        # TODO: we fail the build actually, https://github.com/fedora-copr/copr/issues/2870
+        #rlRun "copr-cli build-distgit ${NAME_PREFIX}ExclusiveArch --name procyon"
```
FrostyX, line 37: The whole PR looks good to me but since it is not a release blocker, we should probably fix the TODO before merging.

FrostyX: Maybe by changing the package that we test or uncommenting this line and commenting out the 

```bash
chroots+=" --chroot fedora-$FEDORA_VERSION-ppc64le"
```

That way we can test that at least some chroots were skipped.

praiskup: The package is noarch, and still using `ExclusiveArch:  %{java_arches} noarch`.  In this case I think that the intention is not to ship on i386, but not sure.  Java folks though, I bet, don't want to [skip builds on ppc64le](https://koji.fedoraproject.org/koji/buildinfo?buildID=2226538) :-/

praiskup: > That way we can test that at least some chroots were skipped.

I mean, this package is not supposed to be skipped at build time on all Fedora Copr-supported architectures.

pkking:
Does this means we(openEuler) should upgrade builder's mock as well?

beaker-tests/Sanity/copr-cli-basic-operations/runtest-exclusivearch.sh
praiskup: What about now?

praiskup:
> Does this means we(openEuler) should upgrade builder's mock as well?

Hello @pkking, I'm not sure I get your question?  This is a test-only fix.  If you mean the released `copr-rpmbuild`, that one depends on Mock 5.0 and you should update it, yes.

FrostyX (approved):

Issue #2884, opened 2023-08-22
Packit build failures after merge


praiskup:
We have "negative" CI karma for quite some time:
 https://github.com/fedora-copr/copr/commits/main

What if we disabled Packit for the time being for pushes?

cc @lachmanfrantisek @lbarcziova

nikromen:
I can see failed builds since https://github.com/fedora-copr/copr/commit/fc73a835add608c13133b9024bbaf7a64715043d

perhaps it's due to the `~~dev` change we did in copr-common version

nikromen:
So it seems like parsed spec file in https://download.copr.fedorainfracloud.org/results/packit/copr/fedora-39-x86_64/06266070-python-copr-common/ and its SRPM is different (look to the spec file in SRPM and spec file inside .tar.gz in SRPM - their version differs) - that's responsible for the majority of errors

another bug I found (some time ago) and is responsible for the first fail in https://copr.fedorainfracloud.org/coprs/packit/copr/package/python-copr-common/ is that versions in spec file in SRPM and in tar.gz (inside SRPM) sometimes differ - usually when I submit a PR with new pkg version then it has still its lower version and one more push to PR/new PR to catch up/refresh the version is needed

but I have no idea whether this is our or packit/specfile issue

nikromen:
also, I don't think it's a ci-blocker since we have our CI still running in parallel

praiskup:
> edit: I guess grep '"%version"' setup.py fails when build is submitted by packit

For some reason, Packit generates version `0.20.1`, while our specfile has `0.20`.

> also, I don't think it's a ci-blocker since we have our CI still running in parallel

Well I need to have green CI mark in recent commits, otherwise it blocks my brain. :-)

praiskup:
#2956 likely fixes this!

praiskup:
And all green, finally!

praiskup:
Thanks @lachmanfrantisek :)

PR #2886, opened 2023-08-23
packit: use tito to generate "python-copr-common" versions

praiskup:
Similar thing was done in mock:
https://github.com/rpm-software-management/mock/blob/7222e29f57f83fc7257f8c0015e112e951d9299f/.packit.yaml#L11-L16

Relates: #2884

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -3,6 +3,10 @@
 # https://packit.dev/docs/configuration/
 upstream_project_url: https://github.com/fedora-copr/copr.git
 
+srpm_build_deps:
+  - tito
+  - git
+
 packages:
 
   python-copr:
@@ -27,6 +31,13 @@ packages:
     specfile_path: python-copr-common.spec
     files_to_sync:
       - python-copr-common.spec
+    actions:
+      create-archive:
+        - bash -c "tito build --tgz --test -o ."
+        - bash -c "ls -1t ./*.tar.gz | head -n 1"
+      get-current-version:
+        - bash -c "git describe --match python-copr-common-[0-9]* --abbrev=0 HEAD | egrep -o
+          [0-9]+\.[0-9]+"
     upstream_tag_template: python-copr-common-{version}
     upstream_tag_include: "^python-copr-common-.*"
 


praiskup:
Are "actions" shared across packages?  I only wanted to use tito for copr-common ATM: https://download.copr.fedorainfracloud.org/results/packit/fedora-copr-copr-2886/srpm-builds/06336913/builder-live.log.gz

.packit.yaml
```diff
@@ -27,6 +31,13 @@ packages:
     specfile_path: python-copr-common.spec
     files_to_sync:
       - python-copr-common.spec
+    actions:
+      create-archive:
+        - bash -c "tito build --tgz --test -o ."
+        - bash -c "ls -1t ./*.tar.gz | head -n 1"
```
praiskup, line 37: This seems to leak to "global" config, because also `python-copr` is affected, not just `python-copr-common`. @lbarcziova fyi

.packit.yaml
```diff
@@ -27,6 +31,13 @@ packages:
...
+      get-current-version:
+        - bash -c "git describe --match python-copr-common-[0-9]* --abbrev=0 HEAD | egrep -o
+          [0-9]+\.[0-9]+"
```
praiskup, line 40: We could happily use this at global level if we could specify `--match` a generic way.  Can't we use something like `{{package_name}}-[0-9]*`?

praiskup:
Wait-for-copr (naturally) cycles here, because wait-for-copr wasn't successful.  I'm curious whether we could change this from `wait-for-copr` to `wait-for-packit-job` instead.  Or, if we couldn't use the `--after-build-id` at least.
https://download.copr.fedorainfracloud.org/results/packit/fedora-copr-copr-2886/srpm-builds/06336910/builder-live.log

.packit.yaml
lbarcziova: currently it is not possible :/ (looks like related to https://github.com/packit/packit/issues/1925) however, do you need to introduce an action for this? The default way Packit gets the version when no action is specified is by extracting it from the tag, and these can be filtered using `upstream_tag_include` (that you have already defined)

.packit.yaml
lbarcziova: oh, can you provide some more info on where I can see this, please?

I had a look into our logs and from there, it looks ok:

```
Package config before loading:
{
    "upstream_project_url": "https://github.com/fedora-copr/copr.git",
    "srpm_build_deps": [
        "wait-for-copr"
    ],
    "packages": {
        "python-copr": {
            "downstream_package_name": "python-copr",
            "upstream_package_name": "copr",
            "paths": [
                "./python"
            ],
            "specfile_path": "python-copr.spec",
            "files_to_sync": [
                "python-copr.spec"
            ],
            "upstream_tag_template": "python-copr-{version}",
            "upstream_tag_include": "^python-copr-.*",
            "actions": {
                "post-upstream-clone": [
                    "sh -c \"wait-for-copr --owner $COPR_OWNER --project $COPR_PROJECT python-copr-common `git rev-parse --short HEAD`\""
                ]
            }
        },
        "python-copr-common": {
            "downstream_package_name": "python-copr-common",
            "upstream_package_name": "copr-common",
            "paths": [
                "./common"
            ],
            "specfile_path": "python-copr-common.spec",
            "files_to_sync": [
                "python-copr-common.spec"
            ],
            "actions": {
                "create-archive": [
                    "bash -c \"tito build --tgz --test -o .\"",
                    "bash -c \"ls -1t ./*.tar.gz | head -n 1\""
                ],
                "get-current-version": [
                    "bash -c \"git describe --match python-copr-common-[0-9]* --abbrev=0 HEAD | egrep -o [0-9]+\\.[0-9]+\""
                ]
            },
            "upstream_tag_template": "python-copr-common-{version}",
            "upstream_tag_include": "^python-copr-common-.*"
        },
        ...
```

lachmanfrantisek:
@praiskup Actually, we are working on a more generic solution for job dependencies but it's not there yet:
https://github.com/packit/packit-service/issues/2105

.packit.yaml
praiskup: Sorry, I probably wrongly interpreted the logs! :(

praiskup:
This is a weird failure, two success/two failed chroots.  Seems like `wait-for-copr` checks only one chroot?
https://copr.fedorainfracloud.org/coprs/packit/fedora-copr-copr-2886/build/6336910

lbarcziova:
@praiskup I think in this particular case the reason was timeout (30 minutes) for `wait-for-copr`... @lachmanfrantisek should we adjust the value for it and/or make the script noticeably fail when the timeout was reached?

praiskup:
FYI, some statuses are not yet updated after 17 hours.

lbarcziova:
@praiskup sorry for that, this is a bug https://github.com/packit/packit-service/issues/2122 that was just fixed and should be in production next Tuesday

praiskup:
Is there a way to make Packit jobs "optional" so the overall "CI" status isn't affected?

.packit.yaml
FrostyX, line 40: Unrelated to your comments, there is another way to get the current version. Not sure if you find it helpful or not.

```
[jkadlcik@zeratul cli]$ cat ../.tito/packages/python-copr-common |egrep -o [0-9]+\.[0-9]+
0.20
```

Also, `egrep` is deprecated, so we should use `grep -E` instead.

lachmanfrantisek:
> Is there a way to make Packit jobs "optional" so the overall "CI" status isn't affected?

I am not sure if this is even possible on GitHub. You/we can try a few things:
* On GitHub, you can configure a set of required statuses.
* GitHub statuses allow us to use a neutral state but this lacks the information if it failed/succeded.
* Packit can run some jobs just on manual request => but this will not run the job so that's not what we need.

praiskup:
Closing in favor of #2956

PR #2892, opened 2023-08-28
frontend: remove redundant dot in form

nikromen:
There is dot in this form but there aren't dots in other forms... and also because of the dot the colon renders like this:
Screenshot_20230828_154226

which hurt my eyes :D


praiskup:
Nah I noted this as well, you were quicker with the PR, thank you!

Can you please change `<form>` so we move the long text from the left label to the small label-like gray text?  I'd like to have something like

```
repo priority: [    ^v]
               (i) Use the priority=<INT> config option .... Defaults to 99.
```

frontend/coprs_frontend/coprs/templates/coprs/_coprs_forms.html
```diff
@@ -159,7 +159,7 @@ <h3 class="panel-title">{{ counter('instructions') }}. Other options</h3>
     {{ render_field(form.repo_priority,
         class="short-input-field",
         placeholder='Optional',
-        info='Set the priority value of this repository. Defaults to 99.') }}
+        info='Use the priority=&lt;INT&gt; config option for repositories in this project, see man dnf.conf(5) for more info. Defaults to 99.') }}
```
praiskup, line 162: Last small wish (I changed my opinion) - I'd rather remove "Defaults to 99." part (this info is both in dnf.conf(5) and our default is actually None; plus I don't think we should unnecessarily raise eyebrows).

praiskup:
Otherwise LGTM, thank you!

frontend/coprs_frontend/coprs/templates/coprs/_coprs_forms.html
nikromen: np

praiskup (approved):

Issue #2621, opened 2023-04-03
Allow passing make variables when using the 'make srpm' build method


praiskup:
Original issue: https://bugzilla.redhat.com/show_bug.cgi?id=1948687
Opened: 2021-04-12 18:29:23
Opened by: Ondrej Mosnacek

---

#### Ondrej Mosnacek commented at 2021-04-12 18:29:23:
I would like to have a way to specify name-value pairs of variables in each package's settings that would passed to the make command when creating the SRPM.

For example, I have a COPR where I would like to build packages from the same source git branch using different dist-gits (fetched dynamically in the make script) for different chroots. I.e. I would have a separate COPR package for each chroot, each specifying a different variable value (FEDORA=33, FEDORA=34, FEDORA=rawhide, ...) that would be interpreted by the 'make srpm' script to fetch the right dist-git branch. All of them would build from the same git repo and branch via a webhook.

---

#### Miroslav Suchý commented at 2021-04-12 20:09:13:
The golden standard is to have SRC.RPM the same for all chroots. It is better to do this logic inside of SPEC file using

%if 0%{fedora} && 0%{fedora} >33

praiskup:
We could export the CHROOT_NAME in variable e.g. together with other variables we already export.

thom311:
How about having a "Script" input field (similar to the one in the "Custom" source type). That script would be sourced(!!) by the shell right before `make -f ... srpm`.

The common usecase would be just to set some environment variables before `make`, but this would be fully flexible.

PR #2895, opened 2023-09-01
all: add "prepare_script" to SCM build method

thom311:
Add a new "prepare_script" option to the SCM method.

It (currently) only works with "make srpm" build type. If specified, we will source(!) the script by the calling shell (bash?) right before `make -f ... srpm`.

This allows to run arbitrary shell commands before `make srpm`. The main usage would be to set environment variables that are honored by the `make srpm` script.

---

This would be my idea for solving https://github.com/fedora-copr/copr/issues/2621

Main problem: it's quite the expert method. The user can provide a shell script that is sourced. Sourcing is already confusing, and may not be understood by the user. Also, copr itself doesn't understand what the script does (but neither does it understand what happens during `make srpm`). Also, the main usage is to set arbitrary environment variables, for which a more specific solution might have better ergonomics. On the other hand, it's very flexible and relatively easy to add.

This is not ready yet. I don't know how to test it, it's certainly broken. This is an RFC.

Also, the web UI will always show the "Prepare Script" field, even if it only works with "make srpm" type.

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -382,6 +382,7 @@ class Commands(object):
             "subdirectory": args.subdirectory,
             "spec": args.spec,
             "scm_type": args.scm_type,
+            "prepare_script": ''.join(args.prepare_script.readlines()) if args.prepare_script else '',
             "source_build_method": args.srpm_build_method,
         }
         return self.process_build(args, self.client.build_proxy.create_from_scm, data)
@@ -794,6 +795,7 @@ class Commands(object):
             "subdirectory": args.subdirectory,
             "spec": args.spec,
             "scm_type": args.scm_type,
+            "prepare_script": ''.join(args.prepare_script.readlines()) if args.prepare_script else '',
             "source_build_method": args.srpm_build_method,
             "max_builds": args.max_builds,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
@@ -1328,6 +1330,8 @@ def setup_parser():
                                         help="relative path from the subdirectory to the .spec file")
     parser_scm_args_parent.add_argument("--type", dest="scm_type", choices=["git", "svn"], default="git",
                                         help="Specify versioning tool. Default is 'git'.")
+    parser_scm_args_parent.add_argument('--prepare-script', type=argparse.FileType('r'),
+                                        help='text file (script) to be used as prepare script for make_srpm method')
     parser_scm_args_parent.add_argument("--method", dest="srpm_build_method", default="rpkg",
                                         choices=["rpkg", "tito", "tito_test", "make_srpm"],
                                         help="Srpm build method. Default is 'rpkg'.")
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -875,6 +875,24 @@ class BasePackageForm(BaseForm):
     )
 
 
+def cleanup_script(string):
+    if not string:
+        return string
+
+    if string.split('\n')[0].endswith('\r'):
+        # This script is most probably coming from the web-UI, where
+        # web-browsers mistakenly put '\r\n' as EOL;  and that would just
+        # mean that the script is not executable (any line can mean
+        # syntax error, but namely shebang would cause 100% fail)
+        string = string.replace('\r\n', '\n')
+
+    # And append newline to have a valid unix file.
+    if not string.endswith('\n'):
+        string += '\n'
+
+    return string
+
+
 class PackageFormScm(BasePackageForm):
     scm_type = wtforms.SelectField(
         "Type",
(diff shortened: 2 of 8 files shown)


thom311:
the use case is for example, NetworkManager.

At https://copr.fedorainfracloud.org/coprs/networkmanager/ there are two projects: "NetworkManager-main"
 and "NetworkManager-main-debug".

they differ in that one of them is build with debug (and assertions) enabled. As the build script is in git, we need to differenciate between the builds. Currently, the copr uses the "Custom" method, where the script looks like:
```
#!/bin/bash

export GIT_REF=main
export DEBUG=1
export LTO=
curl https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/raw/main/contrib/scripts/nm-copr-build.sh | bash
```

To use the `make srpm` method, we would need a way to set parametrize the build, depending on the copr project. The "prepare-script" option could be used for that.

praiskup:
I don't like creating another "custom" method from "make srpm" either.

> To use the make srpm method, we would need a way to set parametrize the build, depending on the copr project. The "prepare-script" option could be used for that.

We have those exports for the custom method, I personally wouldn't be egainst exporting thouse environment variables also for other SCM methods.  Then you can implement whatever logic in your .copr/Makefile
https://github.com/fedora-copr/copr/blob/74c63e9f3bc33e0301f1c023f1e16920dd18a26b/rpmbuild/copr_rpmbuild/providers/custom.py#L87-L92

thom311:
in the https://copr.fedorainfracloud.org/coprs/networkmanager/ projects (that use a custom builds), I can trigger a manual build and adjust the environment variables (i.e. the build configuration) for that manual invocation. That's useful to create a build for testing something.

Variables like "COPR_PROJECT" still per-project, and don't serve the same purpose.

The advantage of the "make srpm" approach compared to the custom-build is that it requires fewer configuration in copr. I can basically leave most options at the default (only require "Type:git", "Clone-URL", "Commitish" and "Build-Type:make-srpm"). The prepare-script would also usually be blank, but I could easily add there a 
```
    MY_BUILD_OPTION=foo
```

praiskup:
If you need a per-build knob, I believe this is a duplicate to #1334; namely the per-build `--with` and `--without` rpmbuild option.  We have this on per-chroot level already, see `copr-cli edit-chroot --help`.

praiskup:
But I agree that the only way to do this now is `copr buildcustom --script ...` with customized per-build script generated on the fly.

thom311:
> namely the per-build --with and --without rpmbuild option

Those options are passed when building the RPM, while `make srpm` and this pull request is about generating the SRPM.

Back to the NetworkManager example, based on the environment variables the script will generate different SRPMs, so that we don't need a `--with` option during the package build.

What makes `make srpm` convenient, is that it automatically checks out the right git repository. With the custom script, that is re-implemented. Also, the custom script is fetched from the internet first with `curl ... | bash`, to bootstrap the build. Again, with `make srpm` that is all not necessary.

Also, it seems to me that github notifications don't allow to filter by branch name. I didn't figure out how to ignore the event for a wrong branch with a custom script (but there might be a way). With `make srpm` that also gets handled automatically and correct.

FrostyX:
> This allows to run arbitrary shell commands before make srpm. The main usage would be to set environment variables that are honored by the make srpm script.

I am going to ask the dumb question here, but what is the benefit of running arbitrary shell commands in a separate script before `make srpm`, instead of running them at the beginning of your `make srpm`? Or making the `srpm` target of your Makefile to depend on some `prep` target?

thom311:
> but what is the benefit of running arbitrary shell commands in a separate script before make srpm

> The main usage would be to set environment variables that are honored by the make srpm script.

To parametrize `make srpm`. Of course, you can do anything in that build script, but you cannot provide parameters from the copr setup to do different things depending on copr the build configuration (in copr).

lachmanfrantisek:
Hi @thom311 !

I have an alternative solution for you...;)  Would you be interested in using [Packit](packit.dev/) for your use case instead?

Besides other features, Packit is trying to provide a more feature-rich and reliable git forge integration with Copr.
(With Packit, you can have reporting right on git forge, automatic retries and Copr project management. As a bonus, you can have tests and automation of the Fedora maintenance.)

Packit internally uses the custom method and can do what you are probably trying to do here. There are also [customizable "actions"](https://packit.dev/docs/configuration/actions/) if you are not happy with the default behaviour. And you don't need to create/maintain/fix-when-it-breaks your custom solution.

And no worries. Packit is definitely not a Copr's opponent..;) We're closely collaborating and thanks to Packit, Copr does not need to focus on the git forge integration itself and can spend the time on other important things.

Here is the onboarding guide: https://packit.dev/docs/guide

If you want any help, ask on [#packit:fedora.im (Matrix)](https://chat.fedoraproject.org/#/room/#packit:fedora.im) or we can also setup a meeting to discuss this if you want.

praiskup:
@thom311 would packit help, or do you need Copr-only solution?

thom311:
I plan to investigate packit (for other reasons too), I just didn't get to it.

In general, there is already a workaround (the custom-build). If packit provides another workaround then ... great.

I expect packit to be very useful, but I don't see that as an conflict/alternative to this feature. 

What this PR would provide, is an ability to customize the `make srpm` approach significantly more. The `make srpm` aproach is pretty nice and easier to use than the custom-build-script. In particular, it nicely solves what I said:

> Also, it seems to me that github notifications don't allow to filter by branch name. I didn't figure out how to ignore the event for a wrong branch with a custom script (but there might be a way). With make srpm that also gets handled automatically and correct.

Issue #2681, opened 2023-04-23
Licensing and Forbidden Items links in docs are archived


penn5:
At https://github.com/fedora-copr/copr/blob/main/doc/user_documentation.rst#faq the links to Licensing and Forbidden Items are no longer updated

FrostyX:
Hello @penn5,
thank you for the report.

I updated the allowed licenses URL in PR #2682 but I don't know how to update the forbidden items URL. Is there any replacement of this page? Even this page
https://docs.fedoraproject.org/en-US/package-maintainers/New_Package_Process_for_Existing_Contributors/
points to the old wiki.

praiskup:
Triage time: We don't have the forbidden items (yet?) or something similar. Can we drop the link or add a link to "bad licenses" page?

praiskup:
Closing as fixed by #2899, thank you for the report, @penn5

PR #2899, opened 2023-09-05
doc: s/Forbidden/Not-Allowed/ licenses

praiskup:
The list of "forbidden" items started over again from scratch as a new doc page.

Relates: #2681


FrostyX:
+1

I think we can close #2681 after that.

Issue #2904, opened 2023-09-08
Cancel builds when disabling a chroot as EOL


FrostyX:
We have some builds running for months, e.g.
https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5942294/
That's because we disabled F36 as EOL and the build couldn't finish `fedora-36-s390x` before that. 

We need to mark all `fedora-36-*` build chroots (or maybe all not yet running) as `canceled` as a part of the chroot EOL process.

FrostyX:
It produced a bunch of tracebacks every few seconds in `frontend.log`, so I ran:

```
coprdb=# UPDATE build_chroot SET status=2 WHERE mock_chroot_id in (149, 150, 151, 152, 153, 154) AND status=4;
UPDATE 3
```

praiskup:
Is this #2842?

FrostyX:
Please see PR #2905, I explained how it relates to #2842 there

praiskup:
Could we do this with all states that are `not finished()`?

@FrostyX canceling builds is not wise - we want to cancel just particular buildroot.

Could we maybe fail the chroot when getting the task, when it is already EOLed and we can not do anything better?  perhaps in #2842?

nikromen:
re-raising something from frontend won't help here since the job is stuck in `/backend/penging-jobs` so frontend will process it every N secs but it won't touch it in the end. 

> Could we maybe fail the chroot when getting the task

Failing already EOLed/disabled chroots sounds reasonable to me

PR #2905, opened 2023-09-08
frontend: don't traceback on generating tasks for disabled chroots

FrostyX:
This happens in cases like #2904 and infinitely produces the following tracebacks:

    ERROR:coprs:'NoneType' object has no attribute 'module_setup_commands'
      File "/usr/share/copr/coprs_frontend/coprs/views/backend_ns/backend_general.py", line 175, in get_build_record
        modules = copr_chroot.module_setup_commands
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: 'NoneType' object has no attribute 'module_setup_commands'

### DIFF
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -172,6 +172,11 @@ def get_build_record(task, for_backend=False):
         })
 
         copr_chroot = CoprChrootsLogic.get_by_name_safe(task.build.copr, task.mock_chroot.name)
+        if not copr_chroot:
+            app.logger.error("The %s chroot is disabled (build %s)",
+                             task.mock_chroot.name, task.build.id)
+            return None
+
         modules = copr_chroot.module_setup_commands
         if modules:
             build_record["modules"] = {'toggle': modules}


FrostyX:
I found out that I am accidentally trying to fix one of the exceptions from #2816 so sorry about that @nikromen.
But I think this is not mutually exclusive with PR #2842. This PR only avoids dumping tracebacks into the `frontend.log` but the route still returns 200 status code.

```
[2023-09-08 22:03:13,787] ERROR in backend_general: The fedora-36-aarch64 chroot is disabled (build 6381264)
```

```
[jkadlcik@zeratul frontend]$ curl -I http://127.0.0.1:5000/backend/get-build-task/6381264-fedora-36-aarch64
HTTP/1.0 200 OK
```

that is something that PR #2842 is trying to fix and I am leaving it to it.

FrostyX:
PR #2842 is merged now, so I am revisiting this PR.

Previously the code was

```python
copr_chroot = CoprChrootsLogic.get_by_name_safe(task.build.copr, task.mock_chroot.name)
```

which returned `None` for inactive chroots and that's why I introduced the conditional return in this PR. However, now we use 

```python
copr_chroot = ComplexLogic.get_copr_chroot_safe(task.build.copr, task.mock_chroot.name) 
```

which raises an exception and therefore this PR is not needed anymore.

Issue #2869, opened 2023-08-14
Repartition disks on Fedora Copr hypervisors


praiskup:
This is likely blocking power9 builds

praiskup:
Now root@vmhost-p08-copr01.rdu-cc.fedoraproject.org is down (not SSH responding).

praiskup:
Requested rights to power-cycle: https://pagure.io/fedora-infrastructure/issue/11476

praiskup:
After another reboot, it started working.

praiskup:
Again both p08 machines down :-( being resolved still in the same infra ticket but reopening to "admit" there still is some issue.

nikromen:
and they were down once again

praiskup:
We want to move `/` partition onto it's own raid, and stop using the `raid6` for swap and disposable epemeral images (raid0 or lvm striping should be used for that).

Tracking progress:

- [ ] vmhost-x86-copr01.rdu-cc.fedoraproject.org
- [ ] vmhost-x86-copr02.rdu-cc.fedoraproject.org
- [ ] vmhost-x86-copr03.rdu-cc.fedoraproject.org
- [ ] vmhost-x86-copr04.rdu-cc.fedoraproject.org
- [ ] vmhost-p08-copr01.rdu-cc.fedoraproject.org
- [ ] vmhost-p08-copr02.rdu-cc.fedoraproject.org
- [ ] vmhost-p09-copr01.rdu-cc.fedoraproject.org

praiskup:
https://nagios.fedoraproject.org/nagios/cgi-bin//status.cgi?hostgroup=copr_hypervisor&style=detail

praiskup:
The (rather long, eventually) post is here: https://pavel.raiskup.cz/blog/fedora-copr-hypervisor-disk-repartitioning.html

praiskup:
I spent at least an hour writing a good report now against LibVirt, and then
I realized that actually, the Power9 machine on Fedora 38 doesn't suffer from the
same qcow2 issues (it is on SSD, but still).  The other machines are slow, but also
very old.  While it doesn't make sense to have a disk slowdown with qcow2 after
`rhel6` to `rhel0` to me, I don't think I want to bother LibVirt folks (OK on newer
hardware, OK also on older after moving to RAW).  Giving up the last sub-task.

PR #2909, opened 2023-09-14
backend: provide per-arch & per-owner worker limit

praiskup:
The default "max_workers" limit might be too large (e.g. in Fedora Copr it is 45) compared to the number of available workers of some architecture type (e.g. for s390x there's now max 18 workers, and ppc64le is often limited too due to #2869).  This patch then brings another FUP mitigation.

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -6,7 +6,7 @@
 %global tests_version 4
 %global tests_tar test-data-copr-backend
 
-%global copr_common_version 0.16.4.dev
+%global copr_common_version 0.20.1.dev1
 
 Name:       copr-backend
 Version:    1.172
--- a/backend/copr_backend/daemons/build_dispatcher.py
+++ b/backend/copr_backend/daemons/build_dispatcher.py
@@ -2,7 +2,7 @@
 BuildDispatcher related classes.
 """
 
-from copr_common.worker_manager import GroupWorkerLimit
+from copr_common.worker_manager import HashWorkerLimit
 from copr_backend.dispatcher import BackendDispatcher
 from copr_backend.rpm_builds import (
     ArchitectureWorkerLimit,
@@ -99,7 +99,7 @@ class BuildDispatcher(BackendDispatcher):
         for limit_type in ['sandbox', 'owner']:
             max_builders = backend_opts.builds_limits[limit_type]
             self.log.info("setting %s limit to %s", limit_type, max_builders)
-            self.limits.append(GroupWorkerLimit(
+            self.limits.append(HashWorkerLimit(
                 lambda x, limit=limit_type: getattr(x, limit),
                 max_builders,
                 name=limit_type,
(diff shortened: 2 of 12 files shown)


backend/copr_backend/rpm_builds.py
```diff
@@ -133,6 +137,18 @@ def predicate(x):
         super().__init__(predicate, limit, name="arch_{}".format(architecture))
 
 
+class ArchitectureUserWorkerLimit(GroupWorkerLimit):
+    """
+    Limit number of machines of specific architecture we give to a single
+    Copr owner (user or group).
+    """
+    def __init__(self, architecture, limit):
+        def hasher(x):
```
FrostyX, line 146: This probably doesn't have to be nested if we make it `@staticmethod`

backend/copr_backend/rpm_builds.py
```diff
@@ -133,6 +137,18 @@ def predicate(x):
         super().__init__(predicate, limit, name="arch_{}".format(architecture))
 
 
+class ArchitectureUserWorkerLimit(GroupWorkerLimit):
```
FrostyX, line 140: There is probably nothing we can do but the inheritance doesn't look right (UserWorkerLimit from GroupWorkerLimit)

praiskup: Yeah, I think we could rename `GroupWorkerLimit` into `HashedWorkerLimit` :-)?

backend/copr_backend/rpm_builds.py
praiskup: that can be lambda, going to change...

backend/copr_backend/daemons/build_dispatcher.py
```diff
@@ -83,9 +84,14 @@ def __init__(self, backend_opts):
         super().__init__(backend_opts)
         self.max_workers = backend_opts.builds_max_workers
 
-        for tag_type in ["arch", "tag"]:
-            lclass = ArchitectureWorkerLimit if tag_type == "arch" else \
-                     BuildTagLimit
+        for tag_type in ["arch", "tag", "arch_per_owner"]:
+            match tag_type:
```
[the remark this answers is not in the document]
praiskup: Yeah, but when someone says "pattern" matching, I expect it to work with `fnmatch` patterns (my shell deformation..) :-) TIL too, I actually thought this is just a dumy `C switch-case` alternative.

backend/copr-backend.spec
```diff
@@ -6,7 +6,7 @@
 %global tests_version 4
 %global tests_tar test-data-copr-backend
 
-%global copr_common_version 0.16.4.dev
+%global copr_common_version 0.20.1.dev1
```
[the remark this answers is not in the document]
praiskup: We ar jumping from 0.20.1 => 0.20.1.dev1;  this is just a requirement that we haven't had to bump for quite some time.

PR #2920, opened 2023-09-18
Collect Fedora Review Logs After Runs

### LINKED ISSUE #2919: Collect Fedora Review Log
When troubleshooting `FedoraReview` bugs (like the fact that every build is currently failing), it would be really helpful to collect the log file.

This file is currently always outputted to `$XDG_CACHE_HOME/fedora-review.log`, which usually means `~/.cache/fedora-review.log`.

thebeanogamer:
This is really hacky and unpleasant. `fedora-review` does not have a way to set a custom log path, and adding one will not be easy. You have to wait for it to have parsed the command line arguments, and by that point it's already expecting to have started logging. See <https://pagure.io/FedoraReview/blob/master/f/src/FedoraReview/settings.py#_44> and its uses for more context.

This PR redirects everything that's looking at `$XDG_CACHE_HOME` into a unique directory per run, then discards everything except the `fedora-review` logs in that directory. I am very open to suggestions of better ways to do this. 

For all its many many faults though, this did work when I tested it locally.

Resolves #2919

### DIFF
--- a/rpmbuild/copr_rpmbuild/automation/fedora_review.py
+++ b/rpmbuild/copr_rpmbuild/automation/fedora_review.py
@@ -10,8 +10,27 @@ https://fedoraproject.org/wiki/Package_Review_Process
 
 import os
 import shutil
+from contextlib import contextmanager
+from typing import Generator
+
 from copr_rpmbuild.helpers import run_cmd
 from copr_rpmbuild.automation.base import AutomationTool
+
+
+@contextmanager
+def cache_directory(resultdir) -> Generator[str, None, None]:
+    """
+    Create a directory for a job to use as the XDG cache.
+
+    :param str resultdir: Parent directory for the run results
+    :return: Path to cache directory
+    """
+    cachedir = os.path.join(resultdir, "cache")
+    try:
+        os.makedirs(cachedir, exist_ok=True)
+        yield cachedir
+    finally:
+        shutil.rmtree(cachedir)
 
 
 class FedoraReview(AutomationTool):
@@ -44,17 +63,17 @@ class FedoraReview(AutomationTool):
             "--name", self.package_name,
             "--mock-config", self.mock_config_file,
         ]
+        with cache_directory(self.resultdir) as cachedir:
+            try:
+                result = run_cmd(cmd, cwd=self.resultdir, env={"XDG_CACHE_HOME": cachedir})
+                self.log.info(result.stdout)
+            except RuntimeError as ex:
+                self.log.warning("Fedora review failed\nerr:\n%s", ex)
+                self.log.warning("The build itself will not be marked "
+                                 "as failed because of this")
+            self._filter_results_directory(cachedir)
 
-        try:
-            result = run_cmd(cmd, cwd=self.resultdir)
-            self.log.info(result.stdout)
-        except RuntimeError as ex:
-            self.log.warning("Fedora review failed\nerr:\n%s", ex)
-            self.log.warning("The build itself will not be marked "
-                             "as failed because of this")
-        self._filter_results_directory()
-
-    def _filter_results_directory(self):
+    def _filter_results_directory(self, cachedir: str):
         """
         Currently, fedora-review tool doesn't have an option to specify
         a destdir, and produces output to a directory called after the package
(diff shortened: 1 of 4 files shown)


praiskup:
Thank you for the PR, @thebeanogamer

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -45,8 +46,10 @@ def run(self):
             "--mock-config", self.mock_config_file,
         ]
 
+        os.makedirs(self.cachedir, exist_ok=True)
```
praiskup, line 49: I fear directory leftovers. Could you please use this pattern?
```
os.makedirs(self.cachedir, exist_ok=True)
try:
    ... other logic ...
finally:
    shutil.rmtree(self.cachedir)
```

I really like context managers for this:
```python
@contextmanater
def with_cache_dir(...):
    create_the_directory
    try:
        yield
    finally:
        drop_the_dir
```

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -86,7 +86,7 @@ def _filter_results_directory(self):
         # The fedora-review command failed so early that it didn't even create
         # the resultdir. Nothing to do here.
         if not os.path.exists(srcdir):
-            self.log.error("Can't find fedora-review results: %s", srcdir)
+            self.log.error(f"Can't find fedora-review results: {srcdir}")
```
praiskup, line 89: The correct way to use logging is to use `%s` so the strings are only evaluated if really used.

thebeanogamer: Reverted this and updated my new log line to match

rpmbuild/copr_rpmbuild/automation/fedora_review.py
thebeanogamer: I've had a go at porting this over to a context manager. I'm not wild about having to change the signature of `_filter_results_directory`, but it felt cleaner than adding it to the `FedoraReview` class. Happy to do it that way if you'd prefer.

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -44,17 +56,17 @@
             "--name", self.package_name,
             "--mock-config", self.mock_config_file,
         ]
+        with cache_directory(self.resultdir) as cachedir:
+            try:
+                result = run_cmd(cmd, cwd=self.resultdir, env={"XDG_CACHE_HOME": self.cachedir})
```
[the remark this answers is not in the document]
thebeanogamer: Fixed

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -10,10 +10,22 @@
...
 import os
 import shutil
+from contextlib import contextmanager
+
 from copr_rpmbuild.helpers import run_cmd
 from copr_rpmbuild.automation.base import AutomationTool
 
 
+@contextmanager
+def cache_directory(resultdir: str):
```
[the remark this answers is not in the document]
thebeanogamer: Fixed

rpmbuild/copr_rpmbuild/automation/fedora_review.py
```diff
@@ -67,6 +86,18 @@ def _filter_results_directory(self):
...
+
+        # The fedora-review command failed to start so badly that it didn't even create a log file.
+        # Given how early that should happen, this merits investigation.
+        if not os.path.exists(logfile):
+            self.log.error("Can't find fedora-review log: %s", logfile)
+            self.log.error("Please raise a bug on https://github.com/fedora-copr/copr including a link to this build")
+            # If the log file doesn't exist, there's zero chance the results exist
+            return
+
+        os.rename(logfile, os.path.join(dstdir, "fedora-review.log"))
```
praiskup, line 99: We need to compress this file (likely on backend, `_compress_live_logs` should be named `_compress_logs` and it should cover even this file.  Would you mind filling issue, or fixing directly?

praiskup:
Otherwise LGTM

praiskup:
Can you squash the commits into one?

rpmbuild/copr_rpmbuild/automation/fedora_review.py
thebeanogamer: See https://github.com/fedora-copr/copr/pull/2920/commits/a3b060fd8b39cc52b71f2a944eb35063f5be4ee4

thebeanogamer:
> Can you squash the commits into one?

Done

praiskup (approved):
LGTM, thank you!

PR #2923, opened 2023-09-20
Explicitly depend on python3-six

praiskup:
~~This means that python-copr-common is no longer compatible with Python 2 and EPEL 7, but we stay compatible with Python 3.6 on EL8+.~~

~~If we ever want make it compatible with EL7, we should even there use the Python 3 (python36) stack.~~

Just adding python-six into BuildRequires.


praiskup:
In theory, python-six should be in requires, too.  But I don't think we have to concentrate on this topic right now.

FrostyX (approved):

Issue #2927, opened 2023-09-26
Delete after days has max 60 days limit


praiskup:
Per error `Error: delete_after_days: Number must be between 0 and 60`

Python team would like to use this feature to auto-delete their rather large `@python` projects in Fedora Copr say after 700 days.

jtranquilli:
I could address this issue if someone assigns it to me.

PR #2930, opened 2023-09-26
Updated config.py to address issue #2927

jtranquilli:
delete_after_days was previously hard-coded to be 60, changed it to a much larger value, 700 as requested in the issue.
From what I can tell, the error message is not hard-coded, and the change in value of delete_after_days is now reflected in the error message, i.e. the error would now be updated to: 
'Error: delete_after_days: Number must be between 0 and 700'
If you were to choose a value of greater than 700.

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -6,7 +6,7 @@
 %global tests_version 4
 %global tests_tar test-data-copr-backend
 
-%global copr_common_version 0.16.4.dev
+%global copr_common_version 0.20.1.dev1
 
 Name:       copr-backend
 Version:    1.172
--- a/backend/copr_backend/daemons/build_dispatcher.py
+++ b/backend/copr_backend/daemons/build_dispatcher.py
@@ -2,7 +2,7 @@
 BuildDispatcher related classes.
 """
 
-from copr_common.worker_manager import GroupWorkerLimit
+from copr_common.worker_manager import HashWorkerLimit
 from copr_backend.dispatcher import BackendDispatcher
 from copr_backend.rpm_builds import (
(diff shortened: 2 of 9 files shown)


frontend/coprs_frontend/coprs/config.py
```diff
@@ -148,7 +148,7 @@ class Config(object):
     CACHE_KEY_PREFIX = "copr_cache_"
 
     # Default value for temporary projects
-    DELETE_AFTER_DAYS = 60
+    DELETE_AFTER_DAYS = 700 #Changed from 60 to 700 at request of issue #2927
```
praiskup, line 151: Thank you for the patch.  I think you need to fix `forms.py` though - not the default value (I think default 60 days is quite OK).

jtranquilli: Very good catch, thanks for the response. I'll update line 559 of forms.py to reflect the change.

praiskup:
Can we have to limit rounded to say 1000 days, or 2x365 at least? :)

frontend/coprs_frontend/coprs/config.py
praiskup: +1, can you drop the first commit and rebase?

jtranquilli: I set the default value in config.py back to what it was before, and rounded the max value in forms.py  as requested, to be 730 days, which is 2 years. I'll drop the first 2 commits now.

jtranquilli:
@praiskup Sorry, I accidentally force pushed this pull request (?). 
Screenshot 2023-09-26 at 5 12 07 PM

I did implement the changes you asked for though.

PR #2938, opened 2023-10-03
How to enlarge RAID on copr-backend

praiskup:


raid-updates
```diff
@@ -0,0 +1,29 @@
+1. create two volumes and one new instance
+
+2. create raid1 on that instance -> and let it sync -> be careful about I/O
+
+    cfdisk -> create partitions, otherwise raid doesn't assemble itself :-(
```
xsuchy, line 5: set type to LINUX RAID

praiskup:
Will be here: https://docs.pagure.org/copr.copr/raid_on_backend.html

raid-updates
```diff
@@ -0,0 +1,29 @@
...
+
+    $ mdadm -Db /dev/md/2ndraid-dev
+
+3. pvcreate /dev/md/2ndraid-dev
+
+3. vgextend copr-backend-data /dev/md/2ndraid-dev
+
+4. lvextend -l +100%FREE /dev/copr-backend-data/copr-backend-data
+
+5. resize2fs /dev/copr-backend-data/copr-backend-data
```
praiskup, line 21: This takes more than 15minutes

praiskup:
PTAL

doc/raid_on_backend.rst
```diff
@@ -45,6 +48,52 @@ Attaching volume
...
+   might have troubles to auto-assemble the disk arrays::
+
+        cfdisk /dev/nvmeXn1
+        cfdisk /dev/nvmeYn1
+
+3. Create the ``raid1`` array on both the new **partitions**::
+
+        $ mdadm --create --name=raid-be-03 --verbose /dev/mdXYZ --level=1 --raid-devices=2 /dev/nvmeXn1p1 /dev/nvmeYn1p1
+
+   Wait till the new array is synchronized (may take hours or days).  Check the
```
FrostyX, line 71: I don't understand two things at this point.

- In the first step, why do we create two new volumes instead of just one?
- We are waiting for the new array to be synchronized. Is it an empty array without any data, right? Maybe you could change the sentence to "Wait till the new empty array" so it is absolutely clear. It really takes days to sync empty array?

FrostyX:
Just two questions, otherwise LGTM.

doc/raid_on_backend.rst
praiskup: > why do we create two new volumes instead of just one?

We do?  X and Y.

> Is it an empty array without any data, right? 

Yes.

> It really takes days to sync empty array?

It depends on the volume size and I/O speed (of the volumes, and actual instance type, which needs to handle twice as much during sync) .  Note we create 16T volumes.  https://raid.wiki.kernel.org/index.php/Initial_Array_Creation

praiskup: Thank you for the review btw! :) I tried to update the things you touched. ptal

FrostyX (approved):

PR #2948, opened 2023-10-11
Fix backend tests

FrostyX:
Some of the backend tests were broken after merging PR #2920

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -3,7 +3,7 @@
 %endif
 
 %global prunerepo_version 1.20
-%global tests_version 4
+%global tests_version 5
 %global tests_tar test-data-copr-backend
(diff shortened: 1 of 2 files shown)


FrostyX:
To fix the one remaining test, we could either merge https://github.com/fedora-copr/test-data-copr-backend/pull/3 or we can remove the assert

```
>           assert level <= logging.INFO
E           assert 30 <= 20
E            +  where 20 = logging.INFO
```

praiskup:
The test PR merged.

FrostyX:
Updated, the tests pass now.

praiskup:
Does it make sense to have those commits separated?  Seems like they are closely related to each other.

FrostyX:
Squashed

PR #2954, opened 2023-10-14
Improve the fallback for generating GPG keys

### LINKED ISSUE #2911: Failed to create key-pair for user
The satellite team (@ehelms) is frequently experiencing the following issue in the last week:

Backend log:

```
[2023-09-13 12:21:34,197][ ERROR][PID:2646164] Build failed: Failed to create key-pair for user: @theforeman, project:foreman-nightly-staging-scratch-1e7659c1-fab9-598e-b6dc-391ce0046763 with error: Attempt to talk to server timeouted (we gave it 5 attempts)
request to copr-keygen: {'url': 'http://54.83.48.73/gen_key', 'data': {'name_real': '@theforeman_foreman-nightly-staging-scratch-1e7659c1-fab9-598e-b6dc-391ce0046763', 'name_email': '@theforeman#foreman-nightly-staging-scratch-1e7659c1-fab9-598e-b6dc-391ce0046763@copr.fedorahosted.org'}, 'method': 'post'}

[2023-09-13 12:21:34,198][  INFO][PID:2646164] Finished build: id=6400124 failed=True timeout=108000 destdir=/var/lib/copr/public_html/results/@theforeman/foreman-nightly-staging-scratch-1e7659c1-fab9-598e-b6dc-391ce0046763 chroot=srpm-builds 
```

Example build:
https://copr.fedorainfracloud.org/coprs/g/theforeman/foreman-nightly-staging-scratch-1e7659c1-fab9-598e-b6dc-391ce0046763/build/6400124/

Is there some workaround in case something like this happens? I am not aware of one.

### LINKED ISSUE #2942: GPG keys are not generated quickly enough to sign SRPM build artifacts
This is done async from the (first) build: https://github.com/fedora-copr/copr/blob/c381e3434ee617dbab4effcf347ddbe06b33bb98/backend/copr_backend/actions.py#L144-L154 

And it may take more time then the actual SRPM build.

a) do we need to sign SRPM build artifacts?  What for, it is just a transport mechanism into copr dist-git

b) or can we delay the srpm build until the GPG key is generated?  it typically takes a few seconds but still

FrostyX:
Fix #2911
Fix #2942

When frontend is creating a new project, it creates an action to generate GPG keys. This may fail because of a temporary keygen downtime or network issues.

Once a SRPM is built, we try to use the GPG key and it doesn't exist, there is a fallback, trying to generate the key again. We can try longer.

To trigger this fallback, you can do:

    docker-compose stop keygen-httpd
    copr create test-srpm-key --chroot fedora-38-x86_64
    copr buildscm --clone-url ... test-srpm-key
    # Wait few minutes until the action fails
    docker-compose start keygen-httpd

### DIFF
--- a/backend/copr_backend/sign.py
+++ b/backend/copr_backend/sign.py
@@ -167,7 +167,7 @@ def sign_rpms_in_dir(username, projectname, path, chroot, opts, log):
     try:
         get_pubkey(username, projectname, log, opts.sign_domain)
     except CoprSignNoKeyError:
-        create_user_keys(username, projectname, opts)
+        create_user_keys(username, projectname, opts, try_indefinitely=True)
 
     errors = []  # tuples (rpm_filepath, exception)
     for rpm in rpm_list:
@@ -185,7 +185,7 @@ def sign_rpms_in_dir(username, projectname, path, chroot, opts, log):
                             .format([err[0] for err in errors]))
 
 
-def create_user_keys(username, projectname, opts):
+def create_user_keys(username, projectname, opts, try_indefinitely=False):
     """
     Generate a new key-pair at sign host
 
@@ -204,7 +204,7 @@ def create_user_keys(username, projectname, opts):
     keygen_url = "http://{}/gen_key".format(opts.keygen_host)
     query = dict(url=keygen_url, data=data, method="post")
     try:
-        request = SafeRequest(log=log)
+        request = SafeRequest(log=log, try_indefinitely=try_indefinitely)
         response = request.send(**query)
     except Exception as e:
         raise CoprKeygenRequestError(


FrostyX:
This is not a direct fix for the mentioned issues, maybe the keygen shouldn't be so slow, maybe the action should be prioritized more. But this should work.

I need to test one more thing though - If two builds trigger the fallback at the same time, one of them creates the key, we need to make sure that the other one is not in an infinite loop until the timeout ends.

FrostyX:
> I need to test one more thing though - If two builds trigger the fallback at the same time, one of them creates the key, we need to make sure that the other one is not in an infinite loop until the timeout ends.

Nope, doesn't happen. Just tested on STG. Once the keygen went up, all the builds signed immediately and succeeded the SRPM phase.

backend/copr_backend/sign.py
```diff
@@ -167,7 +167,11 @@ def sign_rpms_in_dir(username, projectname, path, chroot, opts, log):
     try:
         get_pubkey(username, projectname, log, opts.sign_domain)
     except CoprSignNoKeyError:
-        create_user_keys(username, projectname, opts)
+        # We don't want to try in indefinitely because that could lead to
+        # forver stuck builds but we can afford trying to create the key for
+        # several hours
+        timeout = 60 * 60 * 10
+        create_user_keys(username, projectname, opts, timeout=timeout)
```
praiskup, line 174: This is weird.  What is the default behavior without the timeout, would that wait indefinitely?  Why not wait indefinitely till the keygen is fixed (it needs to be fixed sooner or later)?  Perhaps I just need a cup of coffee :-)

praiskup: Btw., since we create the key -> I'd expect that we should re-try doing the `get_pubkey` right after that.  Doesn't this lead to the situation without `pub` key in resultdir?  If not, can you in-line comment why? :)

FrostyX: So, the default behavior without specifying timeout is this:

```python
class SafeRequest:
    def __init__(..., try_indefinitely=False, timeout=2 * 60):
```

>  I'd expect that we should re-try doing the get_pubkey right after that.

IMHO the `get_pubkey` is only a check and we don't use its value. It is just important, that the key exists when `_sign_one` is called.

Now ... if we do a several hours timeout as I am currently proposing, it could still fail that the key is missing, as it fails now. But after several hours, what is the chance? If we would do `try_indefinitely=True` instead of a timeout, we can be sure that the key will exist.

If you are not worried that `try_indefinitely=True` might cause forever running builds for some obscure reason, we can do that instead :-)

praiskup: Ah, there's the default 120s limit, I missed that.  Yes, then what you propose is better definitely, and I am also for trying `try_indefinitely` :-) indefinitely running job would only force us to observe and fix other bugs.

FrostyX:
Updated

praiskup (approved):
LGTM, thank you.

PR #2956, opened 2023-10-16
Use tito for builds and share the actions

lachmanfrantisek:


lachmanfrantisek:
Looks like it's failing on the explicit `setup.py` version check in specfiles:

```
RPM build errors:
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ cd python-copr-common-git-31.739a509
+ rm -rf /builddir/build/BUILD/python-copr-common-git-31.739a509-SPECPARTS
+ /usr/bin/mkdir -p /builddir/build/BUILD/python-copr-common-git-31.739a509-SPECPARTS
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
+ grep '"0.20"' setup.py
```

praiskup:
> Looks like it's failing on the explicit setup.py version check in specfiles:

Yes, that's what tito does. If you do `tito build test --srpm`, both the Version in spec file file is bumped and the version in `setup.py`.

lachmanfrantisek:
@praiskup Good to know. How should I incorporate this with the existing `tito build --tgz --test -o .` action?

praiskup:
> @praiskup Good to know. How should I incorporate this with the existing tito build --tgz --test -o . action?

Good question, but dunno :-) when I tested such a generated tarball, it seemed correct to me (the correct version was dumped into the archived `setup.py` file).

praiskup:
It rather feels like the spec file generator and tarball generator are too disconnected;  tarball is correct, but the `Version` tag in spec file is not.

praiskup:
Well, if this fixes the problem ... but I'd be much more happy if we well-integrated Tito into packit, and we could use the tito versioning scheme.  It is mostly cosmetics, but if someone is used to work with Tito, having completely different versions feels weird.

common/python-copr-common.spec
```diff
@@ -16,7 +16,7 @@
 %endif
 
 Name:       python-copr-common
-Version:    0.20.1.dev1
+Version:    1.130
```
praiskup, line 19: Weird ... :-) this seems like cross-package version leak

lachmanfrantisek: Ups, I didn't want to commit this...;)

lachmanfrantisek:
> Well, if this fixes the problem ... but I'd be much more happy if we well-integrated Tito into packit, and we could use the tito versioning scheme. It is mostly cosmetics, but if someone is used to work with Tito, having completely different versions feels weird.

Yes, it would be better. I thought that `git describe --match $PACKIT_CONFIG_PACKAGE_NAME-[0-9]* --abbrev=0 HEAD | egrep -o  [0-9]+\.[0-9]+` should provide the same version as tito.

Also, I didn't find a way how to update `setup.py` through tito. I see it only in `tito tag` which we probably don't want to use here...

common/python-copr-common.spec
lachmanfrantisek: Ok, that was tito who committed that... I should be more careful about playing with that..;)

lachmanfrantisek:
OK, I've changed the approach to reading the version from the specfile so I don't need to change anything. Packit will just edit the release field. I hope this is fine.

praiskup:
> I thought that git describe --match ... should provide the same version as tito.

I think we should create a dedicated command in Tito for this, if there's no such command, yet.

> Also, I didn't find a way how to update setup.py through tito.

Well, Tito never touches the files in the current git repo, unless you do a new release (in which case it is not a temporary change, but a committed change).  Do you want just a temporary local change? (not sure Tito knows this, @FrostyX?).  Tito normally modifies the file "on the fly" while generating the distribution tarball, in a temporary directory.

praiskup:
> OK, I've changed the approach to reading the version from the specfile so I don't need to change anything. Packit will just edit the release field. I hope this is fine.

Not sure.  The tarball generated by tito will have a bumped version in `setup.py`, but specfile is not generated by Tito.

praiskup:
We have `tito build --test --tgz` (and `--srpm`), I suppose we could also implement  `tito build --test --spec`?

praiskup:
Ok, after the Monday meeting, I think this is much better than before.  We basically checked that both `setup.py` and the spec file are not touched by our process (the version is used from git as-is).  There's a leftover https://github.com/rpm-software-management/tito/issues/460 that blocks the original issue, but that can be solved separately.

Can't this be merged?

praiskup:
> @packit-as-a-service rpm-build:fedora-39-x86_64:copr-messaging "In progress"

Still in progress after several days.  Is this a new bug?  I thought this should have been fixed already.

lachmanfrantisek:
> Still in progress after several days. Is this a new bug? I thought this should have been fixed already.

I need to take a look but it looks to me like we've hit a project limit for GitHub interactions since [the build](https://dashboard.packit.dev/results/copr-builds/1092585) itself went well.

lachmanfrantisek:
Looks like it was really a flake.

praiskup:
Nice!  Thank you.

PR #2958, opened 2023-10-18
doc: faq - how to handle autospec

xsuchy:
Taken from Diego Herrera from Fedora devel list


praiskup:
I probably need more info here :-) why is not the "built-in" DistGit method enough?

xsuchy:
udpated

praiskup:
+1

Issue #2757, opened 2023-06-05
keygen is relatively slow at certain times


praiskup:
E.g. today, 170 running jobs, and signatures take quite some time:
```
[2023-06-05 13:16:37,151][  INFO][PID:3960066] Calling '/bin/sign -u praiskup#prunerepo-dev@copr.fedorahosted.org -p' (attempt #1)
[2023-06-05 13:16:39,070][  INFO][PID:3960066] Calling '/bin/sign -4 -h sha256 -u praiskup#prunerepo-dev@copr.fedorahosted.org -r /var/lib/copr/public_html/results/praiskup/prunerepo-dev:pr:19/fedora-rawhide-aarch64/06004322-prunerepo/prunerepo-1.22-1.git.95.53510f5.fc39.noarch.rpm' (attempt #1)
[2023-06-05 13:16:43,908][  INFO][PID:3960066] signed rpm: /var/lib/copr/public_html/results/praiskup/prunerepo-dev:pr:19/fedora-rawhide-aarch64/06004322-prunerepo/prunerepo-1.22-1.git.95.53510f5.fc39.noarch.rpm
[2023-06-05 13:16:43,910][  INFO][PID:3960066] Calling '/bin/sign -4 -h sha256 -u praiskup#prunerepo-dev@copr.fedorahosted.org -r /var/lib/copr/public_html/results/praiskup/prunerepo-dev:pr:19/fedora-rawhide-aarch64/06004322-prunerepo/prunerepo-1.22-1.git.95.53510f5.fc39.src.rpm' (attempt #1)
[2023-06-05 13:16:50,703][  INFO][PID:3960066] signed rpm: /var/lib/copr/public_html/results/praiskup/prunerepo-dev:pr:19/fedora-rawhide-aarch64/06004322-prunerepo/prunerepo-1.22-1.git.95.53510f5.fc39.src.rpm
```

Well, seems like `/bin/sign -u` takes more than 2 minutes, not the "signing" action (which takes less than 10s).

praiskup:
Ah, that was a second, not minute. False alarm.

praiskup:
Reopening, seems like signatures might actually significantly delay the build processing on powerful builders (for kernel on EC2 `c7i.12xlarge` instance, build takes 52 minutes and signatures take 22 minutes).

praiskup:
I'm curious if the `sign -4` option actually slows things down.

praiskup:
[...]

PR #2963, opened 2023-10-23
backend: don't sign srpm-builds

praiskup:
This content is not meant to be consumed by users, and it consumes the keygen cycles.

Relates: #2757


FrostyX:
Why didn't we merge this at the meeting?

FrostyX (approved):

praiskup:
> Why didn't we merge this at the meeting?

Not sure, but no hurry ... we can merge after 24h anyway, no need to hurry.

Conan-Kudo:
This doesn't mean that the published SRPMs aren't signed, does it? Because unsigned artifacts that you can get from `dnf download` would not be great...

praiskup:
Correct, published SRPMs stay signed.

PR #2977, opened 2023-11-02
Rebuild and allow user SSH

FrostyX:
Implementation of https://github.com/fedora-copr/debate/pull/6

Fixes: #2364

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -10,10 +10,17 @@ import shutil
 import statistics
 import time
 import json
+import shlex
 
+from datetime import datetime
 from packaging import version
 
 from copr_common.enums import StatusEnum
+from copr_common.helpers import (
+    USER_SSH_DEFAULT_EXPIRATION,
+    USER_SSH_MAX_EXPIRATION,
+    USER_SSH_EXPIRATION_PATH,
+)
 
 from copr_backend.background_worker import BackendBackgroundWorker
 from copr_backend.cancellable_thread import CancellableThreadTask
@@ -51,6 +58,9 @@ MESSAGES = {
 
 COMMANDS = {
     "rpm_q_builder": "rpm -q copr-rpmbuild --qf \"%{VERSION}\n\"",
+    "echo_authorized_keys": "echo {0} >> /root/.ssh/authorized_keys",
+    "set_expiration": "echo -n {0} > " + USER_SSH_EXPIRATION_PATH,
+    "cat_expiration": "cat {0}".format(USER_SSH_EXPIRATION_PATH),
 }
 
 
@@ -139,6 +149,7 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         self.builder_livelog = os.path.join(self.builder_dir, "main.log")
         self.builder_results = os.path.join(self.builder_dir, "results")
         self.ssh = None
+        self.root_ssh = None
         self.job = None
         self.host = None
         self.canceled = False
@@ -307,6 +318,11 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         """
         Parse `results.json` and update the `self.job` object.
         """
+        # When user SSH is allowed, we don't download any results from the
+        # builder for safety reasons. Don't try to parse anything.
+        if self.job.ssh_public_keys:
+            return
+
         path = os.path.join(self.job.results_dir, "results.json")
         if not os.path.exists(path):
             raise BackendError("results.json file not found in resultdir")
@@ -589,12 +605,19 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         """
         Retry rsync-download the results several times.
         """
+        filter_ = None
+        if self.job.ssh_public_keys:
+            self.log.info("Builder allowed user SSH, not downloading the "
+                          "results for safety reasons.")
+            filter_ = ["+ success", "+ *.spec", "- *"]
+
         self.log.info("Downloading results from builder")
         self.ssh.rsync_download(
             self.builder_results + "/",
             self.job.results_dir,
             logfile=self.job.rsync_log_name,
             max_retries=2,
+            filter_=filter_,
         )
 
     def _check_build_success(self):
@@ -683,6 +706,9 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         """
         self.log.info("Listing built binary packages in %s", job.results_dir)
 
+        if self.job.ssh_public_keys:
+            return ""
+
         # pylint: disable=unsubscriptable-object
         assert isinstance(self.job.results, dict)
 
@@ -740,6 +766,123 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         self.log.info("Added pubkey for user %s project %s into: %s",
                       user, project, pubkey_path)
 
+    @skipped_for_source_build
+    def _setup_for_user_ssh(self):
+        """
+        Setup the builder for user SSH
+        https://github.com/fedora-copr/debate/tree/main/user-ssh-builders
+
+        If the builder setup for user SSH becomes more complicated than just
+        installing the public key, we might want to move the code to a script
+        within `copr-builder` and call it here or from `copr-rpmbuild`. There
+        is no requirement for it to be here.
+        """
+        if not self.job.ssh_public_keys:
+            return
+        self._alloc_root_ssh_connection()
+        self._deploy_user_ssh()
+        self._set_default_expiration()
+
+    def _alloc_root_ssh_connection(self):
+        self.log.info("Allocating root ssh connection to builder")
+        self.root_ssh = SSHConnection(
+            user="root",
+            host=self.host.hostname,
+            config_file=self.opts.ssh.builder_config,
+            log=self.log,
+        )
+
+    def _deploy_user_ssh(self):
+        """
+        Deploy user public key to the builder, so that they can connect via SSH.
+        """
+        pubkey = shlex.quote(self.job.ssh_public_keys)
+        cmd = COMMANDS["echo_authorized_keys"].format(pubkey)
+        rc, _out, _err = self.root_ssh.run_expensive(cmd)
+        if rc != 0:
+            self.log.error("Failed to deploy user SSH key for %s",
+                           self.job.project_owner)
+            return
+        self.log.info("Deployed user SSH key for %s", self.job.project_owner)
+
+    def _set_default_expiration(self):
+        """
+        Set the default expiration time for the builder
+        """
+        default = self.job.started_on + USER_SSH_DEFAULT_EXPIRATION
+        cmd = COMMANDS["set_expiration"].format(shlex.quote(str(default)))
+        rc, _out, _err = self.root_ssh.run_expensive(cmd)
+        if rc != 0:
+            # This only affects the `copr-builder show` command to print unknown
+            # remaining time. It won't affect the backend in terminating the
+            # buidler when it is supposed to
+            self.log.error("Failed to set the default expiration time")
+            return
+        self.log.info("The expiration time was set to %s", default)
+
+    def _builder_expiration(self):
+        """
+        Find the user preference for the builder expiration.
+        """
+        rc, out, _err = self.root_ssh.run_expensive(
+            COMMANDS["cat_expiration"], subprocess_timeout=60)
+        if rc == 0:
+            try:
+                return datetime.fromtimestamp(float(out))
+            except ValueError:
+                pass
+        self.log.error("Unable to query builder expiration file")
+        return None
+
+    def _keep_alive_for_user_ssh(self):
+        """
+        Wait until user releases the VM or until it expires.
+        """
+        if not self.job.ssh_public_keys:
+            return
+
+        # We are calculating the limits from when the job started but we may
+        # want to consider starting the watch when job ends.
+        default = datetime.fromtimestamp(
+            self.job.started_on + USER_SSH_DEFAULT_EXPIRATION)
+        maxlimit = datetime.fromtimestamp(
+            self.job.started_on + USER_SSH_MAX_EXPIRATION)
+
+        # Highlight this portion of the log because it is the only part of
+        # the backend.log that is directly for the end users
+        self.log.info("\n\nKeeping builder alive for user SSH")
+        self.log.info("The owner of this build can connect using:")
+        self.log.info("ssh root@%s", self.host.hostname)
+        self.log.info("Unless you connect to the builder and prolong its "
+                      "expiration, it will be shut-down in %s",
+                      default.strftime("%Y-%m-%d %H:%M"))
+        self.log.info("After connecting, run `copr-builder help' for "
+                      "complete instructions\n\n")
+
+        def _keep_alive():
+            while True:
+                if self.canceled:
+                    self.log.warning("Build canceled, VM will be shut-down soon")
+                    break
+                expiration = self._builder_expiration() or default
+                if datetime.now() > expiration:
+                    self.log.warning("VM expired, it will be shut-down soon")
+                    break
+                if datetime.now() > maxlimit:
+                    msg = "VM exceeded max limit, it will be shut-down soon"
+                    self.log.warning(msg)
+                    break
+                time.sleep(60)
+
+        CancellableThreadTask(
+            _keep_alive,
+            self._cancel_task_check_request,
+            self._cancel_running_worker,
+            check_period=CANCEL_CHECK_PERIOD,
+        ).run()
+        if self.canceled:
+            raise BuildCanceled
+
     def build(self, attempt):
         """
         Attempt to build.
@@ -754,6 +897,7 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         self._fill_build_info_file()
         self._cancel_if_requested()
         self._mark_running(attempt)
+        self._setup_for_user_ssh()
         self._start_remote_build()
         transfer_failure = CancellableThreadTask(
             self._transfer_log_file,
@@ -766,6 +910,8 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         if transfer_failure:
             raise BuildRetry("SSH problems when downloading live log: {}"
                              .format(transfer_failure))
+
+        self._keep_alive_for_user_ssh()
         self._download_results()
         self._drop_host()
 
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -240,6 +240,8 @@ def _get_limits_conf(parser):
         parser, "backend", "builds_max_workers_sandbox", 10, mode="int")
     limits['owner'] = _get_conf(
         parser, "backend", "builds_max_workers_owner", 20, mode="int")
+    limits['userssh'] = _get_conf(
+        parser, "backend", "builds_max_userssh", 2, mode="int")
     return limits
 
 
--- a/backend/copr_backend/rpm_builds.py
+++ b/backend/copr_backend/rpm_builds.py
@@ -150,6 +150,25 @@ class ArchitectureUserWorkerLimit(HashWorkerLimit):
         )
 
 
+class UserSSHLimit(HashWorkerLimit):
+    """
+    Limit the number of builders that allow user SSH
+    """
+    def __init__(self, limit):
+        def hasher(x):
+            # We don't allow user SSH for SRPM builds, returning None will
+            # make this unlimited
+            if x.source_build:
+                return None
+
+            # Don't limit builds that doesn't allow user SSH
+            # pylint: disable=protected-access
+            if not x._task.get("allow_user_ssh"):
+                return None
+            return x.owner
+        super().__init__(hasher, limit, name="userssh")
+
+
 class BuildTagLimit(PredicateWorkerLimit):
     """
     Limit the amount of concurrently running builds per given build tag.
--- a/backend/copr_backend/sshcmd.py
+++ b/backend/copr_backend/sshcmd.py
@@ -213,7 +213,7 @@ class SSHConnection:
         return "{}@{}:{}".format(self.user, host, src)
 
     def rsync_download(self, src, dest, logfile=None, max_retries=0,
-                       subprocess_timeout=None):
+                       subprocess_timeout=None, filter_=None):
         """
         Run rsync over pre-allocated socket (by the config)
 
@@ -231,9 +231,9 @@ class SSHConnection:
         directory needs to exist.
         """
         self._retry(self._rsync_download, max_retries, src, dest, logfile,
-                    subprocess_timeout)
+                    subprocess_timeout, filter_)
 
-    def _rsync_download(self, src, dest, logfile, subprocess_timeout):
+    def _rsync_download(self, src, dest, logfile, subprocess_timeout, filter_):
         ssh_opts = "ssh"
         if self.config_file:
             ssh_opts += " -F " + self.config_file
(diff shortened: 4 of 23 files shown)


praiskup:
Per @lachmanfrantisek claim, @nforro wants to be the early adopter! :-)

FrostyX:
There is still work to be done but can you please do a preliminary review @praiskup? I don't want to sink time in fine-tuning the details without knowing, that the general idea looks good.

What remains to be done:

- There is a lot of valid linter errors that need to be fixed
- Obtaining SSH keys from FAS (or fallback to a general textarea) is not there yet
- I only tested in the docker-compose so far. Still requires some STG testing
- Small bugs (mainly `copr-builder show`)
- Some conceptual questions code TODOs, I am very much interested in your opinion

backend/copr_backend/background_worker_build.py
```diff
@@ -766,6 +857,9 @@ def build(self, attempt):
         if transfer_failure:
             raise BuildRetry("SSH problems when downloading live log: {}"
                              .format(transfer_failure))
+
+        self._keep_alive_for_user_ssh()
+        # TODO Should the results be downloaded for jobs with user SSH?
```
praiskup, line 862: If we have a reasonable file-list filter, yes.  Otherwise, NO is safer option at the beginning.

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -862,6 +865,8 @@ def add(cls, user, pkgs, copr, source_type=None, source_json=None,
             copr_dir=copr_dir,
             bootstrap=bootstrap,
             isolation=isolation,
+            allow_user_ssh=allow_user_ssh,
+            ssh_public_key=ssh_public_key,
```
praiskup, line 869: Would it be possible to have this in YAML/JSON format or something...?   We
could have things like:
:
```yaml
---
type: FAS
fas_user: xyzuser
```

```
---
type: plaintext
pub_key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDj5HauAOur/6+...
```

```
---
type: online
url: https://github.com/praiskup.keys
```

We don't have to implement all the options in the web-UI right now, nor in cli.
But we can be "prepared".

frontend/coprs_frontend/coprs/models.py
```diff
@@ -1444,6 +1450,8 @@ def sandbox(self):
             # build separated from any other.
             submitter = uuid.uuid4()
 
+        # TODO If allow_user_ssh, then add unique suffix
```
praiskup, line 1453: Yes please :)

rpmbuild/bin/copr-builder
```diff
@@ -0,0 +1,222 @@
...
+
+
+def cmd_help():
+    """
+    Print full instructions for working with the builder instance.
+    Ideally we would have this in MOTD but that would cause problems in cases
+    like `if $(ssh stroj cat foo|head -n1) == "foo"`. So instead, we instruct
+    users to run `copr-builder help` manually.
+    """
+    # TODO Should this be in the documentation instead?
```
praiskup, line 29: Not sure if "instead", but maybe "also"? :-)

rpmbuild/bin/copr-builder
```diff
@@ -0,0 +1,222 @@
...
+    print(
+        "You have been entrusted with root access to a Copr builder.\n"
+        "Please be responsible.\n"
+        "\n"
+        "This instance should be used only for the purposes of debugging\n"
+        "build failures of RPM packages. It is not a personal VPS :-).\n"
```
praiskup, line 35: I think we could be more "strict", without a joking tone, ala motd on Batcave:
```
This is a PRIVATE computer system, unauthorized access is
strictly prohibited.

This system is to be used for Copr and related purposes only.  This
is not your personal computing system.  Users who are unsure whether or not
they have access to this system, don't have it and should log off immediately.
```

rpmbuild/bin/copr-builder
```diff
@@ -0,0 +1,222 @@
...
+https://github.com/fedora-copr/debate/tree/main/user-ssh-builders
+"""
+
+import argparse
+from datetime import datetime, timedelta
+from contextlib import suppress
+from copr_common.helpers import timedelta_to_dhms
+
+
+RPMBUILD_PID_PATH = "/var/lib/copr-rpmbuild/pid"
```
praiskup, line 14: We seem to have this in config already (helpers.py)

backend/copr_backend/background_worker_build.py
```diff
@@ -740,6 +749,87 @@ def _add_pubkey(self):
...
+                return datetime.fromtimestamp(float(out))
+            except ValueError:
+                pass
+        self.log.error("Unable to query builder expiration file")
+
+    def _keep_alive_for_user_ssh(self):
+        """
+        Wait until user releases the VM or until it expires.
+        """
+        # TODO Should we use CancellableThreadTask for the implementation?
```
praiskup, line 810: That's definitely nice to have.

backend/copr_backend/background_worker_build.py
```diff
@@ -740,6 +749,87 @@ def _add_pubkey(self):
...
+        """
+        if not self.job.allow_user_ssh:
+            return
+        self._deploy_user_ssh()
+        self._set_default_expiration()
+
+    def _deploy_user_ssh(self):
+        """
+        Deploy user public key to the builder, so that they can connect via SSH.
+        """
```
praiskup, line 770: This is a nice first step, if we make sure there's no shell injection (shlex.quote again).  Long term it would be better to have a more controlled environment for root commands.

For some time we are facing subscription-manager headaches (subscribing also requires root access, and takes too much time during builder startup). It would be nice to be able to subscribe "on demand".  Or wait for the async subscription-manager action "on demand".  It would be nice to have a new "root" hook in resalloc doing these per-task preparation actions so we could drop the root access from Backend entirely.

frontend/coprs_frontend/coprs/auth.py
```diff
@@ -139,6 +139,20 @@ def login():
         # If the login is successful, we are redirected to function decorated
         # with the `@oid.after_login`
         team_req = TeamsRequest(["_FAS_ALL_GROUPS_"])
+
+        # TODO Ask for SSH public key. Anybody knows where is the list of
+        # available keys that we can ask for?
```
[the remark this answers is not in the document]
FrostyX: I agree, I will create a separate RFE for this.

frontend/coprs_frontend/coprs/logic/builds_logic.py
FrostyX: This sounds reasonable, not done yet.

frontend/coprs_frontend/coprs/models.py
FrostyX: Done

rpmbuild/bin/copr-builder
FrostyX: Good catch, fixed.

rpmbuild/bin/copr-builder
FrostyX: I mentioned the `copr-builder help` command in the Copr web UI, let's start from there. Maybe we will move it to the documentation page in the future.

rpmbuild/bin/copr-builder
FrostyX: Updated. I tried some middle ground between my half-joking tone and your super strict one :-)
I don't want to be too strict and provoke people to do bad stuff just to try if they can get away with it.

FrostyX (commented):
Thank you very much for the review. I updated almost everything but the PR is still WIP.

praiskup:
Not on a proper review yet, but a few notes after the first "hands on" [experience](https://copr.stg.fedoraproject.org/coprs/g/copr/TEST1701420778014423815-build-attributes/build/2916300/), some comments:
- I first wasn't sure where to ssh, I assume presenting the ssh command in the web-UI isn't a trivial (frontend has no idea about the machine IP/host), but dumping somewhere "check backend.log for more instructions" would is IMO necessary
- the build ended up in the `failed` state (OK) but the source build is still "running", and the machine has already been dealocated (I did not prolong it)
- the log [disappeared](https://download.copr-dev.fedorainfracloud.org/results/@copr/TEST1701420778014423815-build-attributes/srpm-builds/02916300/backend.log.gz) - which is again expected, but we have a non-working link there

praiskup:
- no motd ...
- backend.log contains no info about "prolonging"

praiskup:
- I noticed the instructions say "Public Key" but we should say "Public Keys" - making sure that users know that multiple keys may be passed

backend/copr_backend/helpers.py
```diff
@@ -240,6 +240,8 @@ def _get_limits_conf(parser):
         parser, "backend", "builds_max_workers_sandbox", 10, mode="int")
     limits['owner'] = _get_conf(
         parser, "backend", "builds_max_workers_owner", 20, mode="int")
+    limits['userssh'] = _get_conf(
+        parser, "backend", "builds_max_userssh", 2, mode="int")
```
praiskup, line 244: Might be tricky; we should warn the user that, when the ssh key is specified, only 2 parallel builders will be given to him and that the rest of builds stays in "pending" state.

backend/copr_backend/rpm_builds.py
```diff
@@ -150,6 +151,25 @@ def __init__(self, architecture, limit):
...
+class UserSSHLimit(HashWorkerLimit):
+    """
+    Limit the number of builders that allow user SSH
+    """
+    def __init__(self, _hasher, limit):
+        def predicate(x):
+            # We don't allow user SSH for SRPM builds, return unique string
+            # making this unlimited
+            if x.source_build:
+                return str(uuid.uuid4())
```
praiskup, line 163: I think we can simply return None, and the HashWorkerLimit will not count with those.
Btw., I noticed this because of the memory leak we have in StringCounter() (at least with random strings like this).

backend/copr_backend/rpm_builds.py
```diff
@@ -150,6 +151,25 @@ def __init__(self, architecture, limit):
...
+
+            # Not limit builds that doesn't allow user SSH
+            # pylint: disable=protected-access
```
praiskup, line 166: nit: "Don't limit builds that don't allow users to SSH"

backend/copr_backend/sshcmd.py
```diff
@@ -243,8 +244,23 @@ def _rsync_download(self, src, dest, logfile, subprocess_timeout):
...
+        command = [
+            "/usr/bin/rsync",
+            "-rltDvH",
+            "--chmod=D755,F644",
+            "-e", "'{}'".format(ssh_opts),
+        ]
+        for value in include or []:
+            command.extend(["--include", shlex.quote(value)])
+        for value in exclude or []:
+            command.extend(["--exclude", shlex.quote(value)])
```
praiskup, line 257: Hm, I'm not sure I like the `include` and `exclude` split in this method.  The semantics of these options in rsync is order dependant;  it would be nicer if we had just one argument named `filter`  to have more flexibility for the future.

backend/tests/test_config_reader.py
```diff
@@ -65,6 +66,7 @@ def test_correct_build_limits(self):
                 'ppc64le': 11,
                 's390x': 5,
             },
+            'userssh': 2,
```
praiskup, line 69: Can you add a config above, and change this to a non-default value?

frontend/coprs_frontend/alembic/versions/41763f7a5185_add_allow_user_ssh_column.py
```diff
@@ -0,0 +1,27 @@
...
+revision = '41763f7a5185'
+down_revision = 'ec3528516b0c'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('build', sa.Column('allow_user_ssh', sa.Boolean(),
+                                     server_default='0', nullable=False))
+    op.add_column('build', sa.Column('ssh_public_key', sa.Text()))
```
praiskup, line 22: Do we need both fields?

frontend/coprs_frontend/coprs/models.py
```diff
@@ -1440,13 +1445,20 @@ def sandbox(self):
...
-        submitter, _ = self.submitter
-        if not submitter:
-            # If we don't know build submitter, use "random" value and keep the
-            # build separated from any other.
-            submitter = uuid.uuid4()
-
-        return '{0}--{1}'.format(self.copr.full_name, submitter)
+        # If we don't know build submitter, use "random" value and keep the
+        # build separated from any other.
+        submitter = self.submitter[0] or str(uuid.uuid4())
```
praiskup, line 1450: is submitter[0] correct? not submitter?

frontend/coprs_frontend/coprs/models.py
```diff
@@ -1440,13 +1445,20 @@ def sandbox(self):
...
+        parts = [self.copr.full_name, submitter]
+
+        # If user SSH is allowed, use random suffix to keep the builder
+        # separated from other users and other builds of the same user
+        if self.allow_user_ssh:
+            # We don't give user SSH for SRPM builds, so there is no need for
+            # an unique sandbox
+            is_rpm_build = self.source_state in helpers.FINISHED_STATES
+            if self.allow_user_ssh and is_rpm_build:
+                parts.append(str(uuid.uuid4()))
```
praiskup, line 1460: This is imo unnecessarily complicated, we could somewhere at the beginning do:
```
if self.allow_user_ssh:
    submitter = str(uuid.uuid4())
```

frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_forms.html
```diff
@@ -215,8 +215,21 @@ <h3 class="panel-title">
...
+
+  {% if allow_user_ssh %}
+    <h3>SSH access to the builder</h3>
+    <p>
+      You will obtain a SSH access to the builder to easily debug your
+      package within the Copr infrastructure. After connecting, run
+      <code>copr-builder help</code> for complete instructions.
+    </p>
+    <input type="hidden" name="allow_user_ssh" value="{{ allow_user_ssh }}">
+    {{ render_field(form.ssh_public_key, label='Public SSH key', rows=8, cols=50) }}
```
praiskup, line 229: s/key/keys/, (newline separated)

frontend/coprs_frontend/coprs/models.py
praiskup: I was wrong, submitter[0] is correct.

frontend/coprs_frontend/coprs/logic/builds_logic.py
FrostyX: I thought about this and it is probably not needed now. We IMHO still want one big textarea for user input and then parse the different (types of) keys somewhere in our code. Personally, I would do the parsing on backend (we have workers there) or even on builder. Because we can fetch the keys from FAS or URL right after that.

Right now only plaintext and one key is supported, so the value could only be

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDj5HauAOur/6+.
```

Seeing some other comments in the PR, we will probably implement support for multiple keys in this PR, so newline would be separator, e.g.

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDj5HauAOur/6+.
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDj5HauAOur/6+.
```

Once we start to support other methods, the input could be

```
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAACAQDj5HauAOur/6+.
FAS:xyzuser
https://github.com/praiskup.keys
GitHub:frostyx
```

We can always save the blob to the database as is and parse it when needed. That would be consistent with how e.g. `copr://` external repositories work.

FrostyX:
Thank you very much for the review @praiskup,
I addressed all your comments. Everything is in a separate commit for the reviewing convenience, I will squash them before merging.

> no motd ...

I spent a couple of hours on this before Christmas and wasn't able to set up MOTD on the builders. And then @xsuchy suggested it might be a good idea not having MOTD because of cases like `if $(ssh ... cat foo|head -n1) == "foo"`. So I put in the instructions to run `copr-builder help` instead.

praiskup:
> if $(ssh ... cat foo|head -n1) == "foo"

But MOTD shouldn't affect stdout:

```
$ ssh batcave01.iad2.fedoraproject.org echo ahoj
ahoj
```

praiskup:
jakub to ping pavel and take a look at motd problems..

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: > Right now only plaintext and one key is supported

Huh, I've successfully tested before with multiple keys :-)

praiskup: > We IMHO still want one big textarea for user input and then parse the different (types of) keys somewhere in our code.

I'd prefer to have a separate text area in the future, rather than changing the field and start "guessing" the type of input.

My thought was to start doing it right away in the background -- aka giving the user a simple textarea for multi-line key-only input, converting it to the yaml thing in the background, and store into DB.  This way we would have open doors for later enhancements without touching the old entries.  But I agree that this is rather a nice to have thing, there are many such things.

backend/copr_backend/background_worker_build.py
```diff
@@ -307,6 +318,9 @@ def _parse_results(self):
         """
         Parse `results.json` and update the `self.job` object.
         """
+        if self.job.ssh_public_keys:
```
praiskup, line 321: The implication here deserves an inline comment :-)

backend/copr_backend/background_worker_build.py
```diff
@@ -683,6 +704,9 @@ def _collect_built_packages(self, job):
         """
         self.log.info("Listing built binary packages in %s", job.results_dir)
 
+        if self.job.ssh_public_keys:
+            return "Public SSH access was enabled, no results were stored"
```
praiskup, line 708: Isn't it safer to return empty dict here?

backend/copr_backend/sshcmd.py
```diff
@@ -243,8 +243,21 @@ def _rsync_download(self, src, dest, logfile, subprocess_timeout):
...
+            "-e", "'{}'".format(ssh_opts),
+        ]
+        for value in filter_ or []:
+            command.extend(["--filter", shlex.quote(value)])
+        command.extend([
+            full_source_path,
+            "{}/".format(dest),
+            "&>", log_filepath,
+        ])
+        command = " ".join(command)
```
praiskup, line 260: Nice!

backend/copr_backend/rpm_builds.py
```diff
@@ -150,6 +150,25 @@ def __init__(self, architecture, limit):
         )
 
 
+class UserSSHLimit(HashWorkerLimit):
+    """
+    Limit the number of builders that allow user SSH
+    """
+    def __init__(self, _hasher, limit):
```
praiskup, line 157: You can drop the unused hasher argument I think.

frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_forms.html
```diff
@@ -215,8 +215,31 @@ <h3 class="panel-title">
...
+      After connecting, run <code>copr-builder help</code> for instructions on
+      how to work with the builder.
+      <br>
+      There is a limit for max 2 builders wih SSH access per user. The rest of
+      builds will remain in the <code>pending</code> state until a slot is freed.
+    </p>
+    <input type="hidden" name="allow_user_ssh" value="{{ allow_user_ssh }}">
+    {{ render_field(form.ssh_public_keys, label='Public SSH keys',
+                    placeholder='Newline separated public SSH keys, e.g. ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDR+QU9...',
+                    rows=8, cols=50) }}
```
praiskup, line 239: This seems damn awesome! :-) Can't wait for testing it.

praiskup:
Small style issue:
Screenshot_20240118_164643

praiskup:
I'm a bit uncertain about the windows-like EOLs, in vim I see:
```
... praiskup.usersys.redhat.com^M
```

praiskup:
I tried to drop `.ssh/authorized_keys` file - checking what happens :-) - and it just finished the otherwise succeeding build.  Is this OK?

backend/copr_backend/background_worker_build.py
```diff
@@ -51,6 +58,9 @@
 
 COMMANDS = {
     "rpm_q_builder": "rpm -q copr-rpmbuild --qf \"%{VERSION}\n\"",
+    "echo_authorized_keys": "echo {0} >> /root/.ssh/authorized_keys",
+    "set_expiration": "echo -n {0} > " + USER_SSH_EXPIRATION_PATH,
+    "cat_expiration": "cat {0}".format(USER_SSH_EXPIRATION_PATH),
```
praiskup, line 63: I think we should document here that we must be especially careful about error handling; users may replace the `echo` or `cat` commands, and (in theory) cause damage to the backend machine. This deserves documenting IMO (here ore in a later PR).

backend/copr_backend/background_worker_build.py
```diff
@@ -589,12 +605,19 @@ def _download_results(self):
         """
         Retry rsync-download the results several times.
         """
+        filter_ = None
+        if self.job.ssh_public_keys:
+            self.log.info("Builder allowed user SSH, not downloading the "
+                          "results for safety reasons.")
+            filter_ = ["+ success", "+ *.spec", "- *"]
```
praiskup, line 612: I'm thinking ... what if user replaces the remote side (server) `rsync` binary.  Can that trick the backend to download some unwanted stuff?

Ideas?

praiskup:
Just a few comments above, otherwise it seems OK to squash and merge. The motd and other things can be done later.

FrostyX:
> Small style issue:

Do you want the instructions in the "SSH access to the builder" section to be indented to the right?

> I'm a bit uncertain about the windows-like EOLs, in vim I see:
> ... praiskup.usersys.redhat.com^M

Hmm, I haven't encountered this. Where do you see the ^M characters?

backend/copr_backend/background_worker_build.py
FrostyX: This looks quite dangerous. Should we maybe verify some checksum of the binary, or something like that? Same for the `cat` commands mentioned above? But maybe that leads to the same issue (i.e. we can't be sure that the checksum tool wasn't tampered with)?

backend/copr_backend/background_worker_build.py
```diff
@@ -740,6 +766,122 @@ def _add_pubkey(self):
...
+            # buidler when it is supposed to
+            self.log.error("Failed to set the default expiration time")
+            return
+        self.log.info("The expiration time was set to %s", default)
+
+    def _builder_expiration(self):
+        """
+        Find the user preference for the builder expiration.
+        """
+        rc, out, _err = self.root_ssh.run_expensive(COMMANDS["cat_expiration"])
```
FrostyX, line 827: Don't use `run_expensive` and check output length and/or timeout.

FrostyX: I couldn't use `run` because it doesn't return the output. So I kept `run_expensive` and added a timeout for it.

FrostyX:
> Small style issue:
Screenshot_2024-01-26_12-55-04

I changed the style a bit.

praiskup (approved):
Thank you!

Issue #2706, opened 2023-05-09
Document what needs to be audited


xsuchy:
It will be nice to have documentation that states:

* what credentials need to be rotated
* what to check when we audit who has an access

praiskup:
no progress

praiskup:
no progress

praiskup:
no progress

praiskup:
no progress

praiskup:
Work in progress, note I've filled Fedora Infra issue for better credential maintenance https://pagure.io/fedora-infrastructure/issue/11612

praiskup:
I still need to document the way to rotate those.  So far I just cleaned up ansible.git a bit:

https://pagure.io/fedora-infra/ansible/c/bb04ae62adf461216f488bf125e7fe0dc970330c
https://pagure.io/fedora-infra/ansible/c/8abbdbaea50843c0d1eb77dc4f7301d6fbf53d20
https://pagure.io/fedora-infra/ansible/c/4792241cfef8f5ba9cf9c17197a8368e0727e5f1
https://pagure.io/fedora-infra/ansible/c/63e981214948678c30fed1c65b624bffa5055fcb
https://pagure.io/fedora-infra/ansible/c/47b91aae1f837668fa7f6766c56fa7001d8a0055
https://pagure.io/fedora-infra/ansible/c/41af546c0d9d1a8b868b039ad00b7f152b66581b

I still need to request the corresponding removals from ``ansible-private.git``.

PR #2993, opened 2023-11-10
docs: Fedora Copr credentials

praiskup:
Fixes: #2706

### DIFF
--- a/doc/maintenance/credentials.rst
+++ b/doc/maintenance/credentials.rst
@@ -1,0 +1,215 @@
+.. _fedora_credentials:
+
+Fedora Copr credentials
+=======================
+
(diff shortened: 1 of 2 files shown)


FrostyX:
Thank you for the PR, looks very good.
In order to resolve #2706, we also need to describe

> what to check when we audit who has an access

That means, who has a SSH access to our servers, who has admin/commit rights to git, etc.

praiskup:
> what to check when we audit who has an access

This is still WIP, but yes.  I'll try to grep some link to the `sysadmin-main` group + our `root_auth_users` list.

FrostyX (approved):

PR #3023, opened 2023-11-23
frontend/oidc: add OIDC_UNIQUE_PREFERRED_USERNAMES

### LINKED ISSUE #3008: Copr uses non-standard `username` field of OIDC UserInfo
I ran into a hitch trying to use [Forgejo](https://forgejo.org) as an OpenID Connect provider for Copr: in two places ([here](https://github.com/fedora-copr/copr/blob/44aa094c5389eb8e9caf106f37ff85a4c9111a8d/frontend/coprs_frontend/coprs/auth.py#L395) and [here](https://github.com/fedora-copr/copr/blob/44aa094c5389eb8e9caf106f37ff85a4c9111a8d/frontend/coprs_frontend/coprs/views/misc.py#L118)) the code assumes that the UserInfo contains a field ("claim"?) named `username`. AFAICT (and I apologise if this is incorrect; I'm not familiar with OpenID) this is a non-standard Ipsilonism (see [OpenID Connect Core 1.0 §5.1](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)); at the very least, [Forgejo doesn't provide it](https://codeberg.org/forgejo/forgejo/src/commit/99a93025d27480aa5a6f7e923558a292dcbde28c/routers/web/auth/oauth.go#L269).

It's not obvious to me how to replace the reliance on `username`, though: [OIDCC §5.7](https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability) says of `preferred_username`

>Claims such as email, phone_number, and preferred_username and MUST NOT be used as unique identifiers for the End-User.

It happens to be the case, however, that Forgejo's value for `preferred_username` _is_ unique, and I can imagine that would probably be common for OIDC providers. Maybe there could be an option in `copr.conf` along the lines of `OIDC_UNIQUE_PREFERRED_USERNAMES`?

Bob131:
The current OpenID Connect code assumes the existence of a non-standard field in the UserInfo returned by the provider, limiting the ability to use non-Ipsilon OIDC providers; however, the standard forbids using corresponding standardised field (preferred_username) as a unique user identifier, a property on which Copr relies.

This commit adds a fall back path to the OIDC code that will use preferred_username anyway if the administrator asserts that the values will be unique in practice with a new configuration option.  The upshot of this change is that administrators can now use OIDC providers like Gitea/Forgejo.

Fixes #3008.

### DIFF
--- a/frontend/coprs_frontend/config/copr.conf
+++ b/frontend/coprs_frontend/config/copr.conf
@@ -160,6 +160,21 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
 # OIDC_SCOPES = "" # e.g. "openid username profile email"
 # OIDC_TOKEN_AUTH_METHOD="client_secret_post" # possible: client_secret_post, client_secret_basic, none
 
+# The "claim" (for our purposes: a piece of information in the
+# UserInfo returned by an OIDC provider) to use as a unique
+# username.  There are two supported values:
+#
+#  - "username" (the default): a non-standard Ipsilon extension
+#    that provides unique user names.
+#
+#  - "preferred_username": a claim that is specified in the OpenID
+#    Connect standard, but one that the standard forbids Relying
+#    Parties from using for this purpose since they are not
+#    guaranteed to be unique.  This option is offered regardless,
+#    since oftentimes this claim is unique to a specific user in
+#    practice.
+# OIDC_USERNAME_CLAIM = "username"
+
 # We have supported two types of OIDC client register
 # 1. dynamic register
 # 2. static register
--- a/frontend/coprs_frontend/coprs/oidc.py
+++ b/frontend/coprs_frontend/coprs/oidc.py
@@ -25,17 +25,10 @@ def oidc_enabled(config):
         app.logger.error("OIDC_LOGIN or OIDC_PROVIDER_NAME is empty")
         return False
 
-    if not config.get("OIDC_CLIENT"):
-        app.logger.error("OIDC_CLIENT is empty")
-        return False
-
-    if not config.get("OIDC_SECRET"):
-        app.logger.error("OIDC_SECRET is empty")
-        return False
-
-    if not config.get("OIDC_SCOPES"):
-        app.logger.error("OIDC_SCOPES is empty")
-        return False
+    for key in ["OIDC_CLIENT", "OIDC_SECRET", "OIDC_SCOPES"]:
+        if not config.get(key):
+            app.logger.error("%s is empty", key)
+            return False
 
     if not config.get("OIDC_TOKEN_AUTH_METHOD"):
         app.logger.warning("OIDC_SCOPES is empty, using default method: client_secret_basic")
--- a/frontend/coprs_frontend/coprs/oidc.py
+++ b/frontend/coprs_frontend/coprs/oidc.py
@@ -34,11 +34,32 @@ def oidc_enabled(config):
         app.logger.warning("OIDC_SCOPES is empty, using default method: client_secret_basic")
         config["OIDC_TOKEN_AUTH_METHOD"] = "client_secret_basic"
 
+    username_claim = config.get("OIDC_USERNAME_CLAIM")
+    if username_claim and \
+       not username_claim in ("username", "preferred_username"):
+        app.logger.error(
+            f"Invalid setting {repr(username_claim)} for OIDC_USERNAME_CLAIM, " +
+            "expected one of: \"username\", \"preferred_username\""
+        )
+        return False
+
     return config.get("OIDC_METADATA") or (
         config.get("OIDC_AUTH_URL")
         and config.get("OIDC_TOKEN_URL")
         and config.get("OIDC_USERINFO_URL")
     )
+
+
+def oidc_username_from_userinfo(config, userinfo):
+    """
+    Return a unique user name from UserInfo
+    """
+    try:
+        return userinfo[config.get("OIDC_USERNAME_CLAIM", "username")]
+    except KeyError as exc:
+        raise RuntimeError(
+            "Can't get unique username, see OIDC_USERNAME_CLAIM configuration docs"
+        ) from exc
 
 
 def init_oidc_app(app):
--- a/frontend/coprs_frontend/coprs/auth.py
+++ b/frontend/coprs_frontend/coprs/auth.py
@@ -11,6 +11,7 @@ from coprs import oid
 from coprs import app
 from coprs.exceptions import CoprHttpException, AccessRestricted
 from coprs.logic.users_logic import UsersLogic
+from coprs.oidc import oidc_username_from_userinfo
 
 
 class UserAuth:
@@ -391,8 +392,9 @@ class OpenIDConnect:
 
         zoneinfo = userinfo['zoneinfo'] if 'zoneinfo' in userinfo \
             and userinfo['zoneinfo'] else None
+        username = oidc_username_from_userinfo(app.config, userinfo)
 
-        user = UserAuth.get_or_create_user(userinfo['username'], userinfo['email'], zoneinfo)
+        user = UserAuth.get_or_create_user(username, userinfo['email'], zoneinfo)
         GroupAuth.update_user_groups(user, OpenIDConnect.groups_from_userinfo(userinfo))
         return user
 
--- a/frontend/coprs_frontend/coprs/config.py
+++ b/frontend/coprs_frontend/coprs/config.py
@@ -175,6 +175,21 @@ class Config(object):
     # OIDC is opt-in
     OIDC_LOGIN = False
 
+    # The "claim" (for our purposes: a piece of information in the
+    # UserInfo returned by an OIDC provider) to use as a unique
+    # username.  There are two supported values:
+    #
+    #  - "username" (the default): a non-standard Ipsilon extension
+    #    that provides unique user names.
+    #
+    #  - "preferred_username": a claim that is specified in the OpenID
+    #    Connect standard, but one that the standard forbids Relying
+    #    Parties from using for this purpose since they are not
+    #    guaranteed to be unique.  This option is offered regardless,
+    #    since oftentimes this claim is unique to a specific user in
+    #    practice.
+    OIDC_USERNAME_CLAIM = "username"
+
     PACKAGES_COUNT = False
 
     EXTRA_BUILDCHROOT_TAGS = []
--- a/frontend/coprs_frontend/coprs/views/misc.py
+++ b/frontend/coprs_frontend/coprs/views/misc.py
@@ -17,7 +17,7 @@ from coprs.logic.users_logic import UsersLogic
 from coprs.exceptions import ObjectNotFound
 from coprs.measure import checkpoint_start
 from coprs.auth import FedoraAccounts, UserAuth, OpenIDConnect
-from coprs.oidc import oidc_enabled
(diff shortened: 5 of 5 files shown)


Bob131:
Testing this is kind of a pain in the neck. This is my attempt at a rough guide to reproduce my testing environment:

1. Apply the following patch:
```diff
diff --git a/docker-compose.dev.yaml b/docker-compose.dev.yaml
index d8eec3de..d55d3486 100644
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -11,24 +11,24 @@ services:
   backend-log:
     environment:
       - PYTHONPATH=/opt/copr/backend
-    command: /usr/sbin/runuser -u copr -g copr -- /opt/copr/backend/run/copr_run_logger.py
+    command: /opt/copr/backend/run/copr_run_logger.py
 
   backend-build:
     environment:
       - PYTHONPATH=/opt/copr/backend
-    command: /usr/sbin/runuser -u copr -g copr -- /usr/bin/copr-run-dispatcher builds
+    command: /usr/bin/copr-run-dispatcher builds
 
   backend-action:
     environment:
       - PYTHONPATH=/opt/copr/backend
-    command: /usr/sbin/runuser -u copr -g copr -- /usr/bin/copr-run-dispatcher actions
+    command: /usr/bin/copr-run-dispatcher actions
 
   frontend:
     environment:
       - PYTHONPATH=/opt/copr/frontend/coprs_frontend
-    command: /usr/sbin/runuser -u copr-fe -g copr-fe -- /opt/copr/frontend/coprs_frontend/manage.py runserver -p 5000 -h 0.0.0.0 --without-threads --no-reload
+    command: sh -c 'while true; do /opt/copr/frontend/coprs_frontend/manage.py runserver -p 5000 -h 0.0.0.0 --without-threads --no-reload; done'
 
   distgit:
     environment:
       - PYTHONPATH=/opt/copr/dist-git
-    command: /usr/sbin/runuser -u copr-dist-git -g copr-dist-git -- /opt/copr/dist-git/run/importer_runner.py
+    command: /opt/copr/dist-git/run/importer_runner.py
diff --git a/docker-compose.forgejo.yaml b/docker-compose.forgejo.yaml
new file mode 100644
index 00000000..be543162
--- /dev/null
+++ b/docker-compose.forgejo.yaml
@@ -0,0 +1,12 @@
+version: '3'
+services:
+  forgejo:
+    image: codeberg.org/forgejo/forgejo:1.20.5-0
+    environment:
+      - USER_UID=1000
+      - USER_GID=1000
+    volumes:
+      - ./forgejo:/data
+    ports:
+      - '3000:3000'
+      - '2222:22'
diff --git a/frontend/coprs_frontend/coprs/oidc.py b/frontend/coprs_frontend/coprs/oidc.py
index d965186e..f22c054f 100644
--- a/frontend/coprs_frontend/coprs/oidc.py
+++ b/frontend/coprs_frontend/coprs/oidc.py
@@ -99,6 +99,7 @@ def init_oidc_app(app):
                 access_token_url=app.config.get("OIDC_TOKEN_URL"),
                 authorize_url=app.config.get("OIDC_AUTH_URL"),
                 userinfo_endpoint=app.config.get("OIDC_USERINFO_URL"),
+                jwks_uri=app.config.get("OIDC_JWKS_URL"),
                 client_kwargs=client_kwargs,
             )
     return oidc

```
2. Start up Forgejo: `podman compose -f docker-compose.yaml -f docker-compose.dev.yaml -f docker-compose.forgejo.yaml up forgejo`
3. Navigate to `localhost:3000` and click through the setup UI
4. Stop Forgejo and enable the OIDC provider by adding `ENABLE = true` under `[oauth]` in `forgejo/gitea/conf/app.ini`
5. Start up everything: `podman compose -f docker-compose.yaml -f docker-compose.dev.yaml -f docker-compose.forgejo.yaml up`
6. Create an OAuth application client and token from the Forgejo site administration UI. Add `http://localhost:5000/oidc_auth/` to the redirect URI whitelist.
7. Edit `/etc/copr/copr.conf` in `copr_frontend_1` to include:
```ini
OIDC_LOGIN = True
OIDC_PROVIDER_NAME = "asdf" # e.g "openEuler ID"
OIDC_CLIENT = "<OID client ID from Forgejo>"
OIDC_SECRET = "<OID client secret from Forgejo>"
OIDC_SCOPES = "openid username profile email" # e.g. "openid username profile email"
OIDC_TOKEN_AUTH_METHOD="client_secret_post" # possible: client_secret_post, client_secret_basic, none

# The mixing of hostnames is intentional and required
OIDC_AUTH_URL="http://localhost:3000/login/oauth/authorize"
OIDC_TOKEN_URL="http://forgejo:3000/login/oauth/access_token"
OIDC_USERINFO_URL="http://forgejo:3000/login/oauth/userinfo"
OIDC_JWKS_URL="http://forgejo:3000/login/oauth/keys"
```
8. Restart the frontend with `podman exec -ti copr_frontend_1 bash` and `kill 2 3`
9. Try logging in. It should fail with an exception that hints towards a misconfiguration.
10. Add `OIDC_UNIQUE_PREFERRED_USERNAMES = True` to `copr.conf`, restart the frontend and try again. It should work this time :slightly_smiling_face:

praiskup:
> Testing this is kind of a pain in the neck. This is my attempt at a rough guide to reproduce my testing environment:

I think @FrostyX may comment on the docker-compose scripts if https://docs.pagure.org/copr.copr/contribute.html isn't helpful.

I must say the change is pretty straight-forward and non-intrusive, I like it.

frontend/coprs_frontend/coprs/oidc.py
```diff
@@ -48,6 +48,28 @@ def oidc_enabled(config):
...
 
 
+def oidc_username_from_userinfo(config, userinfo):
+    """
+    Return a unique user name from UserInfo
+    """
+    # Assume Ipsilon-style UserInfo (i.e., has a non-standard username
+    # claim that is guaranteed unique)
+    username = userinfo.get("username")
+    if username is None:
```
praiskup, line 58: This is going to work, but I think it's better to not guess here ... what if the `username` suddenly appears in the output in some of the OIDC providers?

I think that it would be better to just:

```python
unique_user_key = config.get("OIDC_USERNAMES", "preferred_username")
try:
  if unique_user_key == "preferred_username":
      username = userinfo["preferred_username"]
  elif unique_user_key == "username":
      username = userinfo["username"]
except KeyError:
  raise RuntimeError("Can't get unique username, check OIDC_USERNAMES configuration docs")
```

praiskup:
NB, neither Fedora Copr nor Red Hat Copr (internal) have been using OIDC so far, it would be nice to let @pkking from openEuler comment on this PR

praiskup:
Is `@` character part of the `preferred_username`?

frontend/coprs_frontend/config/copr.conf
```diff
@@ -160,6 +160,15 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
...
 # OIDC_TOKEN_AUTH_METHOD="client_secret_post" # possible: client_secret_post, client_secret_basic, none
 
+# Copr expects the OpenID provider to provide unique usernames.  It
+# assumes Ipsilon-style UserInfo by default, which satisfies this
+# requirement; however, if you know that your OID provider uses unique
+# values for preferred_username, you can set this value to True to
+# enable Copr to use non-Ipsilon OID providers.
+# It is set to False by default since the OpenID Connect standard
+# forbids assuming that preferred usernames are unique.
+# OIDC_UNIQUE_PREFERRED_USERNAMES = False
```
FrostyX, line 170: Can you please add this option to `frontend/coprs_frontend/coprs/config.py` as well?

FrostyX (commented):
> I think @FrostyX may comment on the docker-compose scripts if https://docs.pagure.org/copr.copr/contribute.html isn't helpful.

Not sure if I can be of help here. I have no experience with Forgejo. When it comes to the Copr development, my workflow is kind of described here https://frostyx.cz/posts/copr-docker-compose-without-supervisord#running-services-from-git

frontend/coprs_frontend/config/copr.conf
Bob131: Sure

Bob131:
> Is `@` character part of the `preferred_username`?

I'm not sure what you mean. There were no `@`s in my testing, but the [OIDC spec §5.1](https://openid.net/specs/openid-connect-core-1_0.html#rfc.section.5.1) says of `preferred_username`

>	This value MAY be any valid JSON string including special characters such as @, /, or whitespace

frontend/coprs_frontend/coprs/oidc.py
Bob131: Yeah, that makes sense. I'll do that.

pkking:
LGTM, since the behavior is opt-in :)

Bob131:
This should address the suggestions that have been raised. With respect to @praiskup's suggested changes, I have kept `username` as the default rather than `preferred_username` because
1. Defaulting to behaviour that the standard explicitly disallows seems a bit iffy, and
2. I'm concerned about possibly breaking existing deployments

Bob131:
Having a look at the lint warnings, I'm not convinced that addressing them directly would improve the code; however, I'm unsure. Am I okay to just add lint suppression comments to squash those warnings?

FrostyX:
> Having a look at the lint warnings, I'm not convinced that addressing them directly would improve the code;

```
Error: PYLINT_WARNING:
coprs_frontend/coprs/oidc.py:16: R0911[too-many-return-statements]: oidc_enabled: Too many return statements (7/6)
```

There is a lot of repeated code like this

```python
if not config.get("OIDC_CLIENT"):
    app.logger.error("OIDC_CLIENT is empty")
    return False
```

We could do something like (in a separate commit):

```python
for key in ["OIDC_LOGIN", "OIDC_CLIENT", "OIDC_SECRET", "OIDC_SCOPES"]:
    if not config.get(key):
        app.logger.error("%s is empty", key)
        return False
```

but if you'd rather avoid changing the code, feel free to add `pylint: disable` statement.

```
Error: PYLINT_WARNING:
coprs_frontend/coprs/oidc.py:67:8: W0707[raise-missing-from]: oidc_username_from_userinfo: Consider explicitly re-raising using 'except KeyError as exc' and 'raise RuntimeError("Can't get unique username, see OIDC_USERNAME_CLAIM configuration docs") from exc'
```

This is easy to fix, we usually respect this warning.

Bob131:
> There is a lot of repeated code like this
> 
> ```python
> if not config.get("OIDC_CLIENT"):
>     app.logger.error("OIDC_CLIENT is empty")
>     return False
> ```
> 
> We could do something like (in a separate commit):
> 
> ```python
> for key in ["OIDC_LOGIN", "OIDC_CLIENT", "OIDC_SECRET", "OIDC_SCOPES"]:
>     if not config.get(key):
>         app.logger.error("%s is empty", key)
>         return False
> ```

Ah, that makes sense. I can do that.
 
> ```
> Error: PYLINT_WARNING:
> coprs_frontend/coprs/oidc.py:67:8: W0707[raise-missing-from]: oidc_username_from_userinfo: Consider explicitly re-raising using 'except KeyError as exc' and 'raise RuntimeError("Can't get unique username, see OIDC_USERNAME_CLAIM configuration docs") from exc'
> ```
> 
> This is easy to fix, we usually respect this warning.

Ack. Glad I asked :slightly_smiling_face:

FrostyX (approved):
Thank you very much for the changes.

PR #3030, opened 2023-11-27
frontend: use common enum constants for sharing descriptions

### LINKED ISSUE #3029: Use some common descriptions between Fask-restx fields and WTForms
Descriptions in flask-restx fields are often just copy-pasted from forms. Store them in some common place and use them for both from there to prevent inconsistency for users in the future (they will have two different descriptions for some thing which will be confusing)

This is follow-up from #2807 (to split the PR to smaller ones)

nikromen:
(this is follow-up from #2807 containing its commit to prevent the PR to become huge; not done yet, I just began the transition and I'll finish it once #2807 is merged)

TODO:
 - [ ] drop the commit from #2807

Fixes #3029


praiskup:
I must be honest and admit that, while C&P always hurts, I'm not in favor of doing this change "on our own".  Isn't there some flask plugin that would help us to do this a standard way?

frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_forms.html
```diff
@@ -161,15 +161,12 @@ <h3 class="panel-title">{{ counter('instructions') }}. Select chroots and other
...
+    {{ render_field(form.pypi_package_name, placeholder="{{ common_descriptions.PYPI_PACKAGE_NAME.description }}.") }}
     {{ render_field(form.pypi_package_version, placeholder="Optional - Version of the package PyPI") }}
 
     {{ render_field(
          form.spec_generator,
-         info="Tool for generating specfile from a PyPI package. The options "
-               "are full-featured <strong>pyp2rpm</strong> with cross "
-               "distribution support, and <strong>pyp2spec</strong> that is "
-               "being actively developed and considered to be the future."
+         info="{{ common_descriptions.SPEC_GENERATOR.description }}"
```
praiskup, line 169: Here we are losing formatting.

frontend/coprs_frontend/coprs/templates/coprs/detail/_package_forms.html
```diff
@@ -37,14 +37,14 @@ <h3 class="panel-title">{{ counter('instructions') }}. Provide the source</h3>
...
-{% macro render_webhook_rebuild(form) %}
+{% macro render_webhook_rebuild(form, common_descriptions) %}
   <div class="form-group">
     <label class="col-sm-2 control-label" for="textInput-markup">
       Auto-rebuild
     </label>
     <div class="col-sm-10">
       <input type="checkbox" name="webhook_rebuild" {% if form.webhook_rebuild.data == True %}checked="checked"{% endif %}/>
-      Auto-rebuild the package? (i.e. every commit or new tag)
+      {{ common_descriptions.SPEC_GENERATOR.description }}
```
praiskup, line 47: Only part of the message is moved here.

frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_forms.html
praiskup: Also note there are some form field metdata arguments we could perhaps use: https://github.com/fedora-copr/copr/blob/0b7d4587e5ddaf3185684b5c83aef7f18ce4fc97/frontend/coprs_frontend/coprs/forms.py#L581-L584

praiskup:
I think we should do some research first.  Would gettext/babel help us?

praiskup:
Also note that we c&p the messages into `copr-cli --help` output, and then to manual pages.

frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_forms.html
nikromen: > Here we are losing formatting.

that's true... perhaps this one is worth not unifying

PR #3032, opened 2023-11-27
frontend: Add backref relationship between Package and Build to silence sqlalchemy warnings

### LINKED ISSUE #2445: copr-frontend alter-chroot warning
During deactivation of chroot I got:
```
Running command: /usr/bin/copr-frontend alter-chroot --action deactivate fedora-35-i386
/usr/lib/python3.11/site-packages/flask_sqlalchemy/__init__.py:550: SAWarning: relationship 'Build.package' will copy column package.id to column build.package_id, which conflicts with relationship(s): 'Package.builds' (copies package.id to build.package_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="builds"' to the 'Build.package' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)
  mapper = orm.class_mapper(type)
```

nikromen:
Fixes #2445

### DIFF
--- a/frontend/coprs_frontend/coprs/models.py
+++ b/frontend/coprs_frontend/coprs/models.py
@@ -876,8 +876,6 @@ class Package(db.Model, helpers.Serializer, CoprSearchRelatedData):
     def validate_max_builds(self, field, value):
         return None if value == 0 else value
 
-    builds = db.relationship("Build", order_by="Build.id")
-
     # relations
     copr_id = db.Column(db.Integer, db.ForeignKey("copr.id"), index=True)
     copr = db.relationship("Copr", backref=db.backref("packages"))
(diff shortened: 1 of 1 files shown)


praiskup:
Is the testsuite failing because of the change?

FrostyX:
We already have this

```python
class Package(db.Model, helpers.Serializer, CoprSearchRelatedData):
    ...
    builds = db.relationship("Build", order_by="Build.id")
```

which IMHO causes the collision

nikromen:
> which IMHO causes the collision

Yes, thank you. Unluckily... I cannot make alembic work locally on my machine to make migration. Could it somehow be because of the move to F39?

nikromen:
When I run `alembic revision --autogenerate -m "something"` it just generates empty revision

praiskup (approved):
LGTM, thank you.

PR #3043, opened 2023-12-02
More precise description for manual createrepo

### LINKED ISSUE #908: Create repositories manually  should not run createrepo_c
Original issue: https://pagure.io/copr/copr/issue/908
Opened: 2019-08-02 13:00:33

This is a followup from https://pagure.io/copr/copr/issue/907

When I use the **Create repositories manually** option, copr still creates the repository, but it is not "publishing" it. I think that this is wasting resources and should not happen.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2019-08-02 15:07:32:
This is not a bug;  the feature is there to allow users to work on the repositories, without affecting repo consumers.  And once the repo is ready, regenerate the repo for all the new builds (and thus ship all the RPMs at once). 

---

#### [churchyard](https://accounts.fedoraproject.org/user/churchyard) commented at 2019-08-02 15:55:42:
In that case the bug is in the fields description.

> Repository meta data is normally refreshed after each build. If you want to do this manually, turn this option on.

Change to (something like):

> Repository meta data is normally refreshed and published after each build. If you want to publish metadata manually, turn this option on. Note that builds in this copr will always see refreshed metadata regardless of this option.

---

#### [praiskup](https://accounts.fedoraproject.org/user/praiskup) commented at 2019-08-04 11:04:17:
Ack, last `/usr/bin/createrepo_c --database --ignore-lock --local-sqlite --cachedir /tmp/ --workers 8 --update /var/lib/copr/public_html/results/@python/python3.8/fedora-rawhide-x86_64` run took more than 7 minutes.

---

#### [iucar](https://accounts.fedoraproject.org/user/iucar) commented at 2019-08-28 14:36:19:
+1 to this RFE. In my use case (iucar/cran), I generate the dependency tree, and then create a list of build batches, so that there are no interdependencies in a batch. This is actually needed, because Copr does not guarantee that any build will wait for any other (and that's fine). But then we should be able to

1. submit a batch of packages to build, so that none of them calls createrepo;
2. call createrepo when the batch finishes;
3. repeat.

In this way, resources are not wasted. There could be a separate option to publish or not the result from createrepo. But disabling createrepo altogether is required too (and that's what the current wording suggests).

Note that this also requires some way to call createrepo programatically, instead of the button in the web UI (see #960).

---

This is only documentation bug, the build-batch feature is not yet filled in this tracker;
feel free to submit one.

FrostyX:
Fix #908

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -579,9 +579,10 @@ class CoprForm(BaseForm):
 
     disable_createrepo = wtforms.BooleanField(default=False,
             label="Create repositories manually",
-            description="""Repository meta data is normally refreshed
-            after each build.  If you want to do this manually, turn
-            this option on.""",
+            description="""Repository metadata is normally refreshed
+            and published after each build. If you want to publish metadata
+            manually, turn this option on. Note that builds in this copr will
+            always see refreshed metadata regardless of this option.""",
             false_values=FALSE_VALUES)
 
     unlisted_on_hp = wtforms.BooleanField(


FrostyX:
The issue mentions:

> Triage time: We have the table with "when the createrepo_c is run". We should have an explicit documentation for the "Create repoistories manually" option for sure, and link the table.
> https://docs.pagure.org/copr.copr/createrepo.html

but the "manual createrepo event" is described there.

Instead, I went for the suggestion by @hroncok and improved the field description in project settings.

FrostyX:
Updated, PTAL.

doc/user_documentation.rst
```diff
@@ -563,6 +563,27 @@ only in Copr and ignored by Koji. It is easy to do::
   %endif
 
 
+Creating repositories manually
+------------------------------
+
+After a build is finished, Copr automatically adds its results to the project RPM
+repository. When maintaining a large software stack consiting from hundreds of
```
hroncok, line 570:
```suggestion
repository. When maintaining a large software stack consisting of hundreds of
```

FrostyX: Good catch, thank you. Updated.

Issue #3006, opened 2023-11-20
Move Fedora Copr community from IRC to Matrix?


praiskup:
Seems like Fedora Infrastructure suspended the plans to [recover IRC <-> Matrix bridge](https://pagure.io/fedora-infrastructure/issue/11460) - and the current status quo is *move to Matrix*.

My proposal:

- request admin rights on the Matrix channel
- let's fix our docs and front page to point to the Matrix channel
- 6-months IRC period and shutdown the IRC channel

praiskup:
> I'd try to find out what is the plan for other Fedora channels (e.g. #fedora-admin) and do the same.

Ok, the "shut down" was maybe bad wording, sorry.  The channel can not be just "shut-down" I'm afraid; but if there's a 6-months title claiming that the community has been moved to a Matrix channel, I wouldn't put additional requirements on the team to listen both on IRC and Matrix (no matter the #fedora-admin policy/pattern, which is anyway by far more important and wider discussion place).

praiskup:
IRC Title changed

Announcement sent: https://lists.fedorahosted.org/archives/list/copr-devel@lists.fedorahosted.org/thread/HIHEAEWY6MEZDK23EY4YXDRL5FJIOT5A/

PR #3047, opened 2023-12-05
docs: migration to Fedora N+2

praiskup:
WIP

PR #3051, opened 2023-12-07
Copr community is moving from IRC to Matrix

praiskup:
This is because of the IRC <-> Matrix bridge isn't working: https://communityblog.fedoraproject.org/matrix-to-libera-chat-irc-bridge-unavailable/

And it seems that Fedora teams are advised to move (certainly Fedora Infra team, corresponding meetings, etc. is):
https://pagure.io/fedora-infrastructure/issue/11460

Resolves: #3006
Complements: #3047

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -31,7 +31,7 @@ The main subsections of this wiki:
 Communication
 -------------
 
-Copr is discussed on #fedora-buildsys on libera.chat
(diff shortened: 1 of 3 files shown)


FrostyX (approved):
I wanted to point out some more mentions of the IRC channel but they are all covered in #3047 as mentioned in the description. LGTM.

PR #3052, opened 2023-12-07
keygen: try to export a key to find whether it exists

FrostyX:
This seems to be much faster than `--list-secret-keys` on our production instance.


praiskup:
Seems like the tests need some care;  otherwise looks good.

Issue #3073, opened 2023-12-20
ISE -  ValueError: status code is not a 3 digit integer


xsuchy:
I see this in production log:
```[Tue Dec 19 21:55:20.153057 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050] ERROR:coprs:Admin-only exception
[Tue Dec 19 21:55:20.153090 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050] Request: GET https://copr.fedorainfracloud.org/coprs/packit/cockpit-project-cockpit-podman-1512/repo/fedora-39/packit-cockpit-project-cockpit-podman-1512-fedora-39.repo
[Tue Dec 19 21:55:20.153102 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050] User: None
[Tue Dec 19 21:55:20.153112 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050] Traceback (most recent call last):
[Tue Dec 19 21:55:20.153122 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
[Tue Dec 19 21:55:20.153132 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050]     self.dialect.do_execute(
[Tue Dec 19 21:55:20.153142 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
[Tue Dec 19 21:55:20.153152 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050]     cursor.execute(statement, parameters)
[Tue Dec 19 21:55:20.153162 2023] [wsgi:error] [pid 2590221:tid 2590236] [remote 13.59.72.67:57050] psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "counter_stat_pkey"
[...]

PR #3074, opened 2023-12-20
fix race condition when creating new record for repo stat counter

xsuchy:
Previously we claimed that this method does commit. But it did just add() - and that does not commit. We do a commit now. And during a commit we check for integrity error. If that happens that means somebody inserted it just second ago. So lets try read again.

Fixes: #3073

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/stat_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/stat_logic.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm.exc import NoResultFound
 
 from coprs import app
@@ -40,11 +41,21 @@ class CounterStatLogic(object):
         try:
             csl = CounterStatLogic.get(name).one()
             csl.counter = CounterStat.counter + count
+            db.session.add(csl)
         except NoResultFound:
-            csl = CounterStatLogic.add(name, counter_type)
-            csl.counter = count
+            try:
+                csl = CounterStatLogic.add(name, counter_type)
+                csl.counter = count
+                db.session.add(csl)
+                db.session.commit()
+                db.session.refresh(csl)
+            except IntegrityError:
+                # race condition - someone was faster
+                # try again first block
+                csl = CounterStatLogic.get(name).one()
+                csl.counter = CounterStat.counter + count
+                db.session.add(csl)
 
-        db.session.add(csl)
         return csl
 
     @classmethod
--- a/frontend/coprs_frontend/coprs/logic/stat_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/stat_logic.py
@@ -40,23 +40,17 @@ class CounterStatLogic(object):
         """
         try:
             csl = CounterStatLogic.get(name).one()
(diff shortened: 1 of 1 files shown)


xsuchy:
Note that I did not tested this as I have no idea how to test it.

FrostyX:
> Note that I did not tested this as I have no idea how to test it.

Same here

> Previously we claimed that this method does commit. But it did just add() - and that does not commit. We do a commit now.

In this case, I would say the code is as expected and the docstring is outdated. I am looking at the usages of the `incr` method and we always call `db.session.commit()` right after it. We do the same thing in unit tests. There is only one place that doesn't

https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_rpmrepo.py#L146

isn't this our culprit?

praiskup:
Per the initial commit c1a5794e39b3d1ea433cccc1c1cb55e52b72fd4a  it seems that the method was never "dirty" actually, and it never committed.

praiskup:
> db.session.commit() right after it. We do the same thing in unit tests. There is only one place that doesn't
> isn't this our culprit?

This seems like the classic problem "get or create" race.  While incrementing the counter should be race-free, creating the object isn't.

I think we should separate the concerns of creating vs incrementing.  Ie. could we create the stats instance at the same time with the repository configuration (CoprChroot instance)?

praiskup:
> incrementing the counter should be race-free

Hm, but not with ORM, I'm afraid.  We should do things like `update t set c = c + 1`.

praiskup:
> I think we should separate the concerns of creating vs incrementing.

Namely, I don't think [this method](https://github.com/enricobarzetti/sqlalchemy_get_or_create/blob/master/sqlalchemy_get_or_create/__init__.py#L38) is worth it.  We just should have some form of `create_with_defaults()` method initializing the counter to 0 and doing the `commit()` (ignoring the possible IntegrityError).  And then atomically increment with `incr()`.

praiskup:
See the [docs related to sessions](https://docs.sqlalchemy.org/en/20/orm/session_events.html), namely:

>  The reason the [Connection](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection) is passed is that it is encouraged that simple SQL operations take place here, directly on the [Connection](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Connection), such as incrementing counters or inserting extra rows within log tables.

xsuchy:
Updated.
I removed the return() because the values were not used and therefore I could remove refresh().
It now use UPDATE statement.

xsuchy:
Updated. I removed one filter() that was not needed as name is already primary index.

praiskup (approved):
LGTM

frontend/coprs_frontend/coprs/logic/stat_logic.py
```diff
@@ -39,13 +40,17 @@ def incr(cls, name, counter_type, count=1):
         """
         try:
             csl = CounterStatLogic.get(name).one()
```
praiskup, line 42: Sorry, I was confused with this.  This is an unused variable, but that's a nit.

Issue #3058, opened 2023-12-13
ISE psycopg2.errors.UniqueViolation) duplicate key value violates unique constraint "ownername_copr_dir_uniq"


xsuchy:
I have seen this in Fedora's Copr log:

```
[Wed Dec 13 10:25:02.121061 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720] ERROR:coprs:Admin-only exception
[Wed Dec 13 10:25:02.121092 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720] Request: POST https://copr.fedorainfracloud.org/webhooks/custom-dir/%40mock/mock-pull-requests:pr:1270/9d6af83d-7f27-4091-8161-c1098ff5a6dc/mock/
[Wed Dec 13 10:25:02.121103 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720] User: None
[Wed Dec 13 10:25:02.121112 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720] Traceback (most recent call last):
[Wed Dec 13 10:25:02.121123 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
[Wed Dec 13 10:25:02.121135 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720]     self.dialect.do_execute(
[Wed Dec 13 10:25:02.121145 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/default.py", line 736, in do_execute
[Wed Dec 13 10:25:02.121171 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720]     cursor.execute(statement, parameters)
[Wed Dec 13 10:25:02.121180 2023] [wsgi:error] [pid 1882891:tid 1882896] [remote 20.97.191.37:30720] psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "ownername_copr_dir_uniq"
[...]

PR #3075, opened 2023-12-20
fix race condition when creating CoprDir

xsuchy:
We create new CoprDir but then call send_createrepo() and if somebody in meantime create CoprDir we got sqlalchemy.exc.IntegrityError and the end of send_createrepo when SQLAlchemy flushes.

Fixes: #3058


xsuchy:
Note that I did not tested this as I have no idea how to test it.

FrostyX:
> Note that I did not tested this as I have no idea how to test it.

Same here :-/

> and if somebody in meantime create CoprDir

That sounds reasonable to me. But wouldn't the `parallel curl` command from #3058 reproduce the issue? Maybe I used the `parallel` command incorrectly?

praiskup:
I'm curious if we could use this: https://github.com/enricobarzetti/sqlalchemy_get_or_create

PR #3083, opened 2024-01-03
count every download of repo file

xsuchy:
Previously the render_generate_repo_file was memoized() so subsequent downloads was not counted. Now the function is not memoized() and CounterStatLogic.incr() is always called and only rest of the function is memoized().

### DIFF
--- a/frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
+++ b/frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
@@ -876,8 +876,18 @@ def _render_external_repo_template(dep, copr_dir, mock_chroot, dep_idx):
                                  url=baseurl) + "\n"
 
 
-@cache.memoize(timeout=5*60)
 def render_generate_repo_file(copr_dir, name_release, arch=None):
+    name = helpers.get_stat_name(
+        CounterStatType.REPO_DL,
+        copr_dir=copr_dir,
+        name_release=name_release,
+    )
+    CounterStatLogic.incr(name=name, counter_type=CounterStatType.REPO_DL)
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
```diff
@@ -876,8 +876,17 @@ def _render_external_repo_template(dep, copr_dir, mock_chroot, dep_idx):
...
 
-@cache.memoize(timeout=5*60)
 def render_generate_repo_file(copr_dir, name_release, arch=None):
+    name = helpers.get_stat_name(
+        CounterStatType.REPO_DL,
+        copr_dir=copr_dir,
+        name_release=name_release,
+    )
+    CounterStatLogic.incr(name=name, counter_type=CounterStatType.REPO_DL)
+    return render_generate_repo_file_cached(copr_dir, name_release, arch=arch)
```
praiskup, line 886: The `commit()` call should be moved as well.

praiskup:
Otherwise LGTM

frontend/coprs_frontend/coprs/views/coprs_ns/coprs_general.py
xsuchy: If you merge https://github.com/fedora-copr/copr/pull/3074 then incr() will do the commit and it is not necessary to add it here.

xsuchy: Updated

praiskup:
(ignoring the pylint errors, they re-appeared because the method was renamed)

PR #3084, opened 2024-01-03
Fix copr-rpmbuild --dump-configs

evan-goode:
As of 302a78f, `MockBuilder.prepare_configs()` does not take a `configdir` argument, nor does it return a list of written config files.

### DIFF
--- a/rpmbuild/main.py
+++ b/rpmbuild/main.py
@@ -301,10 +301,8 @@ def dump_configs(args, config):
     resultdir = config.get("main", "resultdir")
     builder = MockBuilder(task, None, resultdir, config)
 
-    configdir = os.path.join(resultdir, "configs")
-    config_paths = builder.prepare_configs(configdir)
-    for config_path in config_paths:
-        log.info("Wrote: "+config_path)
+    builder.prepare_configs()
+    log.info("Wrote: %s", builder.mock_config_file)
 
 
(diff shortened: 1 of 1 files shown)


rpmbuild/main.py
```diff
@@ -301,10 +301,8 @@ def dump_configs(args, config):
     resultdir = config.get("main", "resultdir")
     builder = MockBuilder(task, None, resultdir, config)
 
-    configdir = os.path.join(resultdir, "configs")
-    config_paths = builder.prepare_configs(configdir)
-    for config_path in config_paths:
-        log.info("Wrote: "+config_path)
+    builder.prepare_configs()
+    log.info("Wrote: "+builder.mock_config_file)
```
FrostyX, line 305: I know we had the mistake already in our code but since you are touching it, could you please change 

```
log.info("Wrote: "+builder.mock_config_file)
```

to 

```
log.info("Wrote: %s", builder.mock_config_file)
```

That's how logging parameters should be passed.

FrostyX (commented):
Hello @evan-goode,
thank you very much for the PR.

It works perfectly, I have only one small note.

rpmbuild/main.py
evan-goode: Sure thing!

FrostyX:
Thank you :-)

PR #3085, opened 2024-01-04
frontend, python: allow ignoring errors that a project already exists

### LINKED ISSUE #2140: Return the existing project trying to add a project which already exists
Original issue: https://pagure.io/copr/copr/issue/2140
Opened: 2022-03-25 09:58:15
Opened by: [csomh](https://accounts.fedoraproject.org/user/csomh)

(This came up while working with the Python API, but I'm not sure if the solution is going to impact only that one.)

In Packit we hit a race condition, when trying to create the same project from tasks running in parallel. [The solution](https://github.com/packit/packit/pull/1527/files) was to check the exception and if it was raised b/c the project already exists, ignore the error. But this is ugly.

Looking at the bigger picture in that code, we also have `create_copr_project_if_not_exists()`, which first checks if the project already exists, then tries to create it. Which is 2 API calls.

The ask here is to have a new endpoint to `project_proxy.add()` to not to raise an error, if a project with an identical name and configuration (this is important, if projects are configured differently, the caller should find out about this) already exists, OR

to have a new API which either creates the project if it doesn't exist, or returns the existing project, if it does.

FrostyX:
Fix #2140

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
@@ -26,7 +26,7 @@ from coprs.views.apiv3_ns.schema.schemas import (
     project_delete_input_model,
     fullname_params,
     pagination_project_model,
-    ownername_params,
+    project_params,
     pagination_params,
 )
 from coprs.views.apiv3_ns.schema.docs import query_docs
@@ -139,7 +139,7 @@ class Project(Resource):
 class ProjectList(Resource):
     @restx_pagination
     @query_to_parameters
-    @apiv3_projects_ns.doc(params=ownername_params | pagination_params)
+    @apiv3_projects_ns.doc(params=project_params | pagination_params)
     @apiv3_projects_ns.marshal_list_with(pagination_project_model)
     @apiv3_projects_ns.response(
         HTTPStatus.PARTIAL_CONTENT.value, HTTPStatus.PARTIAL_CONTENT.description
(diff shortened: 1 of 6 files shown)


FrostyX:
This was much more complicated than I expected.

- @nikromen Am I doing the query parameters parsing correctly? It was done with the [RequestParser](https://flask-restx.readthedocs.io/en/latest/parsing.html) but that changed in PR #2807
- I am passing the `exist_ok` value all the way into the `CoprUniqueNameValidator` and then hackily trying to access its copr instance from the outside. It works but we may want to do it differently? Maybe not using the validator and instead of `CoprsLogic.add` using `CoprsLogic.get_or_create` (we have a similar method for `CoprDirsLogic`)?

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
```diff
@@ -181,28 +181,35 @@ def get(self, query, **kwargs):
         return {"items": projects, "meta": paginator.meta}
 
 
-@apiv3_projects_ns.route("/add/<ownername>")
+@apiv3_projects_ns.route("/add/<ownername>/")
```
praiskup, line 184: Is this a breaking API change?  Note that some clients do not necessarily have to follow redirects.

FrostyX: Sorry, this is a typo. I will revert this back.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
```diff
@@ -181,28 +181,35 @@ def get(self, query, **kwargs):
...
+        form_class = forms.CoprFormFactory.create_form_cls(user=user, group=group,
+                                                           exist_ok=exist_ok)
         set_defaults(data, form_class)
         form = form_class(data, meta={"csrf": False})
 
         if not form.validate_on_submit():
+            if exist_ok:
+                for validator in form.name.validators:
+                    if isinstance(validator, forms.CoprUniqueNameValidator):
+                        return to_dict(validator.copr)
```
praiskup, line 212: This deserves some in-line comment :-)

praiskup:
otherwise lgtm

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
FrostyX: Agreed. And thank you for poking it, there was a small bug. I fixed it, and added a comment.

FrostyX:
Updated, and I also added a unit test.

praiskup (approved):
LGTM, thank you!

PR #3095, opened 2024-01-10
docker: increment fedora images to 39

### LINKED ISSUE #2945: Switch docker-compose to F39
To test our code works on F39.

nikromen:
We have centos7 images for postgres and nginx which are fairly old images - should we e.g. build our own in quay with rocky linux 9?

Fixes #2945

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:37
+FROM registry.fedoraproject.org/fedora:39
 MAINTAINER copr-devel@lists.fedorahosted.org
 
 ARG ADDITIONAL_COPR_REPOSITORIES="@copr/copr-dev"
@@ -46,8 +46,7 @@ RUN setcap cap_net_raw,cap_net_admin+p /usr/bin/ping
 RUN ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -q
 
 # setup root user
-RUN echo 'root:passwd' | chpasswd && \
-    mkdir /root/.ssh && chmod 700 /root /root/.ssh
+RUN echo 'root:passwd' | chpasswd && chmod 700 /root /root/.ssh
 
 # setup copr user
 RUN set -x ; \
--- a/build_aux/Containerfile.unittest
+++ b/build_aux/Containerfile.unittest
@@ -1,4 +1,4 @@
-FROM fedora:37
+FROM fedora:39
 
 RUN dnf update -y && dnf install python-pytest-cov copr-dist-git rpmdevtools python3-copr copr-cli dnf-plugins-core -y
 
--- a/docker/builder/Dockerfile
+++ b/docker/builder/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:37
+FROM registry.fedoraproject.org/fedora:39
 MAINTAINER copr-devel@lists.fedorahosted.org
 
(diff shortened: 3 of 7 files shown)


FrostyX (approved):
> should we e.g. build our own in quay with rocky linux 9?

Personally, I wouldn't bother. It is only for development, so I don't care how old and insecure it is. But if we decide to run Copr in OpenShift someday, we probably should upgrade those.

praiskup:
> We have centos7 images for postgres and nginx which are fairly old images

No.  That's what [sclorg](https://github.com/sclorg) folks do for us.

docker/backend/Dockerfile
```diff
@@ -47,7 +47,7 @@ RUN ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -q
 
 # setup root user
 RUN echo 'root:passwd' | chpasswd && \
-    mkdir /root/.ssh && chmod 700 /root /root/.ssh
+    mkdir -p /root/.ssh && chmod 700 /root /root/.ssh
```
praiskup, line 50: This change shouldn't be needed.

praiskup:
Otherwise lgtm.

docker/backend/Dockerfile
nikromen: I needed it to make docker-compose work. It seemed to me that f39 has /root/.ssh dir by default. I'll delete images and I'll try it again if it is really needed

nikromen: I got 

```
 => ERROR [ 5/14] RUN echo 'root:passwd' | chpasswd &&     mkdir /root/.ssh && chmod 700 /root /root/.ssh                                                                                                       0.2s 
------                                                                                                                                                                                                               
 > [ 5/14] RUN echo 'root:passwd' | chpasswd &&     mkdir /root/.ssh && chmod 700 /root /root/.ssh:                                                                                                                  
0.132 mkdir: cannot create directory '/root/.ssh': File exists                                                                                                                                                       
------
Dockerfile:49
--------------------
  48 |     # setup root user
  49 | >>> RUN echo 'root:passwd' | chpasswd && \
  50 | >>>     mkdir /root/.ssh && chmod 700 /root /root/.ssh
  51 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c echo 'root:passwd' | chpasswd &&     mkdir /root/.ssh && chmod 700 /root /root/.ssh" did not complete successfully: exit code: 1
ERROR: Service 'backend-log' failed to build : Build failed
```

praiskup: I see ... I'd probably drop the `mkdir` then, if it is not needed?

This really seems like you use the `-p` becuase `/root` doesn't exist.

praiskup: IOW, `ssh-keygen` indeed creates `~/.ssh` (since ever?).  It is surprising this ever worked :-)

praiskup: I tested `ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -q` in a fresh f39 container, and `/root/.ssh` was not created.  openssh-clients-9.3p1-10.fc39.x86_64 weird.

nikromen: Ok I have no idea why it was created but I removed the mkdir since it works without it. Do you want to leave as it is without the mkdir or investigate what created it?

praiskup: I think it is created by installing openssh-server.  I think it is fine as-is (we should long-term avoid doing these permissions tweaks, I think, but it's for a separate fix).

praiskup (approved):

PR #3099, opened 2024-01-11
Messaging: Set the agent_name, usernames and groups properties

abompard:
This sets a few properties that are standard in Fedora Messaging schemas and will allow notifications to be emitted by FMN.

### DIFF
--- a/messaging/copr_messaging/private/hierarchy.py
+++ b/messaging/copr_messaging/private/hierarchy.py
@@ -159,6 +159,19 @@ class _BuildMessage(_CoprProjectMessage):
             release=release,
         )
 
+    @property
+    def usernames(self):
+        """
+        List of usernames that are affected by the action that caused this message.
+
+        Those users will be notified by FMN if they chose to receive their notifications.
+        """
+        usernames = [self.submitter, self.body.get("owner")]
+        # filter out nones and groups and deduplicate
+        usernames = set(u for u in usernames if isinstance(u, str) and not u.startswith("@"))
+        # return it sorted
+        return list(sorted(usernames))
+
 
 class _BuildChrootMessage(_BuildMessage):
     @property
--- a/messaging/copr_messaging/private/hierarchy.py
+++ b/messaging/copr_messaging/private/hierarchy.py
@@ -160,6 +160,15 @@ class _BuildMessage(_CoprProjectMessage):
         )
 
     @property
+    def submitter(self):
+        """The username of the person who submitted the build."""
+        username = self._raw_submitter
+        if isinstance(username, str) and ":" in username:
+            # This is namespaced, such as github.com:foobar
+            return None
+        return username
+
+    @property
     def usernames(self):
         """
         List of usernames that are affected by the action that caused this message.
--- a/messaging/copr_messaging/schema.py
+++ b/messaging/copr_messaging/schema.py
@@ -46,6 +46,15 @@ class BuildChrootEnded(_BuildChrootMessage):
             self.status,
         )
 
+    @property
+    def agent_name(self):
+        """The username who caused the action that generated this message."""
+        # The canceled status can be caused by the submitter but also any other
+        # that has build or admin access. To be safe we don't set it as the
+        # agent_name.
+        # The other statuses are not caused by the user who started the build
+        return None
+
 
 class BuildChrootStarted(_BuildChrootMessage):
     """
@@ -57,6 +66,11 @@ class BuildChrootStarted(_BuildChrootMessage):
             super(BuildChrootStarted, self)._str_prefix(),
             self.chroot,
         )
+
+    @property
+    def agent_name(self):
+        """The username who caused the action that generated this message."""
+        return self.submitter
 
 
 class BuildChrootStartedV1(_PreFMBuildMessage, BuildChrootStarted):
--- a/messaging/copr_messaging/private/schema_old.py
+++ b/messaging/copr_messaging/private/schema_old.py
@@ -129,8 +129,8 @@ class _PreFMBuildMessage(_BuildChrootMessage):
         return self.body.get('pkg')
 
     @property
-    def submitter(self):
-        """The username of the person who submitted the build."""
+    def _raw_submitter(self):
+        """The username of the person who submitted the build. May be namespaced."""
         return self.body.get('user')
 
     def _evr(self):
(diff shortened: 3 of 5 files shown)


messaging/copr_messaging/private/hierarchy.py
```diff
@@ -159,6 +159,21 @@ def package_full_name(self):
...
+        # filter out nones and groups and deduplicate
+        usernames = set(u for u in usernames if isinstance(u, str) and not u.startswith("@"))
+        # return it sorted
+        return list(sorted(usernames))
+
+    @property
+    def groups(self):
+        owner = self.body.get("owner")
+        if isinstance(owner, str) and owner.startswith("@"):
+            return [owner[1:]]
```
praiskup, line 174: Just note that the "group"  means Copr group, not FAS group.

abompard: Oh OK, I thought that copr groups were mapped from FAS groups. Good to know, I'll remove it.

praiskup: They are mapped 1:1, but the names may differ.

messaging/copr_messaging/schema.py
```diff
@@ -46,6 +46,14 @@ def __str__(self):
             self.status,
         )
 
+    @property
+    def agent_name(self):
+        """The username who caused the action that generated this message."""
+        if self.status == "canceled":
+            return self.submitter
+        # The other statuses are not caused by the user who started the build
```
praiskup, line 54: Strictly speaking, anyone who has build or admin access to the project should be able to cancel the build.  I don't think we have the needed info here.

messaging/copr_messaging/schema.py
```diff
@@ -58,6 +65,10 @@
             self.chroot,
         )
 
+    @property
+    def agent_name(self):
```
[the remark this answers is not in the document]
praiskup: The linter probably assigned this warning to a wrong position in code (more methods with the same name).

praiskup:
Thank you for working on this, @abompard!

I'm not sure how precise we have to be here.  E.g. what's the semantics of the `username` or `usernames`, who is going to react on this?  The field will often be None because the submitter is unknown.

messaging/copr_messaging/schema.py
abompard: Indeed, would it be possible to add that info to the message body?

praiskup: I think so, but this seems like a separate issue (the asynchronous cancelation on the remote machine is relatively complicated logic).

praiskup: So, does it make sense to mix up the submitter with the person who canceled the build?  If yes, I think this PR is good to go. Otherwise, we should do something about it first, ideas?

abompard: Fixed! Thanks.

abompard: To be safe I didn't assume the submitter was the one who canceled the build. This may create notifications for people who didn't want to get notified of their own actions, but that's a lesser evil I think.

praiskup:
@knurd I don't think the `submitter` field is changed by this patch, so there is no change.

But you are right I think. The problem is the agent_name, which is not doing any translation, and doing it probably isn't easy. Since `agent_name == submitter` now, this could cause issues I bet.  WDYT @abompard ?

abompard:
Ah yes, I did not know that the submitter would be a namespaced username.
I could set it only when it's not namespaced, but the message schema shouldn't make the FASJSON call to translate the namespaced usernames to FAS usernames, that should be done by the application itself. Do you think Copr could do that @praiskup ? The call would be something like `curl --negotiate -u : https://fasjson.fedoraproject.org/v1/search/users/?github_username=abompard` for my Github account, for example.

praiskup:
Whohoo! This fassjson thingy is awesome, seems like usable for #2977!

> I could set it only when it's not namespaced,

If not namespaced, then it should already be a FAS account, sounds like a good idea to me.

> but the message schema shouldn't make the FASJSON call to translate the namespaced usernames to FAS usernames

The `curl` call takes several tens of seconds (~45s). At some point, copr should do this translation call I think - but things need to be cached somehow, I don't know.  This is again a topic for separate RFE issue I'm afraid.

praiskup:
> If not namespaced, then it should already be a FAS account, sounds like a good idea to me.

This is of course truth only for the Fedora Copr deployment, not other Copr deployments.

abompard:
> The `curl` call takes several tens of seconds (~45s). At some point, copr should do this translation call I think - but things need to be cached somehow, I don't know. This is again a topic for separate RFE issue I'm afraid.

It really shouldn't be so long, I need to check that. It's supposed to be an indexed LDAP call, so it should be very fast. But I'd appreciate if it were cached nonetheless ;-)

Please note that I just realized that [gitlab usernames can't be queried yet](https://github.com/fedora-infra/fasjson/issues/645). It  should be very easy to add though, only a couple lines.

knurd:
Thx to both of you for looking into/working on this, much appreciated. 👏

While at it a comment from the cheap seats that likely is unnecessary at this point: if looking up the github/gitlab names is to expensive/complicatd, I wonder if all of that could be avoided by sending the mails to the copr user that set up the "project package" in copr that acts when the webhook fires.

abompard:
> While at it a comment from the cheap seats that likely is unnecessary at this point: if looking up the github/gitlab names is to expensive/complicatd, I wonder if all of that could be avoided by sending the mails to the copr user that set up the "project package" in copr that acts when the webhook fires.

Is this user available somewhere in the data dict that is used for the body of the message? If so, then yeah that'd be the best solution IMHO.

abompard:
OK I've added a commit that should teach the schemas to ignore namespaced usernames in `submitter` and thus `agent_name` and `usernames`.

praiskup:
> Is this user available somewhere in the data dict that is used for the body of the message?

These are all good requests :-) like the other in this PR before;  we need separate issue I bet, or even PR.
We don't track who creates the project in particular copr.  The token is either per-user (typically not used by automation) or for webhooks per-project (and there's no per-user ownership of those).  Perhaps we speak
about #1435 here.

praiskup:
The PR now LGTM, thank you @abompard

knurd:
> Ah yes, I did not know that the submitter would be a namespaced username. 

That aspect was not fixed, wasn't it? If so: is it on some todo list, or do we need a ticket to ensure that does not fall through the cracks?

knurd:
Ohh, and btw, I still don't get notifications from copr for non-namespaces things I did. Is the new code maybe not yet deployed?

praiskup:
If you want to track that, submit an issue, yes.  And the release/deployment is probably going to happen on Thursday 8th of March.

PR #3106, opened 2024-01-16
cli: dont check dirname presence in action_build

### LINKED ISSUE #2786: Can't upload-SRPM into custom copr directories (Error: Dirname doesn't exist)
`copr-cli build --help` and manual says I can specify the repo in the form `project:<tag>`, but it doesn't work. I tried with a few different repos under my user.

Error:

```
$ copr-cli build playground:foo bar.src.rpm

Something went wrong:
Error: Dirname 'playground:foo' doesn't exist in 'jn64/playground' copr

$ copr-cli --version
copr version 1.109
```

Quoting the manual:

> Finally, instead of just project name, you can also pass project:tag. In that
> case, the build will land into the project side repository instead of the
> main repository. The side repository of the name project:tag will be
> automatically created if it doesn’t exist yet. You can e.g. use this feature
> to launch test builds while the main user-facing repository stays intact. You
> can specify any tag that is appropriate for the given build(s).

nikromen:
Copr build <projectname:dirname> should according to documentation create the dirname if does not exists, but the early check check_before_build in action_build actually runs /build/check-before-build endpoint even before the <projectname:dirname> has a chance to be created, thus complains about non-existent dirname.
Passing only projectname to the check should be enough for the check as is since the /build/check-before-build is called once more after dirname creation.

Fixes #2786

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -325,7 +325,14 @@ class Commands(object):
         # Before we start uploading potentially large source RPM file, make sure
         # that the user has valid credentials and can build in the project.
         self.client.build_proxy.check_before_build(
-            username, projectname, project_dirname, buildopts)
+            ownername=username,
+            projectname=projectname,
+            # documentation says user can create directory with copr build, but
+            # /build/check-before-build endpoint checks for presence of this
+            # dirname even before creating it in this check.
+            project_dirname=None,
+            buildopts=buildopts,
+        )
 
         builds = []
         for pkg in args.pkgs:
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
@@ -228,7 +228,7 @@ def check_before_build():
     copr = get_copr()
 
     # Raises an exception if CoprDir doesn't exist
-    if data["project_dirname"]:
+    if data.get("project_dirname"):
         CoprDirsLogic.get_by_copr(copr, data["project_dirname"])
 
     # Permissions check


cli/copr_cli/main.py
```diff
@@ -325,7 +325,15 @@ def action_build(self, args):
...
         # that the user has valid credentials and can build in the project.
         self.client.build_proxy.check_before_build(
-            username, projectname, project_dirname, buildopts)
+            ownername=username,
+            projectname=projectname,
+            # documentation says user can create directory with copr build, but
+            # /build/check-before-build endpoint checks for presence of this
+            # dirname even before creating it in this check. Passing only projectname
+            # should be sufficient for the comment above.
+            project_dirname=projectname,
```
praiskup, line 334: Have you tried to call this with `project_dirname=None`?

praiskup:
Otherwise looks good to me!

cli/copr_cli/main.py
nikromen: Yes and it failed with `Error: The browser (or proxy) sent a request that this server could not understand`. If `project_dirname` is not passed to cli and only projectname is passed in either `ownername/projectname` or only `projectname` format then the project_dirname is not `None` but `projectname`.

praiskup: This is a frontend bug, isn't it?

praiskup: Note this:
https://github.com/fedora-copr/copr/blob/3b68ad13a4f44229ecb02ef5d9677278a7a69c2e/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py#L230-L232

python/copr/v3/proxies/build.py
```diff
@@ -162,6 +162,9 @@ def check_before_build(self, ownername, projectname,
         :param buildopts: http://python-copr.readthedocs.io/en/latest/client_v3/build_options.html
         :return: Munch
         """
+        if project_dirname is None:
+            project_dirname = projectname
```
praiskup, line 166: I don't want to nitpick, but what happens if `project_dirname=None` goes down to the Frontend API code?  It seems we are hiding a Frontend API bug here ...

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py
```diff
@@ -228,7 +228,7 @@ def check_before_build():
     copr = get_copr()
 
     # Raises an exception if CoprDir doesn't exist
-    if data["project_dirname"]:
+    if data.get("project_dirname"):
```
praiskup, line 231: :+1:

PR #3113, opened 2024-01-22
backend: periodically check if the resalloc ticket isn't failed

### LINKED ISSUE #2888: More reliable SSH connection to builders
There's an interesting situation that has been bothering me for quite some time, and I finally got to a debugging.  Here we go:

1. resalloc-server is controlling hypervisors over virsh+ssh, and VMs started there over SSH, too
2. resalloc-server successfully allocates a new VM on our hypervisor
3. resalloc ticket taken by Copr worker daemon is assigned to that machine
4. some disk failure occurs on the hypervisor, causing every attempt to read/write something on '/' ends up in an uninterruptable process ('D' in ps aux)
5. this affects VMs as well
6. any new attempt to establish a new SSH connection to the hypervisor or VM on it fails
7. resalloc is able to detect this situation, and request the VM removal (but not successfully, of course, no new SSH connection can be established to hypervisor to terminate the machine)
8. Copr doesn't recognize this situation;  we have opened an SSH connection to the worker to run `copr-rpmbuild-log` process, but that is `D` process hanging forever, and the already opened SSH connection stays forever (it doesn't touch disk, so the remote sshd simply keeps posting keepalive packets to keep the SSH connection alive)
9. Since the SSH connection persists, we wait indefinitely keeping the task in a "running" state
10. This likely doesn't timeout (I haven't checked), because the timeout is implemented remotely over `copr-rpmbuild`.

FrostyX:
Fix #2888

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -12,6 +12,7 @@ import time
 import json
 
 from packaging import version
+from cachetools.func import ttl_cache
 
 from copr_common.enums import StatusEnum
 
@@ -381,7 +382,26 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         self.log.debug("setting title: %s", text)
         self.setproctitle(text)
 
+    def _check_build_interrupted(self):
+        """
+        Should we interrupet a running worker?
+        """
+        return (self._check_failed_resalloc_ticket()
+                or self._cancel_task_check_request())
+
+    @ttl_cache(ttl=10*60)
+    def _check_failed_resalloc_ticket(self):
+        """
+        Did the resalloc ticket fail?
+        """
+        self.host.ticket.collect()
+        self.log.info("Status: %s", self.host.ticket.failed)
+        return self.host.ticket.failed
+
     def _cancel_task_check_request(self):
+        """
+        Was the build canceled by the user?
+        """
         self.canceled = bool(self.redis_get_worker_flag("cancel_request"))
         return self.canceled
 
@@ -436,12 +456,15 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
             log=self.log,
         )
 
-    def _cancel_running_worker(self):
+    def _discard_running_worker(self):
         """
         This is "canceling" callback to CancellableThreadTask, so please never
         raise any exception.  The worst case scenario is that nothing is
         canceled.
         """
+        if not self.canceled:
+            return
+
         self._proctitle("Canceling running task...")
         self.redis_set_worker_flag("canceling", 1)
         try:
@@ -757,12 +780,14 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         self._start_remote_build()
         transfer_failure = CancellableThreadTask(
             self._transfer_log_file,
-            self._cancel_task_check_request,
-            self._cancel_running_worker,
+            self._check_build_interrupted,
+            self._discard_running_worker,
             check_period=CANCEL_CHECK_PERIOD,
         ).run()
         if self.canceled:
             raise BuildCanceled
+        if self.host.ticket.failed:
+            transfer_failure = "Resalloc ticket FAILED"
         if transfer_failure:
             raise BuildRetry("SSH problems when downloading live log: {}"
                              .format(transfer_failure))
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -57,6 +57,7 @@ BuildRequires: python3-setproctitle
 BuildRequires: python3-sphinx
 BuildRequires: python3-tabulate
 BuildRequires: python3-zstandard
+BuildRequires: python3-cachetools
 BuildRequires: modulemd-tools >= 0.6
 BuildRequires: prunerepo >= %prunerepo_version
 BuildRequires: dnf
@@ -95,6 +96,7 @@ Requires:   python3-retask
 Requires:   python3-setproctitle
 Requires:   python3-tabulate
 Requires:   python3-boto3
+Requires:   python3-cachetools
 Requires:   redis
 Requires:   rpm-sign
 Requires:   rsync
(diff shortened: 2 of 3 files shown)


backend/copr_backend/background_worker_build.py
```diff
@@ -381,6 +381,18 @@ def _proctitle(self, text):
...
+    def _check_build_interrupted(self):
+        """
+        Should we interrupet a running worker?
+        """
+        # Was the build canceled by the user?
+        if self._cancel_task_check_request():
+            return True
+
+        # Is the resalloc ticket healthy?
+        self.host.ticket.collect()
```
praiskup, line 393: The cancel check period is 5 seconds .... I'd propose doing this at most every minute, maybe 5 minutes.
Just to not cause unnecessary `backend <-> resalloc` traffic.

FrostyX:
I fixed the stuck unit tests and increased the check interval. It wasn't just a number increase because we wanted the interval for checking cancel actions to remain as is.

It behaves a little bit weird and I am not sure if it is because of testing locally or if it requires some polishing. When I do 

```
sqlite> UPDATE tickets SET state='FAILED' WHERE id=28;
sqlite> select * from tickets;
28|6|FAILED||frostyx/testing-failed-tickets--frostyx
```

Then the build gets triggered again (it takes longer than I would expect) and the status of the failed ticket changes

```
sqlite> select * from tickets;
28|6|CLOSED||frostyx/testing-failed-tickets--frostyx
```

backend/copr_backend/background_worker_build.py
```diff
@@ -381,7 +382,25 @@ def _proctitle(self, text):
...
         self.setproctitle(text)
 
+    def _check_build_interrupted(self):
+        """
+        Should we interrupet a running worker?
+        """
+        return (self._cancel_task_check_request()
+                or self._check_failed_resalloc_ticket())
+
+    @ttl_cache(ttl=60)
```
praiskup, line 392: Don't we want to use 5 minutes or something like that?  This is catching rather a rare case, and it's acceptable that a few users wait a bit longer for the resubmit...?

Note we run even >= 200 machines in parallel, ttl=60 means >= 3 requests/s to the resalloc server.

backend/copr_backend/background_worker_build.py
praiskup: The `collect()` may hang when the server goes down because we are using `survive_server_restart`: https://github.com/praiskup/resalloc/blob/7a46fcc269a9e38174786ca3fb365e96c0a46cbd/resalloc/client.py#L43-L53

Might be OK, ... but then, another potential problem is that not all the exceptions are being caught.

backend/copr_backend/background_worker_build.py
```diff
@@ -381,7 +382,25 @@ def _proctitle(self, text):
         self.log.debug("setting title: %s", text)
         self.setproctitle(text)
 
+    def _check_build_interrupted(self):
+        """
+        Should we interrupet a running worker?
+        """
+        return (self._cancel_task_check_request()
+                or self._check_failed_resalloc_ticket())
```
praiskup, line 390: Maybe a nitpick; if the ticket is FAILED, there's no point in trying to cancel the remote process ... so I'd at least reverse these checks.

praiskup: It is somewhat difficult to verify this behaves as expected :-) because the return value is cached.

backend/copr_backend/background_worker_build.py
FrostyX: I will kill resalloc server and see what happens.

backend/copr_backend/background_worker_build.py
```diff
@@ -436,6 +455,15 @@ def _alloc_ssh_connection(self):
...
         )
 
+    def _discard_running_worker(self):
+        """
+        This is "canceling" callback to CancellableThreadTask, so please never
+        raise any exception.  The worst case scenario is that nothing is
+        canceled.
+        """
+        if self.canceled:
+            self._cancel_running_worker()
```
praiskup, line 465: small nit, we could make the patch smaller by moving this `self.canceled` condition into the `_cancel_running_worker()` task (perhaps renaming the method, and keeping the comment).

FrostyX:
Updated, PTAL :-)

backend/copr_backend/background_worker_build.py
```diff
@@ -381,7 +383,30 @@ def _proctitle(self, text):
...
+    def _check_failed_resalloc_ticket(self):
+        """
+        Did the resalloc ticket fail?
+        """
+        try:
+            # Resalloc server can be temporarily down
+            self.host.ticket.collect()
+        except socket.error:
+            self.log.error("Connection to resalloc failed, "
+                           "using cached ticket status")
```
praiskup, line 403: I'm still not sure this is right.  Is this a bug in the resalloc client?  since we use the survive_server_restart option, it should [never return this exception ](https://github.com/praiskup/resalloc/blob/7a46fcc269a9e38174786ca3fb365e96c0a46cbd/resalloc/client.py#L46-L48)

FrostyX: It is possible that is bug in the resalloc client.

I submitted a build, waited until the `fedora-39-x86_64` chroot is in the `running` state. Then I did `docker-compose kill resalloc`, and then `socket.gaierror` exception appeared in the logs.

Adding this try-except block helped, and I started seeing the "Connection to resalloc failed" message instead.

If you want, we can drop the try-except block here, and fix it in resalloc.

praiskup: That's suspicious,
```
In [1]: import socket

In [2]: isinstance(socket.gaierror(), socket.error)
Out[2]: True
```
Are we eventually using the "survives_server_restart=True"?

praiskup: What do you think about https://github.com/praiskup/resalloc/pull/136 ?

FrostyX: The resalloc PR helped the issue, so I dropped the try-except here.
Now it seems that `self.host.ticket.collect()` runs forever until the resalloc servers gets back up again. No exceptions are appearing.

praiskup:
LGTM

PR #3120, opened 2024-01-28
frontend: add runtime dependency on flask-sesison and markupsafe

FrostyX:
Since 3f1413a, we use flask-session and markupsafe, the commit added a BuildRequires for them but forgot to add Requires.


praiskup:
lgtm

Issue #3118, opened 2024-01-26
The ssh run_expensive() method should have a _safe() variant


praiskup:
The command-over-ssh may produce a large output, feeding backend's memory
and DoS it.  With #1366 merged, we should make sure the output is somewhat
safe to read into memory (e.g. piping through /usr/bin/head *on-host*, as the
remote utility is not trusted).

PR #3124, opened 2024-02-02
backend: limit stdout/stderr of ssh.run_expensive() commands

praiskup:
Fixes: #3118


FrostyX:
Thank you for the PR. It looks good to me but the code is too low-level to evaluate in my head if it works as it should. I'd vote for merging and seeing if we observe any regressions on devel.

praiskup:
There still were at least two bugs; (a) process still running but stdout/stderr closed (e.g. `exec >&-`), and (b) undefined `stdout` variable.

> It looks good to me but the code is too low-level to evaluate in my head if it works as it should. I'd vote for merging and seeing if we observe any regressions on devel.

There are unit tests with almost 100% coverage, I deployed this on copr-backend and it seems working. :crossed_fingers: +1 for merging.

common/copr_common/subprocess_live_output.py
```diff
@@ -0,0 +1,169 @@
...
+            readable, _, exc = select.select([fd], [], [fd], poll)
+            if exc:
+                queue.put((identifier, ERR))
+                break
+
+            if not readable:
+                queue.put((identifier, LOOP))
+                continue
+
+            chunk = pipe.read1(1024)
```
nikromen, line 44: is this typo? I haven't find anything about `read1`, does it even exist?

backend/copr-backend.spec
```diff
@@ -6,7 +6,7 @@
 %global tests_version 5
 %global tests_tar test-data-copr-backend
 
-%global copr_common_version 0.20.1.dev1
+%global copr_common_version 0.21.1.dev1
```
nikromen, line 9: I don't understand changing the versions in this PR so I'll skip this :D

nikromen (commented):
otherwise lgtm

common/copr_common/subprocess_live_output.py
praiskup: Yes, read() waits till the buffer is filled, or EOF. We need to process partial/incomplete reads.

praiskup:
I added the comment. PTAL

nikromen (approved):
thanks for adding the comment!

backend/copr-backend.spec
praiskup: Sorry I missed this comment; for the development purposes, we need an NVR >= the_latest_released NVR.  Therefore the bump to 0.21.1 is needed.  I think that 0.21.dev1 woudl be enough, though.

PR #3144, opened 2024-02-16
rpmbuild: don't set bootstrap_image_ready for rawhide

### LINKED ISSUE #3140: Custom method is broken because of DNF vs DNF5
We discovered this with @praiskup while upgrading builders to F39 (but it is unrelated)

Custom method builds fail with:

```
INFO: Package manager dnf5 detected and used (direct choice)
Start: installing minimal buildroot with dnf5
execv(/usr/bin/dnf5) failed: No such file or directory
WARNING: DNF5 command failed, retrying, attempt #2, sleeping 10s
execv(/usr/bin/dnf5) failed: No such file or directory
WARNING: DNF5 command failed, retrying, attempt #3, sleeping 10s
execv(/usr/bin/dnf5) failed: No such file or directory
```

Relevant:

```
INFO: Not updating bootstrap chroot, bootstrap_image_ready=True
```

which is set in `rpmbuild/mock-custom-build.cfg.j2`.

Full error:

[builder-live.log.gz.txt](https://github.com/fedora-copr/copr/files/14238750/builder-live.log.gz.txt)

This is probably because Mock says that rawhide is a DNF5 distribution but the container doesn't have DNF5. The proposed solution is to temporarily replace the `mock-custom-build.cfg.j2` in ansible and set `bootstrap_image_ready=False` (for rawhide)

FrostyX:
Fix #3140

### DIFF
--- a/rpmbuild/mock-custom-build.cfg.j2
+++ b/rpmbuild/mock-custom-build.cfg.j2
@@ -2,9 +2,12 @@ include('/etc/mock/{{ chroot }}.cfg')
 config_opts['rpmbuild_networking'] = True
 config_opts['use_host_resolv'] = True
 
+{# See https://github.com/fedora-copr/copr/issues/3140 #}
+{%- if not chroot.startswith("fedora-40-") %}
 # We don't do 'dnf builddep' for the custom method, so we don't need to install
 # python3-dnf-plugins core or dnf5-plugins at all.
 config_opts["bootstrap_image_ready"] = True
+{% endif %}
 
 # Don't mixup caches with the normal mock build.
 config_opts["root"] = "copr-custom-" + config_opts["root"]


rpmbuild/mock-custom-build.cfg.j2
```diff
@@ -2,9 +2,12 @@ include('/etc/mock/{{ chroot }}.cfg')
 config_opts['rpmbuild_networking'] = True
 config_opts['use_host_resolv'] = True
 
+{# See https://github.com/fedora-copr/copr/issues/3140 #}
+{%- if not chroot.startswith("fedora-rawhide-") %}
```
praiskup, line 6: We really need `"fedora-40" in chroot` condition here.  Because of https://github.com/rpm-software-management/mock/pull/1332

rpmbuild/mock-custom-build.cfg.j2
```diff
@@ -2,9 +2,12 @@ include('/etc/mock/{{ chroot }}.cfg')
 config_opts['rpmbuild_networking'] = True
 config_opts['use_host_resolv'] = True
 
+{# See https://github.com/fedora-copr/copr/issues/3140 #}
+{%- if not chroot.startswith(("fedora-rawhide-", "fedora-40-")) %}
```
FrostyX, line 6: Only F40

FrostyX:
Updated

praiskup (approved):
LGTM, we need to keep the downstream patch till DNF5 appears in the fedora:41 container, though.

PR #3155, opened 2024-02-22
Drop backend-specific ActionType and use ActionTypeEnum

praiskup:


backend/copr_backend/actions.py
```diff
@@ -56,18 +56,18 @@ def create_from(cls, opts, action, log=None):
...
-            ActionType.LEGAL_FLAG: LegalFlag,
-            ActionType.CREATEREPO: Createrepo,
-            ActionType.UPDATE_COMPS: CompsUpdate,
-            ActionType.GEN_GPG_KEY: GenerateGpgKey,
-            ActionType.RAWHIDE_TO_RELEASE: RawhideToRelease,
-            ActionType.FORK: Fork,
-            ActionType.BUILD_MODULE: BuildModule,
-            ActionType.DELETE: Delete,
-            ActionType.REMOVE_DIRS: RemoveDirs,
+            ActionTypeEnum("legal-flag"): LegalFlag,
```
FrostyX, line 59: Probably a typo, the rest of the action names use underscores instead.

FrostyX (commented):
I like the change, thank you.

We should probably (Build)Require the appropriate copr-common version

>Drop backend-specific ActionType and use ActionTypeEnum

Nonstandard commit name :-)

backend/copr_backend/actions.py
praiskup: Seems like a typo that I didn't want to change, that would be a common-API change, not sure if worth it.

praiskup:
> We should probably (Build)Require the appropriate copr-common version

This seems unnecessary, this common-api is sooo old (and %copr_common_version is likely new enough, I didn't want to do the research).

> Nonstandard commit name :-)

Fixed.

xsuchy:
+1

PR #3159, opened 2024-02-24
Actually assign fallback value from self.username

kanru:
Several places that intend to use self.username did not actually assign the return value. This fixes `copr list` without  username argument.

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -210,7 +210,7 @@ class Commands(object):
             owner = m.group(2)
             if not owner:
                 try:
-                    self.username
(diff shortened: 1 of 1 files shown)


praiskup (approved):
LGTM

kanru:
> Weird we haven't noticed...? Thank you for the patch.

I'm also wondering. The bug appears to be introduced in https://github.com/fedora-copr/copr/commit/dd153019487382e1aa7ed022eea8436344ac31db

I only hit this because fedora-create-review uses `copr list` to find existing copr projects. Probably most people who use fedora-create-review use koji scratch build instead.

PR #3162, opened 2024-02-29
Don't upload SRPMs/SPECs before the CoprDir is validated

praiskup:
Follow-up-for: abb1bfed77527ee732ed7b5542b5967bba937b12
Fixes: #2786

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/coprs_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/coprs_logic.py
@@ -664,7 +664,7 @@ class CoprDirsLogic(object):
         return all([c.isnumeric() for c in parts[2]])
 
     @classmethod
-    def get_or_create(cls, copr, dirname, trusted_caller=False):
+    def get_or_create(cls, copr, dirname):
         """
         Create a CoprDir on-demand, e.g. before pull-request builds is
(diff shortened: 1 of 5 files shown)


frontend/coprs_frontend/coprs/logic/coprs_logic.py
```diff
@@ -691,6 +685,26 @@ def get_or_create(cls, copr, dirname):
                 f"Wrong directory '{dirname}' specified.  Directory name can "
                 "consist of alpha-numeric strings separated by colons.")
 
+    @classmethod
+    def get_or_validate(cls, copr, dirname):
```
FrostyX, line 689: `get_or_validate` is a quite unusual concept. Do we need it? 
Can we do only `validate` in `apiv3_builds.py` and move the "get" part into `get_or_create`?

praiskup: We can do that, but wouldn't that be duplication?
I agree that's not usual :-/ though that's exactly the task we do on two places.

praiskup: I updated the doc-string to better explain why we are returning existing directories even if they do not match the current validation patterns.

FrostyX (approved):

PR #3166, opened 2024-03-01
doc: fixes for the latest release cycle

praiskup:


doc/how_to_release_copr.rst
```diff
@@ -168,6 +168,17 @@ For each package do::
     It takes up to 30 minutes to for the override to be available in the buildroot::
 
         koji wait-repo f34-build --build=python-copr-common-0.13-1.fc34
+        for ver in 38 39 40 41; do
```
FrostyX, line 171: I think we don't need to wait for the package in all Fedora versions. Only the one we have in production.

praiskup: You need to wait for the build override to be able to build the next package (for all distros).

praiskup:
I haven't prepared more more updates than originally, sorry.  But seems worth merging anyway.

PR #3171, opened 2024-03-02
frontend: opt-in the log-detective.com button

praiskup:


frontend/coprs_frontend/config/copr.conf
```diff
@@ -203,6 +203,10 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
 # builds in database, otherwise implying rather expensive SQL queries).
 #RECENT_BUILDS_ON_FRONTPAGE = False
 
+# Whether we want to show the green "Feed Log-detective.com" button for
+# failed builds.  Currently only integrated with the Fedora Copr instance.
+#LOG_DETECTIVE_FEED = False
+
```
FrostyX, line 209: Nitpick but I don't like the LOG_DETECTIVE_FEED name because of the "feed" part. It can be easily confused with a feed in the sense of RSS feed, news feed, social media feed, etc.

I'd prefer something like `LOG_DETECTIVE`, `LOG_DETECTIVE_INTEGRATION`, `LOG_DETECTIVE_ENABLED`, `LOG_DETECTIVE_BUTTON`.

praiskup: Going with LOG_DETECTIVE_BUTTON then :) we can call it "_INTEGRATION" in the future once we have more "integration" than this.

FrostyX (approved):

PR #3185, opened 2024-03-12
rpmbuild: don't clean after builds with user ssh access

### LINKED ISSUE #3181: Make "rebuild with ssh access" use "mock --no-clean-after"
Recently, copr and ui learned how to re-run a build with ssh access. This gives a chance to inspect more than the build products, especially when there are none. But so far, inspecting the build dir requires another manual mock run.

I suggest to run mock with "--no-cleanup-after" when run from "rebuild with ssh access". This saves resources because there is no need for another mock run, and the chroot is life-time limited anyways so that the secondary artefacts "lie there" for at most half an hour by default.

FrostyX:
Fix #3181

### DIFF
--- a/rpmbuild/copr_rpmbuild/builders/mock.py
+++ b/rpmbuild/copr_rpmbuild/builders/mock.py
@@ -41,6 +41,7 @@ class MockBuilder(object):
         self.isolation = task.get("isolation")
         self.macros = macros_for_task(task, config)
         self.uniqueext = get_mock_uniqueext()
+        self.allow_user_ssh = task.get("allow_user_ssh")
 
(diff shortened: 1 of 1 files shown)


praiskup:
Thank you!

Have you tested this? :-) I thought we had some cleanup logic directly in `copr-rpmbuild` but I might be wrong.

FrostyX:
I tested it on builders and IMHO works as expected `/var/lib/mock/*` is preserved.

> I thought we had some cleanup logic directly in copr-rpmbuild but I might be wrong.

I think that would be the `mock_clean` function.

praiskup (approved):

PR #3204, opened 2024-04-04
frontend: speedup the removal of temporary projects

### LINKED ISSUE #2489: Temporary projects with automated builds not deleted
I have a couple *temporary project: will be deleted ASAP* projects on Copr and they seem to be quite immortal.

https://copr.fedorainfracloud.org/coprs/churchyard/hypothesis-6.6/
https://copr.fedorainfracloud.org/coprs/churchyard/hypothesis-control/
https://copr.fedorainfracloud.org/coprs/churchyard/generators-extras-case/
https://copr.fedorainfracloud.org/coprs/churchyard/numpy-1.23/
https://copr.fedorainfracloud.org/coprs/churchyard/sequoia/
https://copr.fedorainfracloud.org/coprs/churchyard/clamp_mtime_to_source_date_epoch/
https://copr.fedorainfracloud.org/coprs/churchyard/flit-3.7.1/

Most of them have automated builds enabled, which might or might not be the reason they were not deleted.

FrostyX:
Fix #2489

I tried this on today's database dump and it took 46m24.731s to finish and my laptop was fine. I think the crash was prevented by not committing everything at once.

### DIFF
--- a/frontend/coprs_frontend/commands/clean_expired_projects.py
+++ b/frontend/coprs_frontend/commands/clean_expired_projects.py
@@ -1,6 +1,6 @@
 import click
+from coprs import db
 from . import deprioritize_actions
-from coprs import db_session_scope
 from coprs.logic.complex_logic import ComplexLogic
 
 
@@ -11,6 +11,5 @@ def clean_expired_projects():
     Clean all the expired temporary projects.  This command is meant to be
     executed by cron.
     """
-
-    with db_session_scope():
-        ComplexLogic.delete_expired_projects()
+    while ComplexLogic.delete_expired_projects(limit=100):
+        db.session.commit()
--- a/frontend/coprs_frontend/coprs/logic/complex_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/complex_logic.py
@@ -4,6 +4,7 @@ import os
 import datetime
 import time
 import fnmatch
+from itertools import batched
 import flask
 import sqlalchemy
 
@@ -15,7 +16,7 @@ from coprs import models
 from coprs import exceptions
 from coprs import cache
 from coprs.constants import DEFAULT_COPR_REPO_PRIORITY
-from coprs.exceptions import ObjectNotFound, ActionInProgressException
+from coprs.exceptions import ObjectNotFound
 from coprs.logic.builds_logic import BuildsLogic
 from coprs.logic.batches_logic import BatchesLogic
 from coprs.logic.packages_logic import PackagesLogic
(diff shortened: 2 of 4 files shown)


FrostyX:
The F38 build fails because of

> ImportError: cannot import name 'batched' from 'itertools' (unknown location)

Should I work on the compatibility or can we ditch F38 for our server packages now?

frontend/coprs_frontend/commands/clean_expired_projects.py
```diff
@@ -11,6 +11,5 @@ def clean_expired_projects():
     Clean all the expired temporary projects.  This command is meant to be
     executed by cron.
     """
-
-    with db_session_scope():
-        ComplexLogic.delete_expired_projects()
+    while ComplexLogic.delete_expired_projects(limit=100):
```
praiskup, line 14: do you want to use the limit here?

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -97,34 +98,43 @@ def delete_copr(cls, copr, admin_action=False):
...
             user = flask.g.user
 
-        builds_query = BuildsLogic.get_multiple_by_copr(copr=copr)
+        builds = [x.id for x in BuildsLogic.get_multiple_by_copr(copr=copr)]
 
         if copr.persistent:
             raise exceptions.InsufficientRightsException("This project is protected against deletion.")
 
-        for build in builds_query:
+        for chunk in batched(builds, 1000):
```
praiskup, line 106: I'd suggest much smaller chunks ... consider projects with thousands of builds.

frontend/coprs_frontend/commands/clean_expired_projects.py
praiskup: Nah, isn't this the reason why the script finished in 45 minutes?

praiskup:
https://fedorapeople.org/groups/schedule/f-38/f-38-all-tasks.html EOL in May - let's drop the F38 support

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -97,34 +98,43 @@ def delete_copr(cls, copr, admin_action=False):
...
             # Don't send delete action for each build, rather send an action to delete
             # a whole project as a part of CoprsLogic.delete_unsafe() method.
-            BuildsLogic.delete_build(user, build, send_delete_action=False)
+            BuildsLogic.delete_builds(user, chunk, send_delete_action=False)
```
praiskup, line 109: Where do we commit? At first glance, it seems we don't, and we still create a huge transaction.

frontend/coprs_frontend/commands/clean_expired_projects.py
FrostyX: Yeah, I think we want the limit here. We can change the value though.

frontend/coprs_frontend/coprs/logic/complex_logic.py
FrostyX: We commit in the

frontend/coprs_frontend/commands/clean_expired_projects.py
praiskup: ah, I missed the `s/with/while/` change, it seems OK then!  But don't we want to use a much smaller limit, like 5 :-)?

praiskup (approved):
LGTM

PR #3211, opened 2024-04-09
frontend: fix getting gssapi url_for

nikromen:


frontend/coprs_frontend/coprs/context_processors.py
```diff
@@ -58,7 +60,7 @@ def login_menu():
 
         if config['KRB5_LOGIN']:
             menu.append({
-                'link': flask.url_for("apiv3_ns.gssapi_login"),
+                'link': api.url_for(GssApiLogin),
```
praiskup, line 63: Thank you for the fix, but I don't understand why we are doing this.

frontend/coprs_frontend/coprs/auth.py
```diff
@@ -238,8 +238,15 @@ def user_from_username(username, load_metadata=False):
...
             # Pick the first one for now.
-            return flask.redirect(flask.url_for("apiv3_ns.gssapi_login",
-                                                next=next_url))
+            return flask.redirect(
+                # TODO: this is cursed...
+                #  url_for from flask works with resource class name
+                #  as snake_case. This is bullshit but the flask-restx
+                #  url_for takes as parameter the resource -> cyclic import
+                #  ... the gssapi endpoint depends on the UserAuth
+                #  but how to fix it? :clueless:
```
praiskup, line 247: Cyclic imports are stressful.  Let's have a quick chat about how to fix this, instead of this TODO which will never disappear :)

praiskup:
I'm curious whether the /gssapi/ route belongs to the normal API route category; shouldn't we put it aside the python-restx?

frontend/coprs_frontend/coprs/auth.py
FrostyX: Is this a TODO item actually? Referencing routes using a string like this ` flask.url_for("apiv3_ns._gss_api_login")` is legit, so I don't see a problem here.

The only weird thing is IMHO the leading underscore

frontend/coprs_frontend/coprs/context_processors.py
FrostyX: Using `app.url_for("apiv3_ns._gss_api_login")` here should work as well.

frontend/coprs_frontend/coprs/auth.py
FrostyX: > The only weird thing is IMHO the leading underscore

This can be fixed by e.g.

```diff
diff --git a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
index 48feb66b0..79977656f 100644
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_general.py
@@ -58,7 +58,7 @@ def krb_straighten_username(krb_remote_user):
     return username


-apiv3_general_ns = Namespace("", description="APIv3 general endpoints")
+apiv3_general_ns = Namespace("general", description="APIv3 general endpoints")
 api.add_namespace(apiv3_general_ns)
```

Then the route name would be `apiv3_ns.general_gss_api_login`. But we don't have to do this. The leading underscore is fine.

praiskup: Good catch! Nice, yes - please - rather be explicit with the Namespace naming.

praiskup (approved):
LGTM

PR #3218, opened 2024-04-13
frontend: don't pass raw SQL as string but rather SQLAlchemy text

### LINKED ISSUE #3217: Database error
A workflow failed some hours ago, and since then I see this, e.g.:

```
$ copr-cli build-package --nowait  --with-build-id 7305174 -r fedora-38-x86_64 -r fedora-39-x86_64 -r fedora-40-x86_64 -r fedora-rawhide-x86_64 cran --name R-CRAN-xfun

Something went wrong:
Error: Database error, contact admin
```

This is after starting build 7305174. And the command works fine without ` --with-build-id 7305174`.

FrostyX:
Fix #3217
See 99cc3bc

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/batches_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/batches_logic.py
@@ -7,6 +7,7 @@ import anytree
 import backoff
 
 from sqlalchemy.exc import SQLAlchemyError
+from sqlalchemy.sql import text
 from sqlalchemy_utils.functions import quote as sa_quote
 from coprs import app, db, cache
 from coprs.helpers import WorkList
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/logic/batches_logic.py
```diff
@@ -29,9 +30,9 @@ def _lock_table(table):
     # (while reading is not blocked) until the end of the transaction
     # (commit / rollback)
     with db.engine.connect() as connection:
-        connection.execute("LOCK TABLE {} IN EXCLUSIVE MODE;".format(
-            sa_quote(db.engine, table),
-        ))
+        connection.execute(text("LOCK TABLE {} IN EXCLUSIVE MODE;".format(
```
nikromen, line 33: so if `text` is needed for raw SQL, then what about these:
- https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/logic/batches_logic.py#L32
- https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/logic/builds_logic.py#L175
- https://github.com/fedora-copr/copr/blob/main/frontend/coprs_frontend/coprs/logic/builds_logic.py#L196

FrostyX: That is a good point. The first one is definitely a problem. Thank you for catching that.
The other two are running `connection.execute(query, params)` on a `query` which is a `text`.

praiskup: Let's have a new PR for the other problem.

praiskup (approved):

FrostyX:
LOL, I just noticed, that I forgot to commit also the beaker test that I wrote.

PR #3224, opened 2024-04-16
cli: remove dependency on simplejson

### LINKED ISSUE #2539: Really remove dependency on simplejson
While packaging copr-cli for openSUSE I have removed dependency on super-obsolete module `simplejson` (do you really need compatibility with Python 2.6?) with [this patch](https://github.com/fedora-copr/copr/files/10711838/remove-simplejson.patch.txt). These were forgotten with https://pagure.io/copr/copr/pull-request/350 and I am afraid I still missed something.

FrostyX:
Fix #2539

The output of `json.dumps` and `simplejson.dumps` is equivalent for listing builds and listing packages. I am removing the `for_json` parameter because there is not equivalent. However it shouldn't make any difference because:

> If for_json is true (not the default), objects with a for_json()
> method will use the return value of that method for encoding as JSON
> instead of the object.

We don't define any `for_json` methods for our objects.

### DIFF
--- a/cli/copr-cli.spec
+++ b/cli/copr-cli.spec
@@ -32,7 +32,6 @@ BuildRequires: util-linux
 %if %{with python3}
 Requires:      python3-copr >= %min_python_copr_version
 Requires:      python3-jinja2
-Requires:      python3-simplejson
 Requires:      python3-humanize
 Requires:      python3-koji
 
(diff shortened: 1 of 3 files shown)


FrostyX:
To my surprise this works even on EPEL7

praiskup (approved):

PR #3228, opened 2024-04-20
frontend: first steps towards removing unnecessary APIv3 abstractions

FrostyX:
We intentionally left a huge mess and a lot of magic in `fields.py`, `schemas.py`, and `docs.py` because it made the migration to Flask-RESTX a lot easier and we didn't know the correct abstractions beforehand.

In this patch, I am proposing several refactors to simplify the code:

- Schemas duplicated file types that were already-defined in `fields.py`. I am using the already defined ones to avoid duplication
- There seems to be no need for separately defining schemas and models now. In the end, we care only about models and schemas were IIRC only an implementation detail. I am showing how to remove some schemas and define models directly.
- Having every field defined as a variable in one giant pile in `fields.py` was a temporary measure. We can define the fields within models and then point to them to avoid duplication. See `project_fork_input_model` and its `ownername`.

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/schema/fields.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/schema/fields.py
@@ -418,13 +418,6 @@ bootstrap = String(
     )
 )
 
-confirm = Boolean(
-    description=(
-        "If forking into a existing project, this needs to be set to True,"
-        "to confirm that user is aware of that."
-    )
-)
-
 exist_ok = Boolean(
     description=(
         "Don't fail if a project with this owner and name already exist, "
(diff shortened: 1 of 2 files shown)


FrostyX:
Do you think this approach could be done for the whole APIv3 code, or we will hit a wall somewhere?
The information at the swagger page didn't seem to break and beaker tests mostly pass with exception to:

```
delete-build 2918198 2918213 2918219
```

which tracebacks. But I don't think I caused it in this PR (not sure).

praiskup:
We'll have to wait for the Pydantic support in python3-flask-restx.

The thing Jakub propose works on "outputs", but not on "user inputs" int he API.

PR #3232, opened 2024-04-22
Hide temporary projects

### LINKED ISSUE #3087: Sort the per-owner group so that temporary projects are after normal projects
For https://copr.fedorainfracloud.org/groups/g/theforeman/coprs/ we have a lot of scratch repositories (which we use to test PRs), but because of that you lose track of the real repositories. Because the real names are also prefixes for the names of temporary you can't use search either.

Some solutions that come to mind:
* Allow pinning some repositories (like GitHub does)
* Allow filtering (hide temporary repositories)
* Stricter searching

And there may be more solutions.

### LINKED ISSUE #3136: Hide temporary projects from the search results
Hide temporary projects from search results, or show them after regular projects.

I was wondering if there is a copr repository with the latest-greatest `fedora-license-data` package, so I tried searching for the name in Fedora Copr.
The results were 23 pages of Packit temporary projects. `fedora-license-data` utilizes Packit for MR builds (good!), so this is expected. However, I think it would make sense to simply hide temporary projects from Copr search by default. Right now I have no idea if there is actually a real non-temporary `fedora-license-data` Copr project, without going through 23 pages of results.

FrostyX:
Fix #3136
Fix #3087

### DIFF
--- a/frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
+++ b/frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
@@ -1,6 +1,7 @@
 # coding: utf-8
 
 import flask
+import sqlalchemy
 from flask import render_template, url_for
 from coprs.exceptions import InsufficientRightsException, ObjectNotFound
 from coprs.forms import ActivateFasGroupForm
@@ -9,7 +10,7 @@ from coprs.logic import builds_logic
 from coprs.logic.complex_logic import ComplexLogic
 from coprs.logic.coprs_logic import CoprsLogic, PinnedCoprsLogic
 from coprs.logic.users_logic import UsersLogic
-from coprs import app
+from coprs import app, models
 from coprs.oidc import oidc_enabled
 
 from ... import db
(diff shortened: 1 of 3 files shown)


frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
```diff
@@ -70,6 +71,11 @@ def list_projects_by_group(group_name, page=1):
     pinned = [pin.copr for pin in PinnedCoprsLogic.get_by_group_id(group.id)] if page == 1 else []
     query = CoprsLogic.get_multiple_by_group_id(group.id)
     query = CoprsLogic.filter_without_ids(query, [copr.id for copr in pinned])
+    query = query.order_by(
+        sqlalchemy.nullsfirst(models.Copr.delete_after),
```
praiskup, line 75: do we have index on the delete_after field?

praiskup:
Good job, not sure about the delete_after index, otherwise LGTM

frontend/coprs_frontend/coprs/views/groups_ns/groups_general.py
FrostyX: We don't. I tried to add it, and alternatively, I tried to add an index on a combination of both `delete_after` and `id`. Neither of them didn't seem to bring any performance improvement. For the record, the code was:

```python
def upgrade():
    op.create_index('copr_delete_after_id', 'copr', ['delete_after', 'id'],
                    unique=False)
```

and

```python
def upgrade():
    op.create_index('copr_delete_after_id', 'copr', ['delete_after'],
                    unique=False)
```

Possibly an error on my side? If not, do you want to add the index anyway?

praiskup: Index with predicate could help.. https://stackoverflow.com/questions/31966218/postgresql-create-an-index-to-quickly-distinguish-null-from-non-null-values ?

praiskup: I haven't found a good index either. Seems like we work with a relatively small number of rows (coprs owned by a single user), yet the query is a relatively complicated set of joins. The production DB performs very well, too. +1

praiskup (approved):
LGTM

PR #3243, opened 2024-05-02
frontend, backend: create projects in pulp

### LINKED ISSUE #3112: create project && enable/disable knob

FrostyX:
Fix #3112

Please note that this requires `python3-pulp-cli` from 
https://copr.fedorainfracloud.org/coprs/praiskup/pulp-cli/ 
and properly configured `~/.config/pulp/cli.toml` as described on 
https://github.com/fedora-copr/debate/blob/main/2023-08-20-fedora-infra-ansible-pulp.md#client-configuration


FrostyX:
I used `copr-pulp-dev.aws.fedoraproject.org` even for development so that I don't have to run my local Pulp instance. I can verify, that repositories, distributions, and publications are being created successfully:

```
pulp repository list
pulp distribution list
pulp publication list
```

but the port for listing the results (probably 24816) is being blocked by a firewall. Or some other issue. So I don't know yet if this produces a DNF-consumable repository.

FrostyX:
> but the port for listing the results (probably 24816) is being blocked by a firewall. Or some other issue. So I don't know yet if this produces a DNF-consumable repository.

It does. I couldn't figure out what the networking issue is but when I do 

```
[copr:localhost:frostyx:test-pulp-1]
name=Copr repo for test-pulp-1 owned by frostyx
baseurl=http://localhost:24816/pulp/content/frostyx/test-pulp-1/epel-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=0
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

On the actual `copr-pulp-dev.aws.fedoraproject.org` instance, the repository is being successfully refreshed.

backend/copr_backend/storage.py
```diff
@@ -0,0 +1,97 @@
...
+class Storage:
+    """
+    Storage agnostic, high-level interface for storing and acessing our data
+    """
+
+    def __init__(self, opts, log):
+        self.opts = opts
+        self.log = log
+
+    def createrepo(self, ownername, coprdir, chroot, appstream, devel):
```
praiskup, line 28: We can keep this abstractionfor now, but I fear it is not going to work with PULP where we "add RPM to repo", which "generates" version, and there's nothing like a "request for createrepo".

This is one step more on top of the original trivial issue, and I'm not against it.

backend/copr_backend/storage.py
```diff
@@ -0,0 +1,97 @@
...
+            "--name", name,
+            "--repository", repository,
+            "--base-path", basepath or name,
+        ])
+
+    def _create_publication(self, repository):
+        return run_cmd([
+            "/usr/bin/pulp", "rpm", "publication", "create",
+            "--repository", repository,
+        ])
```
praiskup, line 97: Ah, OK, I thought this is going to be next step.... :)  I really thought that we'll add just "DB and UI" knob, not functionality.  Don't you want to create two PRs?

backend/copr_backend/storage.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _create_repository(self, name):
+        return run_cmd([
+            "/usr/bin/pulp", "rpm", "repository", "create",
+            "--name", name,
+        ])
+
+    def _create_distribution(self, name, repository, basepath=None):
+        return run_cmd([
+            "/usr/bin/pulp", "rpm", "distribution", "create",
```
praiskup, line 87: Don't we want to use some Python API instead of CLI?

frontend/coprs_frontend/tests/test_apiv3/test_projects.py
```diff
@@ -152,6 +152,8 @@ def test_update_copr_api3(self):
         }, {
         }, {
             "appstream": True,
+        }, {
+            "pulp": False,
```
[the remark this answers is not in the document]
praiskup: Or, we should add this field once we need it?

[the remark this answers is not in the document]
praiskup: Scratching -> it's action, not api.

backend/copr_backend/storage.py
```diff
@@ -0,0 +1,97 @@
...
+        return call_copr_repo(repo, appstream=appstream, devel=devel,
+                              logger=self.log)
+
+
+class PulpStorage(Storage):
+    """
+    Store build results in Pulp
+    """
+
+    def createrepo(self, ownername, coprdir, chroot, appstream, devel):
```
praiskup, line 61: Is this something like "init_distribution()"...?

backend/copr_backend/storage.py
FrostyX: > This is one step more on top of the original trivial issue, and I'm not against it.

I probably misunderstood the scope of #3112. I read the "PULP: create project" as "when a Copr is created, generate a repository in Pulp for it". 

> We can keep this abstractionfor now, but I fear it is not going to work with PULP where we "add RPM to repo", which "generates" version, and there's nothing like a "request for createrepo".

All I know about Pulp is what we have in our research documents, so I wasn't sure if we can map 1:1 all of our current storage operations to what needs to be done in Pulp. This `createrepo` works as expected when creating an empty project, so I thought it is a good enough start, that will be split into multiple methods, and renamed as needed as new Pulp features are being added.

But maybe I can do one more step now and add RPM packages after builds, so the needed abstractions are clearer?

backend/copr_backend/storage.py
FrostyX: > Ah, OK, I thought this is going to be next step

I must have misunderstood the scope of #3112, sry :-)

> Don't you want to create two PRs?

Honestly, I would prefer to keep this in one to avoid the additional work of splitting it.

backend/copr_backend/storage.py
FrostyX: Totally agree. I was thinking though, whether we want to use it right now, or using commands is a good enough start, since we have them all in https://github.com/fedora-copr/debate/blob/main/2023-02-17-pulp-intro.md#demo-with-one-possible-work-flow

xsuchy:
Can we use API instead of CLI? https://docs.pulpproject.org/pulpcore/restapi.html#tag/Distributions

backend/copr_backend/background_worker_build.py
```diff
@@ -643,6 +644,14 @@ def _download_results(self):
             filter_=filter_,
         )
 
+    def _upload_results_to_storage(self):
+        """
+        Upload build results to an appropriate storage. Duplicate the data,
+        don't remove them from the original place.
```
praiskup, line 650: Can you document why we need this, namely it is weird why we split `_upload_..()` and `_cleanup()` jobs.

praiskup:
Decent start!  Thank you for moving this forward.

backend/copr_backend/storage.py
```diff
@@ -0,0 +1,194 @@
...
+    """
+
+    def __init__(self, opts, log):
+        self.opts = opts
+        self.log = log
+
+    def init_project(self, job):
+        """
+        Make sure users can enable a DNF repository for this project/chroot
+        """
```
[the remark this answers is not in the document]
FrostyX: > it doesn't seem quite obvious what is Job

The `Job` is `copr_backend.job.BuildJob`. There is one workaround in this code where I am passing a Munch instead of the `BuildJob` because the function is called from an action and there is no `BuildJob` instance for it.

> Is't it better to accept particular options here than a Munch?

There were few reasons. Some of the methods would take quite a lot of parameters and all of them from the same object. It would look something like this:

```python
storage.publish_repository(self.job.project_owner, self.job.project_name, 
                           self.job.chroot, self.job.target_dir_name, 
                           self.job.uses_devel_repo, self.job.appstream)
```

Also, we wanted the interface to be general so that other storages besides `BackendStorage` and `PulpStorage` can be added. I didn't want to predict what parameters other storages might need.

backend/copr_backend/background_worker_build.py
FrostyX: In between them, we do these steps:

```
            self._check_build_success()
            # Build _succeeded_.  Do the tasks for successful run.
            failed = False
            if self.opts.do_sign:
                self._sign_built_packages()
            self._do_createrepo()
            self._parse_results()
            build_details = self._get_build_details(self.job)
            self.job.update(build_details)
            self.job.validate()
            self._add_pubkey()
```

IIRC when I tried to do the cleanup in `_upload_results_to_storage` some of them failed because they couldn't find the RPM files.

But doing `_upload_results_to_storage` and `_cleanup` in one step should be doable. Should I try?

backend/copr_backend/storage.py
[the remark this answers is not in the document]
praiskup: > I didn't want to predict what parameters other storages might need.

This is exactly the "API" design problem that I had in mind.  When you add a new option, you are forced to fix all the implementations of the interface at the same time (this is good). Or that it is much harder to add a new implementation for the interface that can be changed so easily (new attribute of munch).  So my personal preference would be to use the verbose variant :-) but ... no hard push on this.

backend/copr_backend/actions.py
```diff
@@ -112,32 +113,27 @@ class Createrepo(Action):
...
+                # actions for which the interface is different. Until we better
+                # know what we want, I am faking the Job interface.
+                job = Munch(data)
+                job.project_owner = data["ownername"]
+                job.project_name = project_dirname
+                job.chroot = chroot
+                job.uses_devel_repo = data["devel"]
+
+                storage = storage_for_job(job, self.opts, self.log)
+                success = storage.init_project(job)
```
praiskup, line 134: No big deal either, but it would be much more convenient if had a per-project
storage object:

```
self.storage = get_storage(owner, project_name)
for dirname in project_dirnames:
    self.storage.init_dirname(dirname, chroots)
```

We wouldn't have to re-instantiate the storage object fore very dirname/chroot.

frontend/coprs_frontend/coprs/models.py
```diff
@@ -393,6 +393,11 @@ class _CoprPublic(db.Model, helpers.Serializer):
     # priority=NUMBER in repo configs
     repo_priority = db.Column(db.Integer, nullable=True)
 
+    # Should results for project be stored in Pulp instead of our backend
+    # results directory? This column will likely be useful only for a transition
+    # period until all projects are migrated to Pulp.
+    pulp = db.Column(db.Boolean, default=False, nullable=False, server_default="0")
```
praiskup, line 399: Nit: I probably mentioned this somewhere, but having something like `data_storage = "pulp"` would allow us to implement others in the future without changing the model.

praiskup:
I dropped here some notes/wishes, but if we have to merge as is, I don't see any serious reason to make us blocked.  @nikromen did you have a chance to take a look?

FrostyX:
I needed to rebase from `main` but nothing in the previous comments changed. Only the last 3 commits are new.

frontend/copr-frontend.spec
```diff
@@ -4,7 +4,7 @@
 # https://fedoraproject.org/wiki/Packaging:Guidelines#Packaging_of_Additional_RPM_Macros
 %global macrosdir       %(d=%{_rpmconfigdir}/macros.d; [ -d $d ] || d=%{_sysconfdir}/rpm; echo $d)
 
-%global copr_common_version 0.19.1~~dev
+%global copr_common_version 0.24.1.dev0
```
praiskup, line 7: Shouldn't we keep the `~~` requirement here?

backend/copr_backend/actions.py
```diff
@@ -118,19 +118,19 @@ def run(self):
...
+            # Fake a Job interface
+            # So far it has been useful to pass `Job` objects to all storage
+            # methods. Unfortunatelly jobs represent only builds and not
+            # actions for which the interface is different. Until we better
+            # know what we want, I am faking the Job interface.
+            job = Munch(data)
+            job.project_owner = data["ownername"]
+            job.project_name = project_dirname
+            job.uses_devel_repo = data["devel"]
+            storage = storage_for_job(job, self.opts, self.log)
```
praiskup, line 130: Can we set `self.storage = storage_for_job()` in constructor?

praiskup:
Not sure about the version requirement in frontend, but otherwise I'd prefer merging this so we can submit other PRs on top of this :-) thank you for this work!

frontend/copr-frontend.spec
FrostyX: I found it strange that backend and frontend used a different version format. Backend:

```
%global copr_common_version 0.21.1.dev1
```

and frontend:

```
%global copr_common_version 0.19.1~~dev
```

Since backend had a newer version, I thought that's how we do it now.

FrostyX:
Changing the version to use tildes worked. As agreed in a meeting, I am merging.

Issue #2541, opened 2023-02-14
Per project/chroot granularity 


FrostyX:
We have inconsistencies that it is possible to configure some features per project but not per chroot or the other way around. We should add some abstraction to define it once, and have it everywhere for free.

praiskup:
We should start with the --timeout option.

penguinpee:
I just ran up against this looking at the _max number of builds_ setting for packages. This one would definitely benefit from having a global per repo setting, rather than per package. The per package setting could then be used as an overwrite of the repo's default.

xambroz:
I am missing per-build option "--with something" ... currently possible on chroot level, but for rebuilding packages with circular dependencies it is bit cumbersome to change it back and forth.
More info #3272

PR #3259, opened 2024-05-13
frontend: Add per package timeout option

### LINKED ISSUE #1591: Per package timeout setting
Original issue: https://pagure.io/copr/copr/issue/1591
Opened: 2020-11-23 19:09:55

In https://copr.fedorainfracloud.org/coprs/g/python/python3.10/ we build packages automatically.

Some of them need longer timeouts than the default. This means when a new commit happens in rawhide, they build for 5 hours and fail. Then we are notified about the failure and we rebuild the package manually with increased timeout. This wastes resources (machine and human).

I'd appreciate if we could define the default timeout in package settings. If not possible, than in the project settings. Thanks for cosnidering.

cc @thrnciar 

---

#### [tstellar](https://accounts.fedoraproject.org/user/tstellar) commented at 2022-03-24 17:42:52:
+1 for the ability to set a default timeout for each package.

eclipseo:
Stop gap measure before #2541 is implemented.

Fix #1591.

Need more testing that I was able to do.

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -780,6 +780,7 @@ class Commands(object):
             "spec_template": args.spec_template,
             "python_versions": args.pythonversions,
             "max_builds": args.max_builds,
+            "timeout": args.timeout,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
         }
         if args.create:
@@ -800,6 +801,7 @@ class Commands(object):
             "scm_type": args.scm_type,
             "source_build_method": args.srpm_build_method,
             "max_builds": args.max_builds,
+            "timeout": args.timeout,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
         }
         if args.create:
@@ -817,6 +819,7 @@ class Commands(object):
             "namespace": args.namespace,
             "committish": args.committish,
             "max_builds": args.max_builds,
+            "timeout": args.timeout,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
         }
         ownername, projectname = self.parse_name(args.copr)
@@ -835,6 +838,7 @@ class Commands(object):
             "package_name": args.name,
             "gem_name": args.gem_name,
             "max_builds": args.max_builds,
+            "timeout": args.timeout,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
         }
         if args.create:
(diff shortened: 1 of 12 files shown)


cli/copr_cli/main.py
```diff
@@ -853,6 +857,7 @@ def action_add_or_edit_package_custom(self, args):
             "script_builddeps": args.script_builddeps,
             "script_resultdir": args.script_resultdir,
             "max_builds": args.max_builds,
+            "timeout": args.timeout,
             "webhook_rebuild": ON_OFF_MAP[args.webhook_rebuild],
```
FrostyX, line 861: You will have to add the parameter to `parser_add_or_edit_package_parent`. Otherwise

```
$ copr-cli edit-package-scm ... --timeout 60
copr: error: unrecognized arguments: --timeout 60
```

frontend/coprs_frontend/coprs/models.py
```diff
@@ -877,6 +877,13 @@ class Package(db.Model, helpers.Serializer, CoprSearchRelatedData):
     def validate_max_builds(self, field, value):
         return None if value == 0 else value
 
+    # max package build time
+    timeout = db.Column(db.Integer, index=True)
```
FrostyX, line 881: I don't think we want the index. There is no sorting or searching based on the timeout.

frontend/coprs_frontend/coprs/templates/coprs/detail/package.html
```diff
@@ -56,6 +56,11 @@ <h3 class="panel-title">General Information</h3>
           <dd>{{ package.max_builds }}</dd>
           {% endif %}
 
+          {% if package.timeout %}
+          <dt>Timeout:</dt>
+          <dd>{{ package.timeout }}</dd>
```
FrostyX, line 61: Can you please do `{{ build.timeout|natural_time_delta() }}`, so the time is in a human-friendly format?

FrostyX (commented):
Thank you very much for the PR @eclipseo,
I pointed out some minor issues in the code but otherwise it looks really good.

Also, unit tests are failing for some reason. Can you please take a look?
https://download.copr.fedorainfracloud.org/results/@copr/copr-pull-requests:pr:3259/fedora-40-x86_64/07444968-copr-cli/builder-live.log.gz

frontend/coprs_frontend/coprs/models.py
eclipseo: Done.

frontend/coprs_frontend/coprs/templates/coprs/detail/package.html
eclipseo: Done.

FrostyX:
Thank you very much for the changes @eclipseo

Issue #2933, opened 2023-09-27
How to handle "leftover" projects that built in rolling chroots


praiskup:
The problem with `fedora-rawhide-*` chroots is that we never make them EOL.  So far there's no automatic removal other storage consumption optimization other than [prunerepo](https://pagure.io/prunerepo/) that keeps at most one version of each package in one repository.

This is not enough;  some projects doing mass rebuilds are kept lingering around and consume a nontrivial amount of storage.  For example some projects in [this 140GB namespace inactive for 2+ years](https://copr.fedorainfracloud.org/coprs/odubaj/).  Some of those projects seem to be created for mass rebuilds, but the user is no longer active in Fedora for us to ask.  Well actually asking is a bit sub-optimal for us, it would be better to have some automation.. but dunno.

I'm not inundated with ideas, but could we e.g. take a look at RPMs in the directory, check the "buildtime" baked into RPMs, and if all the RPMs are older than say 2 years mark the project "staled" in DB?  I'm not trying to propose a policy for removals right now -> it is a hard decision to remove something when it still might be used by someone.  But for us administrators, it is hard to identify such projects to at least start some conversation about possible removals.

praiskup:
Let's do the discussion in the next 3 months.

FrostyX:
> but could we e.g. take a look at RPMs in the directory, check the "buildtime" baked into RPMs, and if all the RPMs are older than say 2 years mark the project "staled" in DB? I'm not trying to propose a policy for removals right now 

I had a similar idea but I didn't have time to think it through yet.

If I understand correctly, the issue is only with Rawhide chroots. So I was thinking about not marking the project as stale but rather doing something with Rawhide packages that were built many releases ago. For example, Rawhide is currently F40, so considering doing something with Rawhide results from the era when Rawhide was F35 and lower. They will most likely not work today.

kwk:
@praiskup I guess this is somewhat related to our internal discussion on storage size, correct? If you come up with ideas and if needed, feel free to ping us and brainstorm. We're happy to help if we can. Your team tremendously helped us in the past.

praiskup:
Thank you for the nice words, @kwk, this ticket is related to our discussion "by accident" -> I realized that your team doesn't keep data in stale/useless projects, I just needed to confirm that fact :-)  I think the point of this ticket will be to identify such projects or prevent them from occurring.  We'll see, your ideas are definitely welcome.

praiskup:
[...]

PR #3265, opened 2024-05-16
    frontend: automatically EOL lifeless rolling chroots

praiskup:
Implements: https://github.com/fedora-copr/debate/blob/main/2024-04-23-rawhide-chroots-eol.md
Fixes: #2933

- [x] Expiry/prolong works
- [x] Prolong with new build works
- [x] Users are notified about EOLed rolling chroots
- [x] Chroots are deleted eventually
- [ ] anything else??

### DIFF
--- a/doc/developer_documentation/eol-logic.rst
+++ b/doc/developer_documentation/eol-logic.rst
@@ -1,0 +1,48 @@
+.. _eol_logic:
+
+Handling EOL CoprChroot entries
+-------------------------------
+
+There are currently three cases when we schedule a CoprChroot for removal but
+preserve the data for some time to allow users to recover:
+
+    1. User disables the chroot in a project ("unclicks" the checkbox).  We give
+       them 14 days "preservation period" to reverse the decision without
+       forcing them to rebuild everything.
+    2. Copr Admin makes a chroot EOL, e.g., `fedora-38-x86_64` because the
+       Fedora 38 version goes EOL.  We keep the builds for several months, and
+       users can extend the preservation period.
+    3. We disable rolling chroots (e.g., Fedora Rawhide or Fedora ELN) after a
+       reasonable period of inactivity.
+
+There are three database fields used for handling the EOL/preservation policies:
+``CoprChroot.deleted`` (bool), ``CoprChroot.delete_after`` (timestamp),
+and ``MockChroot.is_active`` (bool, 1:N mapped to ``CoprChroot``).  The
+following table describes certain implications behind the logic:
+
+
+.. table:: Logical implications per in-DB chroot state
+
+
+    =========   ============    ======= ====================    =========       ===========
+    is_active   delete_after    deleted e-mail |br|             can build       State && |br| Description
+                                        notifications
+    =========   ============    ======= ====================    =========       ===========
+    yes         yes             yes     no                      no              |p_yel| ``preserved`` manual removal |p_end|
+    yes         no              yes     --                      no              |p_red| ``deleted`` manual removal or rolling removal (or EOL removal, and reactivated) |p_end|
+    yes         yes             no      yes                     yes             |p_yel| ``preserved`` rolling |p_end|
+    yes         no              no      --                      yes             |p_gre| ``active`` normal chroot state |p_end|
+    no          yes             yes     no                      no              |p_yel| ``preserved`` (deleted manualy, then mock chroot EOL or deactivation) |p_end|
+    no          no              yes     --                      no              |p_red| ``deleted`` manually OR rolling deleted, and THEN EOLed/deactivated by Copr admin |p_end|
+    no          yes             no      yes                     no              |p_yel| ``preserved`` mock chroot EOLed by Copr admin |p_end|
+    no          no              no      --                      no              |p_gra| ``deactivated`` deactivated by Copr admin, data preserved |p_end|
+    =========   ============    ======= ====================    =========       ===========
+
+There's also a chroot state ``expired``, which is a special state of
+the ``preserved`` state.  It is "still preserved", but the time for removal is
+already there, namely ``now() >= delete_after``.
+
+Note that when ``e-mail notifications`` are ``yes``, the time for removal has
+come (``now() >= delete_after``) and we **were not** able to send the
+notification e-mail, we **don't** remove the chroot data.  **No unexpected
+removals.**
--- a/frontend/coprs_frontend/alembic/versions/d23f84f87130_eol_rolling_builds.py
+++ b/frontend/coprs_frontend/alembic/versions/d23f84f87130_eol_rolling_builds.py
@@ -1,0 +1,30 @@
+"""
+record last copr_chroot build, allow marking Rawhide as rolling
+Create Date: 2024-05-13 08:56:31.557843
+"""
+
+import time
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.sql import text
+
+revision = 'd23f84f87130'
+down_revision = '2d1feab6b2d8'
+
+def upgrade():
+    op.add_column('mock_chroot', sa.Column('rolling', sa.Boolean(), nullable=True))
+    op.add_column('copr_chroot', sa.Column('last_build_timestamp', sa.Integer(), nullable=True))
+    conn = op.get_bind()
+    conn.execute(
+        text("update copr_chroot set last_build_timestamp = :start_stamp;"),
+        {"start_stamp": int(time.time())},
+    )
+    op.create_index('copr_chroot_rolling_last_build_idx', 'copr_chroot',
+                    ['mock_chroot_id', 'last_build_timestamp', 'delete_after'],
+                    unique=False)
+
+
+def downgrade():
+    op.drop_index('copr_chroot_rolling_last_build_idx', table_name='copr_chroot')
+    op.drop_column('copr_chroot', 'last_build_timestamp')
+    op.drop_column('mock_chroot', 'rolling')
--- a/frontend/coprs_frontend/config/copr.conf
+++ b/frontend/coprs_frontend/config/copr.conf
@@ -207,6 +207,16 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
 # failed builds.  Currently only integrated with the Fedora Copr instance.
 #LOG_DETECTIVE_BUTTON = False
 
+# After a given _INACTIVITY_WARNING period (DAYS), when no new build appeared
+# in a rolling chroot, we start the "rolling" EOL policy.  It means that — after
+# the next _INACTIVITY_REMOVAL DAYS — the chroot gets disabled, and the
+# corresponding builds are removed to save storage.  This is only applicable to
+# MockChroots.rolling = True, and the command
+# `copr-frontend eol-lifeless-rolling-chroots` must be executed periodically via
+# cron or similar.
+#ROLLING_CHROOTS_INACTIVITY_WARNING = 180
+#ROLLING_CHROOTS_INACTIVITY_REMOVAL = 180
+
 #############################
 ##### DEBUGGING Section #####
 
--- a/frontend/coprs_frontend/coprs/logic/complex_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/complex_logic.py
@@ -785,6 +785,7 @@ class ReposLogic:
         """
         In case some of the `copr_chroots` is going to be deleted in the future,
         describe the reason why
+        TODO: merge with the `CoprChroot.delete_status` property
         """
         # Do we even want to show a trash icon? I.e. are any of the project
         # chroots set to be deleted in the future?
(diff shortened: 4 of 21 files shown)


frontend/coprs_frontend/coprs/models.py
```diff
@@ -1812,9 +1819,20 @@ def delete_status(self):
...
             return ChrootDeletionStatus("active")
 
+        # EOLed rolling chroots (reactivate by a new build)
+        if self.delete_after and self.mock_chroot.rolling:
+            # We can never ever remove EOL chroots that we didn't send
+            # a notification about
+            if not self.delete_notify:
+                return ChrootDeletionStatus("preserved")
+            if self.delete_after < datetime.datetime.now():
+                return ChrootDeletionStatus("expired")
```
praiskup, line 1833: The "preserved" chroots are never removed, so we require an e-mail notification here, and return "expired".

frontend/coprs_frontend/coprs/logic/coprs_logic.py
```diff
@@ -1262,7 +1262,9 @@ def filter_outdated(cls, query):
                      .filter(models.CoprChroot.deleted.isnot(True))
 
                      # Filter only inactive (i.e. EOL) chroots
-                     .filter(not_(models.MockChroot.is_active)))
+                     .filter(or_(
+                         not_(models.MockChroot.is_active),
+                         models.MockChroot.rolling.is_(True))))
```
praiskup, line 1267: This is hard to validate :-/ sorry.  The `is_active` check exists to not "notify" users about disabled chroots, only EOLed.

praiskup: No, I don't get why we limit to only non-active chroots, but ... I had to relax this with `or_` to a) notify user so b) we can later remove.  We don't b) without a).

frontend/coprs_frontend/alembic/versions/d23f84f87130_eol_rolling_builds.py
```diff
@@ -0,0 +1,30 @@
...
+revision = 'd23f84f87130'
+down_revision = '41763f7a5185'
+
+def upgrade():
+    op.add_column('mock_chroot', sa.Column('rolling', sa.Boolean(), nullable=True))
+    op.add_column('copr_chroot', sa.Column('last_build_timestamp', sa.Integer(), nullable=True))
+    conn = op.get_bind()
+    conn.execute(
+        text("update copr_chroot set last_build_timestamp = :start_stamp;"),
+        {"start_stamp": int(time.time())},
```
praiskup, line 20: note that we do not remove anything sooner than in 360 days till we run this migration

praiskup:
Screenshot_20240520_102019

frontend/coprs_frontend/config/copr.conf
```diff
@@ -207,6 +207,15 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
...
 #LOG_DETECTIVE_BUTTON = False
 
+# After a given _INACTIVITY_WARNING period (DAYS), when no new build appeared
+# new build appeared in a rolling chroot, we start the EOL policy.  It means
+# that - after the next _INACTIVITY_REMOVAL DAYS - the chroot gets disabled, and
+# builds the corresponding builds removed to save storage.  This is only
+# applicable to MockChroots.rolling = True, and the ... TODO ... command must be
+# configured in cron.
+#ROLLING_CHROOTS_INACTIVITY_WARNING = 180
+#ROLLING_CHROOTS_INACTIVITY_REMOVAL = 180
```
praiskup, line 217: @FrostyX please take a look how often you build in your projects

FrostyX: For example these projects:

- https://copr.fedorainfracloud.org/coprs/frostyx/mandown (3 years since the last build)
- https://copr.fedorainfracloud.org/coprs/frostyx/cask (3 years since the last build)
- https://copr.fedorainfracloud.org/coprs/frostyx/light/builds/ (recent build but 4 years window between builds at some point)

I don't actively use `cask` and `mandown` but I tried to install them in a Rawhide container and they both worked. And I've been using `light` since 2017 and it always worked.

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -798,13 +799,21 @@ def _delete_reason(cls, copr_chroots):
...
         for chroot in delete_chroots:
             reason_format = "{0} and will remain available for another {1} days"
-            reason = reason_format.format(
-                "disabled by a project owner" if chroot.is_active else "EOL",
-                chroot.delete_after_days,
-            )
+            reason = "EOL"
+            if chroot.is_active:
+                if chroot.deleted:
+                    # This is in `preserved` state, not `deleted`, otherwise we
```
FrostyX, line 809: Can we change the `if` condition to include a check that `chroot.delete_after` is in the future, to make this explicit?

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -798,13 +799,21 @@ def _delete_reason(cls, copr_chroots):
...
+                    # wouldn't even ask for the _delete_reason().
+                    reason = "disabled by a project owner"
+                elif chroot.mock_chroot.rolling:
+                    reason = "a rolling chroot inactive for a long time"
+                else:
+                    raise exceptions.BadRequest(f"Unknown EOL reason {chroot.name}")
```
FrostyX, line 815: This will probably show a 500 error page, making it unable to see the "broken" project?
I don't know if we should rather log this and display a big warning saying, that Copr admins should be contacted?

frontend/coprs_frontend/coprs/models.py
```diff
@@ -1921,6 +1941,14 @@ def isolation_setup(self):
             return {}
         return settings
 
+    def build_done(self):
+        """
+        Record that a build has been submitted into this CoprChroot.
+        """
```
FrostyX, line 1947: Nit: This is an unfortunate name. It reads more like a build just finished. Can we rename to `build_submitted` please?

FrostyX (commented):
Thank you very much for the PR and I am sorry for late review.

Wow, the mechanism is complicated (not a fault of this PR, it was already complicated before). I don't know if it is worth simplifying somehow or moving to one place. Or better to keep as is.

I think I had a good idea with the `ChrootDeletionStatus` back then. Maybe it would help us to have something like `ChrootDeletionReason` to represent the descriptions in the last column of the "Logical implications per in-DB chroot state" table (e.g. `requested_by_user`, `rolling`, `eol`). Or maybe even merging those two enums together and having one enum saying that the chroot is e.g. `deleted_rolling` or `preserved_manual`. Do you think it would help?

frontend/coprs_frontend/coprs/logic/complex_logic.py
praiskup: BadRequest should be 400. This is a situation that shouldn't happen, probably caused by some wrong request, but :shrug:  I think exception should be preferred, so we fix the caller.

frontend/coprs_frontend/config/copr.conf
praiskup: Sooo, do we need an opt-out knob?  For your use-cases, it's hard to imagine an useful default for those options :-)

doc/developer_documentation/eol-logic.rst
```diff
@@ -0,0 +1,48 @@
...
+
+There are three database fields used for handling the EOL/preservation policies:
+``CoprChroot.deleted`` (bool), ``CoprChroot.delete_after`` (timestamp),
+and ``MockChroot.is_active`` (bool, 1:N mapped to ``CoprChroot``).  The
+following table describes certain implications behind the logic:
+
+
+.. table:: Logical implications per in-DB chroot state
+
+
```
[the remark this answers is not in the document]
praiskup: No, it is agnostic.  What am I missing? :-) It was meant to realize what a particular state means by just "taking a look at the database field contents".

doc/developer_documentation/eol-logic.rst
```diff
@@ -0,0 +1,48 @@
...
+    2. Copr Admin makes a chroot EOL, e.g., `fedora-38-x86_64` because the
+       Fedora 38 version goes EOL.  We keep the builds for several months, and
+       users can extend the preservation period.
+    3. We disable rolling chroots (e.g., Fedora Rawhide or Fedora ELN) after a
+       reasonable period of inactivity.
+
+There are three database fields used for handling the EOL/preservation policies:
+``CoprChroot.deleted`` (bool), ``CoprChroot.delete_after`` (timestamp),
+and ``MockChroot.is_active`` (bool, 1:N mapped to ``CoprChroot``).  The
+following table describes certain implications behind the logic:
```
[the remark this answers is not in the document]
praiskup: Can you elaborate?  Shall I update the statement?

frontend/coprs_frontend/config/copr.conf
```diff
@@ -207,6 +207,16 @@ HIDE_IMPORT_LOG_AFTER_DAYS = 14
...
 
+# After a given _INACTIVITY_WARNING period (DAYS), when no new build appeared
+# appeared in a rolling chroot, we start the "rolling" EOL policy.  It means
+# that - after the next _INACTIVITY_REMOVAL DAYS - the chroot gets disabled, and
+# the corresponding builds are removed to save storage.  This is only applicable
+# to MockChroots.rolling = True, and the command
+# `copr-frontend eol-lifeless-rolling-chroots` must be executed periodically via
+# cron or similar.
+#ROLLING_CHROOTS_INACTIVITY_WARNING = 180
+#ROLLING_CHROOTS_INACTIVITY_REMOVAL = 180
```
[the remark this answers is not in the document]
praiskup: Yeah, I was lazy to implement the `assert(WARNING < REMOVAL)`...

praiskup:
I resolved some of your remarks and marked the corresponding threads resolved.  Can you PTAL?

doc/developer_documentation/eol-logic.rst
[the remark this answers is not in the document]
FrostyX: There are some conditions that involve both `delete_after` and `delete_notify`:

```
$ git grep delete_notify |grep delete_after
frontend/coprs_frontend/coprs/models.py:1806:            if not self.delete_after and not self.delete_notify:
frontend/coprs_frontend/coprs/models.py:1822:        if not self.delete_after and not self.delete_notify:
```

so the `delete_notify` field somehow matters as well. I don't know if it needs to be mentioned in the documentation though.

doc/developer_documentation/eol-logic.rst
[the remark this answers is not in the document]
FrostyX: You are right. I was missing the fact that we set `delete_after=None` when deleting the data.

frontend/coprs_frontend/coprs/logic/complex_logic.py
FrostyX: /me nods

frontend/coprs_frontend/config/copr.conf
FrostyX: I don't know if there is something we can do. I (and maintainers in a similar situation) will receive a notification email and will probably have to rebuild the package or click an "extend" button once in a while.

FrostyX (approved):

praiskup:
We could merge, but seems there's a problem with tests.  Going to fix & merge.

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -798,13 +799,19 @@ def _delete_reason(cls, copr_chroots):
...
         reasons = {}
         for chroot in delete_chroots:
             reason_format = "{0} and will remain available for another {1} days"
-            reason = reason_format.format(
-                "disabled by a project owner" if chroot.is_active else "EOL",
-                chroot.delete_after_days,
-            )
+            reason = "EOL"
+            if chroot.is_active:
+                if chroot.deleted and chroot.delete_after is None:
```
praiskup, line 808: This `and chroot.delete_after is None` is eventually wrong;  because `.delete_after` is set to a proper value when user disables the chroot.  Going to drop it :)

praiskup: No no no, the condition is just reversed, there should be `delete_after is not None`.

frontend/coprs_frontend/config/copr.conf
praiskup: Ok, let's create the knob when users ask for it.  The effect of this change won't come earlier than after a year.

praiskup:
This part is not not related to this PR, but #3314  - merging, thank you for the review!
```
Package "util-linux-2.40.1-2.fc41.x86_64" is already installed.
Problem 1: conflicting requests
  - nothing provides python3.12dist(flask) needed by python3-flask-caching-2.3.0-1.fc41.noarch from fedora
  - nothing provides python(abi) = 3.12 needed by python3-flask-caching-2.3.0-1.fc41.noarch from fedora
 Problem 2: conflicting requests
  - nothing provides python(abi) = 3.12 needed by python3-flask-session-0.5.0-3.fc40.noarch from fedora
  - nothing provides python3.12dist(flask) >= 2.2 needed by python3-flask-session-0.5.0-3.fc40.noarch from fedora

PR #3286, opened 2024-05-26
Remove the dollar signs in the command in the readme

lachsdachs:
GitHub has this super useful feature where you can click that icon in the top right of a markdown code element to copy the text, and it kind of annoys me when i can't simply copy&paste the text to the terminal.


FrostyX (approved):
Thank you for the PR @lachsdachs,
in some parts of the documentation we use the `$` to show that the command is supposed to be executed on your machine instead of a container or a remote server (mainly the maintenance part of the documentation) but here it probably doesn't matter.

PR #3292, opened 2024-05-30
docker: don't depend other backend containers on log container

FrostyX:
It is annoying when doing for example

    docker-compose \
        -f docker-compose.yaml \
        -f docker-compose.shell.yaml \
        up -d backend-build

Then it re-creates also `backend-log`.


praiskup:
But are acutally other containers working, without `backend-log`  consuming the redis log messages?

praiskup (approved):
LGTM

PR #3296, opened 2024-06-04
utcnow() is deprecated

xsuchy:
Addressing:
/usr/bin/copr-backend-analyze-results:153: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
  timestamp = datetime.utcnow().isoformat()

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -18,12 +18,10 @@ from operator import methodcaller
 
 import traceback
 
-from datetime import datetime
+import datetime
 from threading import Thread
 
 import subprocess
-
-import pytz
 
 import munch
 from munch import Munch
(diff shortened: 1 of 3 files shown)


FrostyX (approved):

backend/copr_backend/helpers.py
```diff
@@ -452,21 +450,6 @@ def uses_devel_repo(front_url, username, projectname, project=None):
...
-#                 fcntl.flock(lfh, fcntl.LOCK_EX)
-#                 lfh.write(str(now) + ":" + msg + "\n")
-#                 fcntl.flock(lfh, fcntl.LOCK_UN)
-#         except (IOError, OSError) as e:
-#             sys.stderr.write(
-#                 "Could not write to logfile {0} - {1}\n".format(lf, str(e)))
-#     if not quiet:
-#         print(msg)
-#
-
```
FrostyX, line 469: An unrelated change but it only removes a commented code, so it's probably no problem.

xsuchy: Yes. The commented code uses utcnow(). I was torn between updating the commented code and deleting it. I choose the later as we did not use this code for ages.

PR #3303, opened 2024-06-07
docker automatic update external host ipv4 during install

andykimpe:
automatic update external host ipv4 during install

update using bind-utils dig command with Opendns Server

update all config file


praiskup:
Some of the parts are valid, would you mind opening the PR again keeping just that stuff?

PR #3311, opened 2024-06-24
backend: add support for pulp domains

FrostyX:
The pulp.stage.devshift.net instance has domains enabled while our STG instance has them disabled. As long as it is easy to support both cases, I'd do that.

### DIFF
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -5,7 +5,6 @@ Pulp doesn't provide an API client, we are implementing it for ourselves
 import os
 import tomllib
 import requests
-from six.moves.urllib.parse import urlencode
 
 
 class PulpClient:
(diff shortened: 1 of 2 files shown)


FrostyX:
The pulp.stage.devshift.net currently gives me 500 error when trying to upload an artifact, so I couldn't properly test this yet.

backend/copr_backend/pulp.py
```diff
@@ -53,7 +52,20 @@ def url(self, endpoint):
...
-        return self.config["base_url"] + self.config["api_root"] + endpoint
+        domain = self.config["domain"]
+        if domain == "default":
+            domain = ""
+
+        relative = os.path.normpath("/".join([
+            self.config["api_root"],
+            domain,
+            endpoint,
+        ]))
```
praiskup, line 63: Since we use `os.path` for urls already (sub-optimal), don't you want to use `os.path.join()`?

backend/copr_backend/pulp.py
```diff
@@ -53,7 +52,20 @@ def url(self, endpoint):
...
+
+        # Normpath removes the trailing slash. If it was there, put it back
+        if endpoint[-1] == "/":
+            relative += "/"
```
praiskup, line 67: os.path.join() preserves trailing slashes

backend/copr_backend/pulp.py
FrostyX: The problem with `os.path.join` is that it does something like this:

```python
>>> os.path.join("/foo/", "/bar/", "/baz/")
'/baz/'
```

while the `"/".join` variant produces something more expected

```python
ipdb> os.path.normpath("/".join(["/foo/", "/bar/", "/baz/"]))
'/foo/bar/baz'
```

praiskup: ah, OK

praiskup (approved):
Thank you!

PR #3322, opened 2024-07-03
backend: remove fake jobs and pass specific arguments to storage calls

FrostyX:
There is one small exception and that is `publish_repository` which takes `**kwargs` for parameters that differ across storages.


praiskup (approved):
LGTM

Issue #3334, opened 2024-07-19
URL redirect for centos stream does not work correctly


thrix:
It would be great if this:

```
❯ curl https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/repo/centos-stream-$releasever/rhcontainerbot-podman-next-centos-stream-$releasever.repo
[copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next]
name=Copr repo for podman-next owned by rhcontainerbot
baseurl=https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/centos-stream-10-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

Would return a generic copr repo with $releasever (as for Fedora):

```
❯ curl https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/repo/centos-stream-$releasever/rhcontainerbot-podman-next-centos-stream-$releasever.repo
[copr:copr.fedorainfracloud.org:rhcontainerbot:podman-next]
name=Copr repo for podman-next owned by rhcontainerbot
baseurl=https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/centos-stream-$releasever-$basearch/
type=rpm-md
skip_if_unavailable=True
gpgcheck=1
gpgkey=https://download.copr.fedorainfracloud.org/results/rhcontainerbot/podman-next/pubkey.gpg
repo_gpgcheck=0
enabled=1
enabled_metadata=1
```

lsm5:
@praiskup any ETA for this?

praiskup:
https://redhat-internal.slack.com/archives/C03BRN71JAF/p1721395805625869

xsuchy:
@thrix, Can you please summarize the motivation and use case for why you need/want this? It will help us prioritize this issue.

nikromen:
Patches are welcome for this!

praiskup:
[...]

PR #3341, opened 2024-07-24
frontend: use the same repofile for all CentOS Stream versions

nforro:
Fixes https://github.com/fedora-copr/copr/issues/3334.


praiskup (approved):
LGTM, thank you!

FrostyX:
The patch is simple enough, I am thinking about hotfixing it in production, WDYT @praiskup?

PR #3355, opened 2024-08-03
Yet another attempt to remove APIv3 magic

FrostyX:
An attempt number 356 to remove magic and excess complexity from our APIv3 schemas. I won't sleep soundly until we figure it out, so in order to preserve my own sanity, here we go.

- I am keeping schemas as `@dataclass`, we already figured out that it is what we want
- I am keeping shared fields in `fields.py` but moving only once used fields definitions directly to schemas
- I am removing some magic:
    - It is no longer possible to mark all fields as required. The largest schema that did it had 3 fields.
    - We are no longer magically matching schema fields by their name with their definitions in `fields.py`
    - Because of this ^^, we don't need the hack for renaming field names
- I simplified a lot of code in `Schema`, `InputSchema`, and `ParamsSchema`

When reviewing, I recommend going commit by commit.

In a separate follow-up PR, I'd like to take a look at the schema class hierarchy.

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/schema/fields.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/schema/fields.py
@@ -527,12 +527,3 @@ url = Url()
 webhook_rebuild = Boolean()
 
 
-def clone(field):
-    """
-    Return a copy of a field
-    """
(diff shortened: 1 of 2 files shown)


nikromen (approved):
+1, very nice improvement and keeping the dataclasses will help us with https://github.com/fedora-copr/debate/pull/9 (if we will ever get there :D)

PR #3356, opened 2024-08-04
github: add mypy workflow

FrostyX:
This will run mypy only on changed files


FrostyX:
Don't mind this PR (yet), I am just experimenting with static type checking

mfocko:
For Packit we have it in the [pre-commit hooks](https://github.com/packit/packit/blob/fafec1bd080b0f74f62cb215f59942c4ad048e9a/.pre-commit-config.yaml#L35-L54) (along with `ruff`), you can run it on GitHub via pre-commit.ci or GH Action

praiskup:
Have you considered enabling mypy in vcs diff lint?

FrostyX:
I want to also test `ruff` which is gaining popularity recently. I want to borrow a configuration from opensuse-distro-aliases
https://github.com/rpm-software-management/opensuse-distro-aliases/blob/main/.github/workflows/ci.yml
They seem to be running the check only on some diff as well (`poetry run ruff format --check --diff .`)

praiskup:
Analysing diffs is never that interesting or powerful because it has a limited scope (the diff);  what csdiff does OTOH is that it does the analysis of both old and new base codes, and only prints _new_ issues.

praiskup:
But ruff sounds very good!

PR #3361, opened 2024-08-10
backend: implement timeout fail-safe in case copr-rpmbuild hangs up

### LINKED ISSUE #3343: Some of the builds stay 20 days+ in running state
Screenshot_20240727_175217_Firefox

FrostyX:
Fix #3343

I am not 100% sure this PR will fully prevent issues like #3343 but I am trying to address cases like this here.

Normally, copr-rpmbuild is responsible for terminating itself when the build timeout is reached. This is indicated by the "Copr timeout => sending INT" message in the builder-live.log.

This proves itself to not be reliable enough because sometimes the whole copr-rpmbuild process hangs up, doesn't terminate itself, and the build gets stuck in the running state for months. This PR implements a fail-safe mechanism on backend, which waits a short period of time after the timeout is reached and then terminates the build.

Backend terminates the timeouted builds by connecting to the broken builders and running `copr-rpmbuild-cancel` on them. This successfully terminated the currently stuck builds mentioned by #3343.

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -404,10 +404,11 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
 
     def _check_build_interrupted(self):
         """
-        Should we interrupet a running worker?
+        Should we interrupt a running worker?
         """
         return (self._check_failed_resalloc_ticket()
-                or self._cancel_task_check_request())
+                or self._cancel_task_check_request()
+                or self._build_timeouted())
 
     @ttl_cache(ttl=10*60)
     def _check_failed_resalloc_ticket(self):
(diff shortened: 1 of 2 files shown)


backend/copr_backend/background_worker_build.py
```diff
@@ -404,20 +404,34 @@ def _proctitle(self, text):
...
+        When build timeouts, it should be handled by `copr-rpmbuild` and the
+        builder machine should mark itself as finished. When it fails to do so,
+        we have this fail-safe to know if a build timeouted and should be
+        terminated.
+        """
+        # Wait some time (1 hour) after the configured timeout for the builder
+        # to terminate itself.
+        timestamp = self.job.started_on + self.job.timeout + 60 * 60
+        limit = datetime.fromtimestamp(timestamp)
+        return datetime.now() > limit
```
nikromen, line 433: I think this is correct approach - killing the builder from backend when it timeouts. Handling the timeout should be done by layer above (read as - should be handled by the caller) in case of any bug inside the inner layer (builder here) which will result in stuck builder thus it never timeouts. 

Take CPU scheduler as example - the process has some time given and after that, the resource is taken from it by outer layer - the process does not decide when it want's to stop. This way the process could maliciously stole the CPU resource for itself. In our case it won't probably be maliciously but by mistake.

Considering this above, I am more inclined to handle timeout only by backend (at the timeout time, without waiting another hour) and drop the timeout handlind at builder level. I'll give it more thought after sleep tho.

praiskup: The rpmbuild logic implements a "gently" timeout; I don't think it is desirable to remove that logic (or at even spend the time on its removal).  The PR here seems like a valid implementation. +1

PR #3362, opened 2024-08-10
Replace deprecated pkg_resources with importlib.metadata

### LINKED ISSUE #3349: common: DeprecationWarning: pkg_resources is deprecated as an API.
During building on F40 I see:

```
copr_common/request.py:7
  /tmp/tito/rpmbuild-python-copr-commons0c1jtyg/BUILD/python-copr-common-git-38.8fc73f4/copr_common/request.py:7: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    import pkg_resources
```
We should handle it, because one day this will stop working.

FrostyX:
Fix #3349

### DIFF
--- a/cli/copr_cli/util.py
+++ b/cli/copr_cli/util.py
@@ -67,3 +67,26 @@ def serializable(result):
 
 def json_dumps(result):
     return json.dumps(serializable(result), indent=4, sort_keys=True)
+
+
+def package_version(name):
+    """
+    Return version of a given Python package
+
+    The `importlib.metadata` module was introduced in Python 3.8 while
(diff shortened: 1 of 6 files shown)


cli/copr_cli/util.py
```diff
@@ -67,3 +67,26 @@ def serializable(result):
 
 def json_dumps(result):
     return json.dumps(serializable(result), indent=4, sort_keys=True)
+
+
+def package_version(name):
```
nikromen, line 72: why we need to have this implemented in common and copy-paster 2x times in cli and rpmbuild instead of just importing it from common?

PR #3366, opened 2024-08-11
Don't check that specfile matches the repository name

### LINKED ISSUE #3365: Broken CI for distgit
These tests started failing all of the sudden

```
FAILED tests/test_crazy_merging.py::TestMerging::test_merged_everything - Ass...
FAILED tests/test_crazy_merging.py::TestMerging::test_diverge_middle_branches
FAILED tests/test_crazy_merging.py::TestMerging::test_no_op_1 - AssertionErro...
```

Even our current `main` is broken.

FrostyX:
Fix #3365

The rpkg-1.67-1 first introduced this check which was implemented in
https://pagure.io/rpkg/pull-request/700

At first I thought the check is valid and we only have incorrectly named spec files and repositories in our tests. But rpkg-1.67-1 actually breaks importing even for our real usage. This leads me to think that when they compare the spec and repo name, they forget to strip a namespace from the repo name.

### DIFF
--- a/dist-git/copr-dist-git.spec
+++ b/dist-git/copr-dist-git.spec
@@ -19,7 +19,7 @@ BuildRequires: systemd
 BuildRequires: python3-devel
 BuildRequires: python3-munch
 BuildRequires: python3-requests
-BuildRequires: python3-rpkg >= 1.66-6
+BuildRequires: python3-rpkg >= 1.67-1
 BuildRequires: python3-pytest
 BuildRequires: python3-copr-common >= %copr_common_version
 BuildRequires: python3-oslo-concurrency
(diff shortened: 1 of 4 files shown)


FrostyX:
This successfully fixed the #3365, the CI is failing only for Fedora Rawhide now because of a python3-oslo-concurrency dependency (we have an issue for that somewhere).

nikromen (approved):
lgtm
thanks for looking to the rpkg's PR which introduced this. I think that your [comment](https://pagure.io/rpkg/pull-request/700#comment-207011) was probably unseen. From what I've got from your comment, the `self.repo_name` fix should be fairly easy? Can we create quick and easy PR for rpkg so we can turn the check back again? Or if not, at least creating an issue would be nice, since it seems like `onosek` didn't see your comment :/

PR #3369, opened 2024-08-11
Make coprdirs have their repo available in the buildroot

FrostyX:
This is useful in general but especially for pull requests that build multiple packages which depend on each other (e.g. `python-copr` and `copr-cli`). Up until now, it had to be workarounded by building the dependency into the main copr repository.

### DIFF
--- a/beaker-tests/Sanity/copr-cli-basic-operations/runtest-coprdir.sh
+++ b/beaker-tests/Sanity/copr-cli-basic-operations/runtest-coprdir.sh
@@ -1,0 +1,67 @@
+#! /bin/bash
+#
+# Copyright (c) 2024 Red Hat, Inc.
+#
+# This program is free software: you can redistribute it and/or
+# modify it under the terms of the GNU General Public License as
+# published by the Free Software Foundation, either version 2 of
+# the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see http://www.gnu.org/licenses/.
+
+
+# Include Beaker environment
+. /usr/share/beakerlib/beakerlib.sh || exit 1
+
+# Load config settings
+HERE=$(dirname "$(realpath "$0")")
+source "$HERE/config"
+source "$HERE/helpers"
+
+
+quick_package_script ()
+{
+    cp "$HERE/files/quick-package.sh" script
+    echo "$1" >> script
+}
+
+
+rlJournalStart
+    rlPhaseStartSetup
+        setup_checks
+        rlAssertRpm "jq"
+        setupProjectName "coprdir"
+    rlPhaseEnd
+
+    rlPhaseStartTest
+        # Test that CoprDirs have their own repositories in the buildroot
+        tmp=`mktemp -d`
+        rlRun "copr-cli create --chroot $CHROOT $PROJECT"
+
+        # For an ease of implementation, the dependency and the package that
+        # requires it are both `hello`. In a real life the first package would
+        # be something like `python-copr` or `python-copr-common`, and the
+        # second package would be something like `copr-cli` or `copr-backend`.
+
+        # This is the dependency (e.g. python-copr)
+        rlRun "curl https://src.fedoraproject.org/rpms/hello/raw/rawhide/f/hello.spec > $tmp/hello-1.spec"
+        rlRun "sed -i '1s/^/Epoch: 6\n/' $tmp/hello-1.spec"
+        rlRun "copr-cli build $PROJECT:custom:foo $tmp/hello-1.spec"
+
+        # And this is the package that builds on top of it (e.g. copr-cli)
+        rlRun "curl https://src.fedoraproject.org/rpms/hello/raw/rawhide/f/hello.spec > $tmp/hello-2.spec"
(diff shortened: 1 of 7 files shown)


praiskup:
This would bring security issue into the `src.fedoraproject.org` integration (the pagure events handler)....  a single Copr project may integrate multiple packages from s.f.o. being developed by multiple packagers from s.f.o, and all such integrated packages can build from the corresponding "pull-request 1" (separate PRs, heading the same CoprDir!) and then, the package builds are done against each other.

We need to solve the s.f.o integration first, somehow.

FrostyX:
Took me only half a year but updated, PTAL.

beaker-tests/Sanity/copr-cli-basic-operations/runtest-coprdir.sh
```diff
@@ -0,0 +1,67 @@
...
+
+        # This is the dependency (e.g. python-copr)
+        rlRun "curl https://src.fedoraproject.org/rpms/hello/raw/rawhide/f/hello.spec > $tmp/hello-1.spec"
+        rlRun "sed -i '1s/^/Epoch: 6\n/' $tmp/hello-1.spec"
+        rlRun "copr-cli build $PROJECT:custom:foo $tmp/hello-1.spec"
+
+        # And this is the package that builds on top of it (e.g. copr-cli)
+        rlRun "curl https://src.fedoraproject.org/rpms/hello/raw/rawhide/f/hello.spec > $tmp/hello-2.spec"
+        rlRun "sed -i '1s/^/BuildRequires: hello >= 6:\n/' $tmp/hello-2.spec"
+        rlRun "copr-cli build $PROJECT:custom:foo $tmp/hello-2.spec"
```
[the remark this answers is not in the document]
FrostyX: I wrote this test a long time ago, so I don't remember my thought process. But there is no `--nowait` for the first build, so the second one should be submitted only after the first one is finished. Using build batches should make no difference, no?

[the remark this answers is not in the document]
praiskup: Waiting is necessary IMVHO, otherwise we are not testing anything :-/
And then, batches allow you to wait for just one of them (because the second one doesn't start untill the first one is finished).

frontend/coprs_frontend/coprs/logic/coprs_logic.py
```diff
@@ -692,8 +694,7 @@ def validate(cls, copr, dirname):
                 f"Please use directory format {copr.name}:custom:<SUFFIX_OF_CHOICE> "
                 f"or {copr.name}:pr:<ID> (for automatically removed directories)"
             )
-
-        if not all(x.isalnum() for x in dirname.split(":")[1:]):
+        if not all(x.isalnum() for x in re.split(r"[:-]+", dirname)[1:]):
```
praiskup, line 697: this deserves an in-line note :-) I'm not sure this is correct

We basically only want to allow one more `:`-separated field, right?  I'd wish we had a test-case for the validation.

frontend/coprs_frontend/tests/test_logic/test_copr_dirs_logic.py
```diff
@@ -116,4 +116,4 @@ def test_coprdir_build_normal_then_pr(self):
         response = self.test_client.get(url)
         assert response.status_code == 200
         result_dict = json.loads(response.data)
-        assert result_dict["git_repo"] == repo_url.format("test-pr-dirs:pr:1")
+        assert result_dict["git_repo"] == repo_url.format("test-pr-dirs:pr:test-copr-copr-1")
```
praiskup, line 119: This puzzles me.  :sweat_smile:  Are we only affecting the src.fp.org-triggered builds, or are we also affecting the normal `foo:pr:1` builds?  People still can submit `copr build test:pr:1 <src.rpm>`, right?

praiskup: I see, this is `PullRequestTrigger` that uses get_pagure_pr_events().  So this test-case is pagure-only.

FrostyX:
Mtg: Test for validating dir names and make sure the Cron for cleaning up old dirs still work.

FrostyX:
> Test for validating dir names 

I didn't change any code but I added a test for the validation method. All patterns that I would expect to fail, fail. @praiskup were you worried about some different ones?

> make sure the Cron for cleaning up old dirs still work.

I submitted a build:

```
copr --config ~/.config/copr.docker build-distgit frostyx/test-coprdir-deletion:pr:foo-bar-1 --name hello
```

and tried:

```
[copr-fe@frontend /]$ copr-frontend delete-dirs
```

which returned nothing. Then I did:

```
coprdb=# update build set submitted_on=1 where id=8561786;
```

and tried again:

```
[copr-fe@frontend /]$ copr-frontend delete-dirs
frostyx/test-coprdir-deletion:pr:foo-bar-1 is going to be deleted
```

Also looking at the deletion code, there doesn't seem to be any logic related to the dir name. So I think the cleanups should still work as expected.

praiskup:
Only a nit;  there are no "Fixes: " labels in git commit messages; but this certainly has some issue counterparts.

hroncok:
This means we can no longer submit all builds to `$COPR:custom:isolated` and assume they are isolated from each other. We must now send them to `$COPR:custom:isolated-$NUMBER` hen each NUMBER is unique, right?

FrostyX:
Hello @hroncok,
I think the answer to your question is "no".

Sorry, I implemented two unrelated things in this PR, to make it more confusing. The "isolation" part of this PR fixes #767, and it is only relevant for `:pr:` CoprDirs that are automatically created from pagure pull requests. This change should be NOOP for everybody who uses `:custom:`.

PR #3383, opened 2024-08-23
Point Copr repositories to Pulp

### LINKED ISSUE #3375: Pulp - fix repofiles to point to pulp

FrostyX:
Fix #3375

Of course only for Copr projects that has configured to have their storage in Pulp, not all Copr projects.

There are three possible options how to implement this.

1. The Most straightforward is to not generate repofiles ourselves and let Pulp
   do it. The downside is that we would have to pass additional parameters (it
   is possible though) such as repo priority, module hotfixes attribute, etc,
   and figure out how to count repo downloads
2. Generate repofiles ourselves, only point to `baseurl` in Pulp. This IMHO the
   best option for keeping compatibility between storage on the backend and in
   Pulp, so I implement this option in this PR.
3. Adding a Lighttpd redirect for `baseurl` of the Pulp-stored projects to Pulp
   content URL. This will be needed in the future to ensure that users don't
   have to re-enable Copr repositories, once those projects are migrated to
   Pulp. But this will be hard to implement, so we decided to do it when the
   time comes.

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -696,12 +696,6 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         :param pkg: path to the source package
 
         """
-
-        # Ideally, we would like to have this decision in our storage classes
-        if self.job.storage == StorageEnum.pulp:
-            self.log.info("Not going to sign pkgs, Pulp will take care of that")
-            return
-
         self.log.info("Going to sign pkgs from source: %s in chroot: %s",
                       self.job.task_id, self.job.chroot_dir)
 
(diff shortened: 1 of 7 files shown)


FrostyX:
I submitted this only as a draft because there was an issue with installing packages from the repositories (even though DNF was able to sync the repository metadata). It turned out to be a bug in Pulp code - https://github.com/pulp/pulp_rpm/issues/3719

I think there is no need to wait until it is fixed, this PR is ready for review.

praiskup:
Please cover dnf5 copr enable use-case, too

praiskup (approved):

Issue #3315, opened 2024-06-27
Statistic showing how many active users we have


FrostyX:
Probably just a number, how many users submitted a build within the last month. Something like that.

praiskup:
in progress, thinking about sqlalchemy query

PR #3401, opened 2024-09-04
frontend: provide statistics on how many users submitted at least one build in past month

xsuchy:
Fixes: #3315

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -130,6 +130,18 @@ class BuildsLogic(object):
         return result
 
     @classmethod
+    def get_users_by_time(cls, start, end):
+        """ Returns number of user that submitted at least one build in
+            the period.
+        """
+        result = db.session.query(models.Build.user_id)\
+            .join(models.BuildChroot)\
+            .filter(models.BuildChroot.ended_on > start)\
+            .filter(models.BuildChroot.started_on < end)\
+            .group_by(models.Build.user_id).count()
+        return result
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -129,6 +129,18 @@ def get_running_tasks_by_time(cls, start, end):
...
         return result
 
+    @classmethod
+    def get_users_by_time(cls, start, end):
+        """ Returns number of user that submitted at least one build in
+            the period.
+        """
+        result = db.session.query(models.Build.user_id)\
+            .join(models.BuildChroot)\
+            .filter(models.BuildChroot.ended_on > start)\
```
praiskup, line 139: This filter seems redundant.  To check that the user is active, it is enough to monitor the "started_on" field.

xsuchy: True for the params we are passing. But not true for any time period.
And I basically steal this from the fuction above.

xsuchy:
The failing lint is on file that I did not touch. I am going to ignore it.

praiskup:
> The failing lint is on file that I did not touch. I am going to ignore it.

Yes, please ignore.  These files have been modified recently (the warnings disappeared) in the `main` branch, and since the linter doesn't rebase - your latest commit still contains those  (the diff is done against the old and new code).

praiskup:
I believe I finally [fixed the linter](https://github.com/fedora-copr/vcs-diff-lint-action/blob/75074d785212a9276d8f5219783b7d051e8d323a/container/cmd#L91-L93), seems green now.

praiskup (approved):

Issue #2533, opened 2023-02-08
Use PULP as the RPM repository backend 


praiskup:
**This is mostly a tracking issue.**

The major benefit of this movement is to be able to grow in the future
(PULP offers features like `s3` storage backend), but also a possibility
to implement many useful features in Copr (some nice devel -> staging ->
production build promotion mechanisms, ssl-cert private content, faster
repo metadata generated from a PG database, and that's just from top of my
head).

We researched the possibilities of using PULP in Copr and it seems the
task should be pretty straightforward.  There are some missing features
though in PULP, and some steps that we need to do:

  - [ ] https://github.com/pulp/pulp_rpm/issues/2902
  - [ ] https://github.com/pulp/pulp_rpm/issues/2903
  - [ ] https://github.com/pulp/pulp_rpm/issues/2909
  - [ ] https://github.com/pulp/pulp_rpm/issues/2271
  - [ ] https://github.com/pulp/pulp_rpm/issues/2947
  - [ ] We need a to do a performance measurement (~15M of packages, in tens of thousands repositories, some repos >= 100k packages)
  - [ ] Create a Copr-specific deployment playbook(s) for PULP instance

PR #3402, opened 2024-09-05
Configurable storage for new projects

FrostyX:
See #2533

### DIFF
--- a/frontend/coprs_frontend/coprs/forms.py
+++ b/frontend/coprs_frontend/coprs/forms.py
@@ -676,6 +676,12 @@ class CoprForm(BaseForm):
         default=None,
     )
 
+    storage = wtforms.SelectField(
+        "Admin only - what storage should be used for this project",
+        choices=[(x, x) for x in ["backend", "pulp"]],
+        validators=[wtforms.validators.Optional()],
+    )
+
     @property
     def errors(self):
         """
--- a/python/copr/v3/proxies/project.py
+++ b/python/copr/v3/proxies/project.py
@@ -72,7 +72,7 @@ class ProjectProxy(BaseProxy):
             delete_after_days=None, multilib=False, module_hotfixes=False,
             bootstrap=None, bootstrap_image=None, isolation=None, follow_fedora_branching=True,
             fedora_review=None, appstream=False, runtime_dependencies=None, packit_forge_projects_allowed=None,
-            repo_priority=None, exist_ok=False):
+            repo_priority=None, exist_ok=False, storage=None):
         """
         Create a project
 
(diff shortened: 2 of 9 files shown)


frontend/coprs_frontend/coprs/forms.py
```diff
@@ -676,6 +676,12 @@ class CoprForm(BaseForm):
         default=None,
     )
 
+    storage = wtforms.SelectField(
+        "Admin only - what storage should be set for new projects",
+        choices=[(x, x) for x in ["backend", "pulp"]],
+        validators=[wtforms.validators.Optional()],
+    )
+
```
nikromen, line 684: Can we hide this option for normal users so they can't see it?

python/copr/v3/proxies/project.py
```diff
@@ -110,6 +110,7 @@ def add(self, ownername, projectname, chroots, description=None, instructions=No
             enabled together with this project repository.
         :param list packit_forge_projects_allowed: List of forge projects that
             will be allowed to build in the project via Packit
+        :param str storage: What storage should be set for new projects
```
nikromen, line 113: Please add here the comment about admin only action too

nikromen: Also I can see that in cli and frontend there are choices between pulp and BE. Can we somehow here check whether this is also storage enum or some typo nonsense?

frontend/coprs_frontend/coprs/forms.py
FrostyX: I may have missed something but users shouldn't see it already. When creating/editing a project in the web UI, the option shouldn't be there for either admins or normal users. For API, it will probably have to be visible?

Also, I just realized that I forgot to add the field into the API schema and it works regardless. Is this expected @nikromen? I will add it though :-)

python/copr/v3/proxies/project.py
FrostyX: > Please add here the comment about admin only action too

Good idea, I will mention it.

> Can we somehow here check whether this is also storage enum or some typo nonsense?

That would be a precedent because so far we don't have validation for any fields in `python-copr`, only in CLI and then on frontend. Having validation there could be nice if the code could be shared between `python-copr` and `copr-cli` but I don't like the idea of duplicating the validation code on yet another place.

FrostyX: > Please add here the comment about admin only action too

Done

frontend/coprs_frontend/coprs/forms.py
FrostyX: >  I will add it though :-)

Done

praiskup (approved):

PR #3403, opened 2024-09-05
Add pulp container to our docker-compose

FrostyX:
See #2533

We are using a multi-process container
https://pulpproject.org/pulp-oci-images/docs/admin/guides/deploy-multi-process-images/

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -182,6 +182,22 @@ services:
       - .:/opt/copr:z
       - copr-keygen:/var/lib/copr-keygen:z
 
+  pulp:
+    image: ghcr.io/pulp/pulp:latest
+    hostname: pulp
+    stdin_open: true
+    tty: true
+    environment:
+      - PULP_DEFAULT_ADMIN_PASSWORD=admin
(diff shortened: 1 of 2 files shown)


nikromen (commented):
Could you please add `pulp` cli as dependency to the backend? Apart from that, works for me \o/
```
bash-5.2$ lftp http://pulp:80/pulp/content/
cd ok, cwd=/pulp/content
lftp pulp:/pulp/content> ls
lftp pulp:/pulp/content>
```

FrostyX:
> bash-5.2$ lftp http://pulp:80/pulp/content/

Oh nice, I am glad this works too. I tested only the Pulp API.

> Could you please add pulp cli as dependency to the backend?

I was thinking about it but I wasn't sure since the pulp container is optional (by default projects are stored on backend)

praiskup (approved):

nikromen:
> I was thinking about it but I wasn't sure since the pulp container is optional
ah, that makes sense!

nikromen (approved):

PR #3342, opened 2024-07-24
Webhook request history

jaitjacob:
This is a two part PR. 
1. To implement the database side of things(models, migrations and data insertion up webhook arrival).
2. To implement the UI which is to display the webhook history records in one of a chosen Project page.

Here's what I've done so far, 
1. Add a new table named `webhook_history` to `coprdb` DB. 
   - It has three Columns. 
   - `id` : An auto incrementing primary key.
   - `hook_uuid` : This is extracted from the webhook POST request. 
   - `build_id` : Build number that was initiated by a webhook. 
   - Note: A webhook may initiate multiple builds this implies there may be multiple records inserted with the same `hook_id`. 

2. A function named `def addHookHistoryRecord(hook_uuid, builds_initiated_via_hook)` is used to insert a record in `webhook_history` table. As of now, I am calling it only within Github webhook handler. 

3. Migration script using Alembic. 

4. There are NO tests as of now. I expect changes so holding that for now. 

Closes #304

PR #3407, opened 2024-09-06
frontend: show Webhook History under Settings->Integrations

jaitjacob:
This is part 2 of PR #3342 that implements a table under Settings -> Integration page displaying data from `webhook_history` table. See image below for demo.
image

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/webhooks_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/webhooks_logic.py
@@ -1,0 +1,22 @@
+"""
+    Module for Webhooks History related backend logic.
+"""
+
+from coprs import db, models
+class WebhooksLogic:
+    """
+        Class to retrieve Webhook History records from database.
+    """
+    @classmethod
+    def get_all_webhooks(cls, copr):
+        """
+        Returns all webhooks received for a copr in newest first order.
+        """
+        builds = (db.session.query(models.Build).join(models.WebhookHistory)
+                            .filter(models.Build.webhook_history is not None,
+                                    models.Build.copr_id == copr.id).all())
+
+        #Remove any duplicates from the webhook history list.
+        webhook_history = {x.webhook_history for x in builds}
+
+        return webhook_history
--- a/frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
+++ b/frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
@@ -1,6 +1,6 @@
 {% extends "coprs/detail/settings.html" %}
 
-{% from "_helpers.html" import render_field, render_form_errors, copr_url %}
+{% from "_helpers.html" import render_field, render_form_errors, copr_url, initialize_datatables %}
 
 {% set selected_monitor_tab = "integrations" %}
 {%block settings_breadcrumb %}Integrations{% endblock %}
@@ -66,6 +66,31 @@
         </div>
         <div class="well well-sm">
             {{ custom_dir_url }}
+        </div>
+
+        <h3>Webhook History</h3>
+        <div>
+            {% if webhook_history %}
+            <table class="datatable dataTable table table-striped table-bordered">
+                <thead>
+                  <tr>
+                    <th>Timstamp</th>
+                    <th>Webhook ID</th>
+                    <th>User Agent</th>
+                  </tr>
+                </thead>
+                {% for webhook in webhook_history %}
+                <tr>
+                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}})</td>
+                    <td>{{webhook.webhook_uuid}} </td>
+                    <td>{{webhook.user_agent}}</td>
+                </tr>
+                {% endfor %}
+            </table>
+            {% else %}
+                <p>No webhook received recently.</p>
+            {% endif %}
+            {{initialize_datatables(order="desc")}}
         </div>
     </div>
 </div>
--- a/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
+++ b/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import tempfile
+import time
 import shutil
 from typing import Optional
 
@@ -91,7 +92,8 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
         log.debug("No build initiated. Webhook not logged to db.")
         return
 
-    webhookRecord = models.WebhookHistory(webhook_uuid=webhook_uuid,
+    webhookRecord = models.WebhookHistory(created_on=int(time.time()),
+                                          webhook_uuid=webhook_uuid,
                                           user_agent=user_agent)
     db.session.add(webhookRecord)
     db.session.commit()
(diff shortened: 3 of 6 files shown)


praiskup:
Thank you for the PR!

How difficult would be to also present a list of builds (if any) submitted by each webhook?

Otherwise, I'm curious if we need to make this page private eventually.  Opinions?

FrostyX:
> I'm curious if we need to make this page private eventually.

The whole page should already be private. Accessing https://copr.fedorainfracloud.org/coprs/g/copr/copr/integrations/ in a private window automatically redirects me to login

> How difficult would be to also present a list of builds (if any) submitted by each webhook?

I would be careful here, so it doesn't start easy and then become annoying for large projects. What about going the other way around and in our current builds table or build details linking the webhook?

praiskup:
> The whole page should already be private

And do we want it?  Is there a reason to?

praiskup:
I mean, the question is "if we could present the NEW data publicly", not to make the access credentials "public" :-) wich would basically imply a completely new route.

frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
```diff
@@ -67,6 +67,42 @@ <h3> Custom webhook(s) </h3>
...
+                <thead>
+                  <tr>
+                    <th>Timstamp</th>
+                    <th>Webhook ID</th>
+                    <th>User Agent</th>
+                  </tr>
+                </thead>
+                {% for webhook in webhook_history %}
+                <tr>
+                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.timestamp}}">{{webhook.timestamp|humanize_datetime()}}</td>
```
FrostyX, line 84: Looking at some dates that we already print on the page, e.g.
https://copr.fedorainfracloud.org/coprs/g/copr/copr/build/7470434/

The date format is `2024-05-21 12:49 UTC (3 months ago)`. I'd like to have the same format also here. Please see my comment above regarding the `localized_time` filter.

Then the `title` could be the UNIX timestamp?

FrostyX (commented):
Thank you very much for the PR @jaitjacob, a good progress there. Everything works as expected, I only have some suggestions to improve the implementation details.

frontend/coprs_frontend/coprs/logic/webhooks_logic.py
```diff
@@ -0,0 +1,28 @@
...
+                    app.logger.info(
+                    "Querying webhook_history by ID: '%s' associated with build.",
+                        build.webhook_history_id)
+                    webhook = (models.WebhookHistory.query
+                            .filter_by(id=build.webhook_history_id).one())
+                    webhook_history.append(webhook)
+
+        #Remove any duplicates from the webhook history list.
+        seen = set()
+        webhook_history = [wh for wh in webhook_history if not (wh.id in seen or seen.add(wh.id))]
```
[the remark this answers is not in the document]
jaitjacob: New commit will have simpler list comprehension along with DB to return only what's needed(builds that have a webhook ID).  

`webhook_history = {x.webhook_history for x in builds}`

FrostyX:
Thank you for the update, the code is much simpler now :-)

The `humanize_datetime` filter works well but we already have `time_ago` to do the same thing, so adding it makes me a little uneasy. I realize that `time_ago` doesn't work in this case because it expects the input to be a timestamp, not datetime. But I'd rather modify it with something like:

```diff
--- a/frontend/coprs_frontend/coprs/filters.py
+++ b/frontend/coprs_frontend/coprs/filters.py
@@ -165,6 +165,9 @@ def time_ago(time_in, until=None):
     """
     if time_in is None:
         return " - "
+
+    if isinstance(time_in, datetime.datetime):
+        now = time_in
     if until is not None:
         now = datetime.datetime.fromtimestamp(until)
     else:
```

instead of adding a new filter.

FrostyX:
> But I'd rather modify it with something like:
> if isinstance(time_in, datetime.datetime):

I can see you already do this workaround in another function. Maybe it would be even better to add a new migration to change `WebhookHistory.timestamp` from

```python
timestamp = db.Column(DateTime(timezone=True), server_default=func.now())
```

to

```python
created_on = db.Column(db.Integer, nullable=True)
```

so it is consistent with our other timestamps and therefore could use the same function without workarounds. What do you think?

jaitjacob:
Sounds good to me. From a maintenance and readability POV it would be better to stay consistent with how date and times are handled within the codebase. In hindsight, praiskup did mention a preference over UNIX timestamps, I see why.

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
```diff
@@ -91,7 +92,8 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
         log.debug("No build initiated. Webhook not logged to db.")
         return
 
-    webhookRecord = models.WebhookHistory(webhook_uuid=webhook_uuid,
+    webhookRecord = models.WebhookHistory(created_on=int(time.time()),
```
FrostyX, line 95: @praiskup would you expect to save the timestamp of when we received the webhook, or would you expect to have a timestamp parsed from the webhook itself?

praiskup: Not sure, do we want to have both?  I think the one "when it was delivered to us" is more important because we can be curious in which order they were delivered.  But no strong preference.

praiskup:
> In hindsight, praiskup did mention a preference over UNIX timestamps,

Hehe, I don't recall stating my preference - but indeed +1 from me for timestamps.  The major "why?" benefit for me is that it stupid simple format and it has the same value on every single place on this planet at the same time :-) - it's up to the client/UI to convert it like:

```
$ date --date "@$(date +%s)"
pá 20. září 2024, 12:02:24 CEST
```

I'm not even against having the sub-second resolution.

frontend/coprs_frontend/coprs/filters.py
```diff
@@ -131,7 +131,12 @@ def localized_time(time_in, timezone):
         user_tz = pytz.timezone(timezone)
     else:
         user_tz = utc_tz
-    dt_aware = datetime.datetime.fromtimestamp(time_in).replace(tzinfo=utc_tz)
+
+    dt_aware = time_in
+
+    if not isinstance(time_in, datetime.datetime):
+        dt_aware = datetime.datetime.fromtimestamp(time_in).replace(tzinfo=utc_tz)
+
```
FrostyX, line 139: We don't need this change now

frontend/coprs_frontend/coprs/filters.py
```diff
@@ -167,7 +172,6 @@ def time_ago(time_in, until=None):
     diff = now - datetime.datetime.fromtimestamp(time_in)
     return humanize.naturaldelta(diff)
 
-
```
FrostyX, line 170: A leftover blank line :-)

FrostyX (commented):
Pylint complains about this:

>  Error: PYLINT_WARNING:
frontend/coprs_frontend/coprs/models.py:26: W0611[unused-import]: Unused func imported from sqlalchemy.sql

frontend/coprs_frontend/coprs/filters.py
jaitjacob: hmm, I will fix this line. However, I do notice neighbouring filter methods in `filters.py` have double blanks between them.

FrostyX:
Everything looks good to me. Can you please squash all the commits together into one, so that I can merge?

FrostyX (approved):

praiskup (approved):

PR #3408, opened 2024-09-09
rpmbuilds: dynamically specify tmp_fs size

### LINKED ISSUE #3268: Create a pool of builders that have more storage
Follow up of https://github.com/fedora-copr/copr/issues/3267

We have powerful builders, but this adds "only" CPU and RAM. But not storage.
We have a hardcoded option to mock config to use 140GB for storage using tmpfs
  https://pagure.io/fedora-infra/ansible/blob/main/f/roles/copr/backend/files/provision/files/mock/site-defaults.cfg#_18
We must pass an option allowing powerful builders to use more storage.

nikromen:
This allows us to specify tpm fs size to rpmbuild in order to be able to
automatically generate its size for performance builders.

before merge (please review):
- [ ] update vm-provisioning - https://pagure.io/fedora-infra/ansible/pull-request/2241
- [x] update resalloc cloud providers - https://github.com/fedora-copr/resalloc-ibm-cloud/pull/28

Fixes #3268

### DIFF
--- a/rpmbuild/copr_rpmbuild/builders/mock.py
+++ b/rpmbuild/copr_rpmbuild/builders/mock.py
@@ -6,12 +6,14 @@ import shutil
 import subprocess
 
 from jinja2 import Environment, FileSystemLoader
-from ..helpers import (
+from copr_rpmbuild.config import Config
+from copr_rpmbuild.helpers import (
     locate_spec,
     CONF_DIRS,
     get_mock_uniqueext,
     GentlyTimeoutedPopen,
     macros_for_task,
+    mock_snippet_for_tags,
 )
 
 log = logging.getLogger("__main__")
@@ -42,6 +44,10 @@ class MockBuilder(object):
         self.macros = macros_for_task(task, config)
         self.uniqueext = get_mock_uniqueext()
         self.allow_user_ssh = task.get("allow_user_ssh")
+        self.tags = task.get("tags", [])
+
+        self.copr_rpmbuild_config = Config()
+        self.copr_rpmbuild_config.load_config()
 
     def run(self):
         open(self.logfile, 'w').close() # truncate logfile
@@ -82,6 +88,9 @@ class MockBuilder(object):
             copr_build_id=self.build_id,
             isolation=self.isolation,
             macros=self.macros,
+            mock_snippet=mock_snippet_for_tags(
+                self.copr_rpmbuild_config.tags_to_mock_snippet, self.tags
+            ),
         )
 
     def produce_srpm(self, spec, sources, resultdir):
--- a/rpmbuild/copr_rpmbuild/config.py
+++ b/rpmbuild/copr_rpmbuild/config.py
@@ -1,0 +1,29 @@
+"""
+Configuration class for copr-rpmbuild
+"""
+
+import yaml
+
+
+CONFIG_PATH = "/etc/copr-rpmbuild/copr-rpmbuild.yml"
+
+
+class Config:
+    """
+    Configuration class for copr-rpmbuild
+    """
+    def __init__(self):
+        self.tags_to_mock_snippet = []
+
+    def load_config(self):
+        """
+        Load configuration from the config file
+        """
+        config_data = {}
+        try:
+            with open(CONFIG_PATH, "r", encoding="utf-8") as file:
+                config_data = yaml.safe_load(file) or {}
+        except FileNotFoundError:
+            pass
+
+        self.tags_to_mock_snippet = config_data.get("tags_to_mock_snippet", [])
--- a/rpmbuild/copr_rpmbuild/helpers.py
+++ b/rpmbuild/copr_rpmbuild/helpers.py
@@ -469,3 +469,20 @@ def package_version(name):
             return pkg_resources.require(name)[0].version
         except pkg_resources.DistributionNotFound:
             return "git"
+
+
+def mock_snippet_for_tags(tags_to_mock_snippet, tags):
+    """
+    Return mock snippets as string separated by newlines for a given
+    list of tags.
+    """
+    if not tags or not tags_to_mock_snippet:
+        return ""
+
+    tags_set = set(tags)
+    snippets = []
+    for item in tags_to_mock_snippet:
+        if set(item["tagset"]).issubset(tags_set):
+            snippets.append(item["snippet"])
+
+    return "\n".join(snippets)
--- a/rpmbuild/mock-custom-build.cfg.j2
+++ b/rpmbuild/mock-custom-build.cfg.j2
@@ -16,6 +16,9 @@ config_opts["root"] = "copr-custom-" + config_opts["root"]
 # /bin/mock calls (when tmpfs_enable is on).
 config_opts['plugin_conf']['tmpfs_opts']['keep_mounted'] = True
 
+# Custom mock snippets configured in copr-crpmbuild config file - can be empty
+{{ mock_snippet }}
+
 {%- for key, value in macros.items() %}
 config_opts['macros']['{{ key }}'] = '{{ value }}'
 {%- endfor %}
--- a/rpmbuild/tests/test_mock.py
+++ b/rpmbuild/tests/test_mock.py
@@ -147,6 +147,9 @@ config_opts.setdefault('plugin_conf', {})
 config_opts['plugin_conf'].setdefault('tmpfs_opts', {})
 config_opts['plugin_conf']['tmpfs_opts']['keep_mounted'] = True
 
+# Custom mock snippets configured in copr-crpmbuild config file - can be empty
+
+
 
 config_opts['chroot_additional_packages'] = 'pkg1 pkg2 pkg3'
 
@@ -163,6 +166,30 @@ config_opts['macros']['%dist'] = '%nil'
 config_opts['macros']['%_disable_source_fetch'] = '0'
 
 """  # TODO: make the output nicer
+
+    @mock.patch("copr_rpmbuild.builders.mock.subprocess.call")
+    # pylint: disable-next=unused-argument, redefined-outer-name
+    def test_mock_config_hp_fs_size(self, call, f_mock_calls):
+        """ test that fs_size for performance builders is correctly set """
+        self.task["tags"] = ["blabla_tag", "on_demand_powerful", "aarch64"]
+        builder = MockBuilder(
+            self.task, self.sourcedir, self.resultdir, self.config
+        )
+        mock_snippet = "config_opts['plugin_conf']['tmpfs_opts']['max_fs_size'] = '280g'"
+        builder.copr_rpmbuild_config.tags_to_mock_snippet = [
+            {
+                'tagset': ['on_demand_powerful', "aarch64"],
+                'snippet': mock_snippet,
+            },
+        ]
+        builder.run()
+
+        with open(self.child_config, "r", encoding="utf-8") as f:
+            config = f.readlines()
+
+        config = ''.join(config)
+        assert mock_snippet in config
+
 
     @mock.patch("copr_rpmbuild.builders.mock.MockBuilder.prepare_configs")
     @mock.patch("copr_rpmbuild.builders.mock.MockBuilder.archive_configs")
--- a/rpmbuild/tests/test_mock.py
+++ b/rpmbuild/tests/test_mock.py
@@ -227,7 +227,7 @@ config_opts['macros']['%_disable_source_fetch'] = '0'
              '--scrub', 'root-cache', '--quiet'])
         assert get_mock_uniqueext_mock.call_count == 1
 
-    @mock.patch('{0}.open'.format(builtins), new_callable=mock.mock_open())
+    @mock.patch('{0}.open'.format(builtins), new_callable=mock.mock_open)
     def test_touch_success_file(self, mock_open):
         builder = MockBuilder(self.task, self.sourcedir, self.resultdir, self.config)
         builder.touch_success_file()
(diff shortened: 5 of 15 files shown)


nikromen:
"deployed" on dev and works

powerful:
https://copr.stg.fedoraproject.org/coprs/g/copr/powerful-test/build/2918769/

normal:
https://copr.stg.fedoraproject.org/coprs/nikromen/not-pulp/build/2918768/

powerful builders have 360GB volume. (how do I confirm the max_fs size of 280gb?)

FrostyX:
LGTM. The only thing worth pointing out is the hardcoded volume size numbers but I think it's fine.

rpmbuild/copr_rpmbuild/builders/mock.py
```diff
@@ -42,6 +42,10 @@ def __init__(self, task, sourcedir, resultdir, config):
         self.macros = macros_for_task(task, config)
         self.uniqueext = get_mock_uniqueext()
         self.allow_user_ssh = task.get("allow_user_ssh")
+        self.tags = task.get("tags", [])
+        self.max_fs_size = "140g"
+        if "on_demand_powerful" in self.tags:
+            self.max_fs_size = "280g"
```
praiskup, line 48: I'm not against having (some) default, but in 99% this is deployment-specific info.  And 140G is _typically_ quite too much for local reproducibility (the user tries to run the copr-rpmbuild locally to reproduce a build failure, e.g.).

praiskup:
> powerful builders have 320GB volume (304 swap). (how do I confirm the max_fs size of 280gb?)

On my system I run:
```
mock_chroot_tmpfs on /var/lib/mock/fedora-40-x86_64-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-40-x86_64/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-rawhide-ppc64le-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-rawhide-ppc64le/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-39-ppc64le-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-41-x86_64-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-39-ppc64le/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-41-x86_64/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-38-ppc64le-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/fedora-38-ppc64le/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/rhel+epel-9-ppc64le-bootstrap/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
mock_chroot_tmpfs on /var/lib/mock/rhel+epel-9-ppc64le/root type tmpfs (rw,relatime,seclabel,size=146800640k,nr_inodes=0,mode=755,inode64)
```
I believe the `size=*` mount argument is related.

nikromen:
> Not needed right now IMVHO, or has s390x support been requested?

this needs to be done for every arch that has `on_demand_powerful` supported - so even s390x

rpmbuild/mock-custom-build.cfg.j2
```diff
@@ -16,6 +16,12 @@ config_opts["root"] = "copr-custom-" + config_opts["root"]
 # /bin/mock calls (when tmpfs_enable is on).
 config_opts['plugin_conf']['tmpfs_opts']['keep_mounted'] = True
 
+{%- if mock_snippets %}
+{% for snippet in mock_snippets %}
+'{{ snippet }}'
```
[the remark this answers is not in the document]
praiskup: Nb. the outer quotes seem wrong, and breaking the syntax of configuration file.

rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -469,3 +469,15 @@ def package_version(name):
...
+
+def mock_snippets_for_tags(job_tag_to_mock_snippets, tags):
+    """
+    Return a list of mock snippets for a given list of tags
+    """
+    mock_snippets = []
+    for tag in tags:
+        snippet = job_tag_to_mock_snippets.get(tag)
+        if snippet:
+            mock_snippets.append(snippet)
```
praiskup, line 482: Shouldn't we use something like this?

```python
mock_snippet = ""
for tags_to_snippet in job_tags_to_mock_snippet:
  configured_tagset = tags_to_snippet["tags"]
  mock_snippet = tags_to_snippet["mock_snippet"]
  if configured_tagset.issubset(tags):
    mock_snippet += mock_snippet + "\n"
return mock_snippet
```

This way you may configure different snippet for `x86_64+on_demand` and
for `aarch64+on_demand`.

rpmbuild/tests/test_mock.py
```diff
@@ -164,6 +165,21 @@ def test_mock_config(self, call, f_mock_calls):
 
 """  # TODO: make the output nicer
 
+    @mock.patch("copr_rpmbuild.builders.mock.subprocess.call")
+    # pylint: disable-next=unused-argument, redefined-outer-name
+    def test_mock_config_hp_fs_size(self, call, f_mock_calls):
+        """ test that fs_size for performance builders is correctly set """
+        self.task["tags"] = ["blabla_tag", "on_demand_powerful"]
```
praiskup, line 172: Thank you for the test-case, very appreciated!

praiskup:
> this needs to be done for every arch that has on_demand_powerful supported - so even s390x

Not really, we can have different configurations for different architectures;  but it seems the original requestor wanted to have support for all the architectures.

praiskup:
Please provide the configuration file as `%config(noreplace)` in the RPM file,
and self-document the option.  Something like:

```yaml
---
# Configure special mock configuration snippets per given set of tags.
#tags_to_mock_snippets:
#  - tags:
#      - on_demand_powerful
#      - arch_x86_64
#    snippet: |
#        config_opts['plugin_conf']['tmpfs_opts']['max_fs_size'] = '140g'
```

rpmbuild/copr_rpmbuild/helpers.py
praiskup: I think I'd prefer to map "tag sets -> snippets" (while now it is for free, it might be a complicated move in the future).  But I don't really insist.

praiskup:
This is in a good shape, thank you for the update! Two small things that would make it perfect :) but basically good to go.

rpmbuild/copr_rpmbuild/helpers.py
nikromen: I still don't see how the final conf file should look like then? For me this way it's somehow counter-intuitive to have so much complicated yaml file with "tags" and "mock_snippet" fields - it's 3x times nested just to add one line to config file which sound too complicated to me.

rpmbuild/copr_rpmbuild/builders/mock.py
```diff
@@ -12,7 +12,9 @@
     get_mock_uniqueext,
     GentlyTimeoutedPopen,
     macros_for_task,
+    mock_snippet_for_tags,
 )
+from copr_rpmbuild.config import Config
```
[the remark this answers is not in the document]
praiskup: Yeah, please either swap thos imports, or even better - replace also the previous relative import with an absolute one.

rpmbuild/copr_rpmbuild/config.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def load_config(self):
+        """
+        Load configuration from the config file
+        """
+        config_path = None
+        for file_type in ALLOWED_FILE_TYPES:
+            if os.path.exists(f"{CONFIG_PATH}.{file_type}"):
+                config_path = f"{CONFIG_PATH}.{file_type}"
+                break
```
praiskup, line 29: There is a TOCTOU issue, and a file-not-found bug.  I don't think we need to support two suffixes, so simpler approach:

```patch
-CONFIG_PATH = "/etc/copr-rpmbuild/copr-rpmbuild"
-ALLOWED_FILE_TYPES = ["yaml", "yml"]
+CONFIG_PATH = "/etc/copr-rpmbuild/copr-rpmbuild.yml"
 
 
 class Config:
@@ -22,16 +21,10 @@
         """
         Load configuration from the config file
         """
-        config_path = None
-        for file_type in ALLOWED_FILE_TYPES:
-            if os.path.exists(f"{CONFIG_PATH}.{file_type}"):
-                config_path = f"{CONFIG_PATH}.{file_type}"
-                break
-
-        if config_path is None:
-            return
-
-        with open(CONFIG_PATH, "r", encoding="utf-8") as file:
-            config_data = yaml.safe_load(file) or {}
-
+        config_data = {}
+        try:
+            with open(CONFIG_PATH, "r", encoding="utf-8") as file:
+                config_data = yaml.safe_load(file) or {}
+        except FileNotFoundError:
+            pass
         self.job_tag_to_mock_snippets = config_data.get('job_tag_to_mock_snippets', {})
```

nikromen: > I don't think we need to support two suffixes

sooner or later someone will wonder why yaml file doesn't work because of this... I did last week :D

the bug was with CONFIG_FILE -> config_file so this should fix it... and I don't see where's the TOCTOU issue :/

praiskup: > I did last week :D

This mistake is unlikely to be repeated (we already provide the "stub" config file for in-place edists).

It's similar like with `.cfg`  vs. `.conf` vs. `.ini`.

> I don't see where's the TOCTOU issue

If file found => open it.  It may be created after doing the check (you should always open the file, and react if the file is not found).

rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -469,3 +469,20 @@ def package_version(name):
...
+    Return mock snippets as string separated by newlines for a given
+    list of tags.
+    """
+    if not tags or not job_tag_to_mock_snippets:
+        return ""
+
+    tags_set = set(tags)
+    mock_snippets = []
+    for tag, snippet in job_tag_to_mock_snippets.items():
+        if set(tag.split("+")).issubset(tags_set):
```
praiskup, line 485: Don't you want to accept list instead?  What if the `+` symbol is part of the label?

praiskup: Check this option:
```
- tags:
    - on_demand_powerful
    - arch_x86_64
  the_snippet: |
    some_multi_
    line_snippet of your choice

File: rpmbuild/copr_rpmbuild/config.py (hunk 6 above)
Comment (nikromen) [reply rpmbuild/copr_rpmbuild/config.py:29]: if file is not found in this case it'll use the default config so the end result is the same anyway

File: rpmbuild/copr_rpmbuild/config.py [hunk 8]
```diff
@@ -0,0 +1,37 @@
...
+    def __init__(self):
+        self.job_tag_to_mock_snippets = {}
+
+    def load_config(self):
+        """
+        Load configuration from the config file
+        """
+        config_path = None
+        for file_type in ALLOWED_FILE_TYPES:
+            if os.path.exists(f"{CONFIG_PATH}.{file_type}"):
```
praiskup, line 27: Use of `os.path.exists` ;) it's an antipattern, you just do two syscalls without reason.  Just open the file, done.

rpmbuild/copr_rpmbuild/config.py
```diff
@@ -0,0 +1,37 @@
...
+Configuration class for copr-rpmbuild
+"""
+
+
+import os
+import yaml
+
+
+CONFIG_PATH = "/etc/copr-rpmbuild/copr-rpmbuild"
+ALLOWED_FILE_TYPES = ["yaml", "yml"]
```
praiskup, line 11: useless complication, but not a blocker

rpmbuild/copr_rpmbuild/helpers.py
praiskup: resolved on slack

rpmbuild/copr_rpmbuild/config.py
praiskup: marking as resolved (I'm fine if we agree on disagreement)

praiskup:
Fun -> I thought we had broken Fedora Copr!  But the testsuite hangs on `tests/test_base.py::TestProvider::test_create_rpmmacros` for some reason (and eats more and more memory till it fails).

praiskup (approved):

PR #3413, opened 2024-09-18
beaker-tests-sanity: run basic tests for every supported storage

FrostyX:
I wanted a simple test to make sure all of the supported storages work regardless of what storage is configured as default. The coverage is very low and limited only to the most basic workflow (creating a project, building a package, and installing it). That's good enough to know that every storage is configured correctly. I don't want to add any more tests to this script. In case we want a thorough testing of a specific storage, we should change the default storage on copr-fe-dev

    # What storage should be set for new projects.
    # Possible options are "backend" and "pulp"
    DEFAULT_STORAGE = "backend"

and then run the full beaker tests suite.


FrostyX:
Removing the blocked label now. The tests should work since #3383 is merged.

nikromen (approved):

nikromen:
let's try

PR #3419, opened 2024-09-23
backend: unknown resalloc tickets helper cleanup

praiskup:
If no tickets are taken (which often happens in the staging environment), this script encountered corner case issues.

### DIFF
--- a/backend/run/copr-backend-unknown-resalloc-tickets.py
+++ b/backend/run/copr-backend-unknown-resalloc-tickets.py
@@ -22,7 +22,8 @@ from copr_common.helpers import script_requires_user
 
 def used_ids():
     """
-    Return a set of ticket_ids that can be found in the "ps aux" output.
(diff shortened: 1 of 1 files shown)


backend/run/copr-backend-unknown-resalloc-tickets.py
```diff
@@ -33,7 +33,9 @@ def used_ids():
     output = output.strip()
     tickets = set()
     for ticket_id in output.split("\n"):
-        assert all(c.isdigit() for c in ticket_id)
+        if len(ticket_id) <= 0:
```
FrostyX, line 36: This needs an explanation. A length of a string ticket ID, is a weird thing to calculate. Are you basically doing `if ticket_id:`?

praiskup: yes, updated

FrostyX (approved):

PR #3420, opened 2024-09-23
rpmbuild: do not require qemu-user-static on rhel

xsuchy:
Resolves: RHBZ#2313879

### DIFF
--- a/rpmbuild/copr-rpmbuild.spec
+++ b/rpmbuild/copr-rpmbuild.spec
@@ -63,7 +63,9 @@ Requires: git
 Requires: git-svn
 # for the /bin/unbuffer binary
 Requires: expect
-%if !0%{?openEuler}
+%if 0%{?openEuler} > 0 || 0%{?rhel} > 0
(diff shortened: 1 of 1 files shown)


xsuchy:
Added commit that resolves https://bugzilla.redhat.com/show_bug.cgi?id=2313878 too.

rpmbuild/copr-rpmbuild.spec
```diff
@@ -104,7 +106,9 @@ Requires: nosync
 %endif
 Requires: openssh-clients
 Requires: podman
-%if !0%{?openEuler}
+%if 0%{?openEuler} > 0 || 0%{?rhel} > 0
+# not supported
+%else
```
praiskup, line 111: This means that we make copr-rpmbuild unusable on RHEL at all? (=not just emulated builds).

praiskup: ok, I see thre are two commits

praiskup (approved):

PR #3428, opened 2024-09-26
Fix traceback caused by None timeout

FrostyX:
Running `runtest-custom-method.sh`, we are seeing this failure in backend.log:

    Unexpected exception (in /usr/lib/python3.12/site-packages/copr_backend/background_worker_build.py:431)

The problematic line is

    timestamp = self.job.started_on + self.job.timeout + 60 * 60

and the hidden exception is

    TypeError: unsupported operand type(s) for +: 'float' and 'NoneType'

The `self.job.timeout` is taken from the frontend task, if it is set. The problem is when frontend sets it to `None`.

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -790,8 +790,10 @@ class BuildsLogic(object):
 
         if "timeout" in build_options:
             build.timeout = build_options["timeout"]
+        elif package and package.timeout:
+            build.timeout = package.timeout
         else:
-            build.timeout = package.timeout if package else app.config["DEFAULT_BUILD_TIMEOUT"]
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -790,8 +790,10 @@ def create_new(cls, user, copr, source_type, source_json, chroot_names=None, pkg
 
         if "timeout" in build_options:
             build.timeout = build_options["timeout"]
+        elif package and package.timeout:
+            build.timeout = package.timeout
         else:
-            build.timeout = package.timeout if package else app.config["DEFAULT_BUILD_TIMEOUT"]
+            build.timeout = app.config["DEFAULT_BUILD_TIMEOUT"]
```
praiskup, line 796: Nice example for #534.

praiskup (approved):
Thank you.

PR #3431, opened 2024-09-27
beaker: update fedora version to f41 for testing

nikromen:
-> runtest-modules.sh used old f38 already EOLed chroots


FrostyX (approved):

nikromen:
you are too fast, I found out that f41 won't work with dnf5 currently :D so I'll go with f40

PR #3426, opened 2024-09-26
Activate Red Hat subscription on demand

FrostyX:
Fix #2132

PR #3434, opened 2024-09-30
rpmbuild: add copr-builder-rhsm-subscribe script

praiskup:
This allows us to specify RHSM password so it is not visible on `ps aux` output

### DIFF
--- a/rpmbuild/bin/copr-builder-rhsm-subscribe
+++ b/rpmbuild/bin/copr-builder-rhsm-subscribe
@@ -1,0 +1,48 @@
+#! /usr/bin/python3
+
+"""
+Run `subscription-manager register` without having the --activation key in
+/proc/self/environ or /proc/self/comm.
+"""
+
+import argparse
+import getpass
+import sys
+
+from subscription_manager.scripts.subscription_manager import main as rhsm
+
+
+def read_key():
+    """
+    Read key from stdin if not a tty.  Never export as env-var.
+    """
+    if sys.stdin.isatty():
+        key = getpass.getpass("rhsm key: ")
+    else:
+        key = sys.stdin.read()
+    return key.strip()
+
+
+def _arg_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--org-id", required=True)
+    parser.add_argument("--system-name", required=True)
+    return parser
+
+
+def _main():
+    opts = _arg_parser().parse_args()
+    a_key = read_key()
+    sys.argv = [
+        "subscription-manager", "register", "--force",
+        "--org", opts.org_id,
+        "--name", opts.system_name,
+        "--activationkey", a_key,
+    ]
+    # Call subscription-manager's internal method to avoid polluting the
+    # /proc/self/cmdline with the key credential!
+    rhsm()
+
+
+if __name__ == "__main__":
+    _main()
--- a/rpmbuild/bin/copr-builder-rhsm-subscribe-daemon
+++ b/rpmbuild/bin/copr-builder-rhsm-subscribe-daemon
@@ -1,0 +1,44 @@
+#! /bin/bash
+
+# Try to run subscription-manager register repeatedly (till it succeeds),
+# without polluting environment variables or command-line options with
+# passwords/keys.
+
+if test $# -ne 2; then
+cat <<EOHELP >&2
+Usage: $0 ORG_ID SYSTEM_NAME <<<"\$password"
+
+Provide the activation_key on stdin!
+EOHELP
+exit 1
+fi
+
+opt_org_id=$1
+opt_system=$2
+
+try_indefinitely()
+{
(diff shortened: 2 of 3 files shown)


rpmbuild/bin/copr-builder-rhsm-subscribe
```diff
@@ -0,0 +1,13 @@
+#! /bin/expect
+
+# Provide username+password for subscription-manager register
+# through RHSM_USER and RHSM_PASS.
+# https://bugzilla.redhat.com/show_bug.cgi?id=2315673
+
+set timeout 30
```
praiskup, line 7: or more?

rpmbuild/bin/copr-builder-rhsm-subscribe
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/expect
+
+# Provide username+password for subscription-manager register
+# through RHSM_USER and RHSM_PASS.
+# https://bugzilla.redhat.com/show_bug.cgi?id=2315673
+
+set timeout 30
+set USER $env(RHSM_USER)
+set PASS $env(RHSM_PASS)
```
praiskup, line 9: Unfortunately, env vars are not safe either as long as user can ssh to the box:
https://github.com/fedora-copr/copr/blob/35a23e6c655967c499492dabbef10abce2fde57e/backend/copr_backend/background_worker_build.py#L993-L994

rpmbuild/bin/copr-builder-rhsm-subscribe-daemon
```diff
@@ -0,0 +1,93 @@
...
+}
+
+fail=false
+for env_required in RHSM_PASS RHSM_USER; do
+    eval 'test -z "$'$env_required'"' && echo >&2 "ENV ${env_required} required" && fail=true
+done
+$fail && exit 1
+
+try_indefinitely copr-builder-rhsm-subscribe
+try_indefinitely subscription-manager attach --pool "$opt_pool_id"
```
FrostyX, line 93: Can you please touch some `/var/run` file  here? Please see https://github.com/fedora-copr/copr/pull/3426#discussion_r1791207717

praiskup:
Some new observations:
- the activation keys have a much smaller identity scope compared to passwords (we are not risking the loss of our identity, just that we give someone rights to enable new systems)
- the key is read from stdin so it is neither in /proc/self/comm nor in /proc/self/environ
- the daemon process that has the password in memory disappears relatively quickly, before we let the user in over ssh, unless there's a problem with RHSM
- because problems with RHSM happen "from time to time", we should count that /proc/self/mem analysis might lead to some leaks, so we should rotate the key (I doubt there's a huge motivation to do this, because getting RH Developer license is much easier).

rpmbuild/copr-rpmbuild.spec
```diff
@@ -60,7 +61,9 @@ Requires: %{python_pfx}-specfile >= 0.21.0
 Requires: python3-backoff >= 1.9.0
 Requires: python3-pyyaml
 
+Requires: expect
```
FrostyX, line 64: A leftover dependency on `expect`

rpmbuild/bin/copr-builder-rhsm-subscribe
```diff
@@ -0,0 +1,46 @@
...
+"""
+Run `subscription-manager register` without having the --activation key in
+/proc/self/environ or /proc/self/comm.
+"""
+
+import argparse
+import getpass
+import sys
+
+from subscription_manager.scripts.subscription_manager import main as rhsm
```
FrostyX, line 12: I am a bit afraid of them changing the location or the internals of the script and I would probably rather call it through `subprocess` but we can do this.

FrostyX:
This PR is probably blocked by the switch of our account to SCA, right?

rpmbuild/bin/copr-builder-rhsm-subscribe
praiskup: This is sub-optimal, but subprocess.call means the key goes to /proc/self/cmdline

FrostyX: Can you please add a code comment about that so that I don't get the great idea to re-implement it someday? :-)

rpmbuild/copr-rpmbuild.spec
praiskup: dropped, hehe, I realized expect already was in requires

praiskup:
> This PR is probably blocked by the switch of our account to SCA, right?

Not really, this can stay unused till we switch.

rpmbuild/bin/copr-builder-rhsm-subscribe
praiskup: done

FrostyX (approved):

PR #3436, opened 2024-09-30
beaker: deprecate vera package, use enum

nikromen:
Vera package is unmaintained for a while and it won't build (for whatever complicated reason). Deprecate vera in favor of enum which is kind of taken care of by us via rpm-package-workshop.


FrostyX:
`vera.spec` is also used here at `beaker-tests/Sanity/copr-cli-basic-operations/runtest-enable-net.sh:44`

FrostyX:
FTR the "whatever complicated reason" is:

> No match for argument: python2-devel

nikromen:
python2-devel wasn't the problem inside the build though: https://download.copr-dev.fedorainfracloud.org/results/@copr/TEST1727691521488417474BuildSpec/fedora-40-x86_64/02919286-vera/builder-live.log.gz

FrostyX (approved):

PR #3456, opened 2024-10-06
Custom Webhook now adds a record in Webhook History table

### LINKED ISSUE #3455: No webhook history for @copr/copr-pull-requests
There are no webhooks displayed for our project 
https://copr.fedorainfracloud.org/coprs/g/copr/copr-pull-requests/integrations/
even though we webhook-rebuilt a few builds since the last Copr release.

```
Webhook History
No webhook received recently.
```

I am curious what we missed @jaitjacob. I'll let you know once I have some debugging information.

jaitjacob:
Custom Webhooks are not expected to send a UUID. Instead the UI checks for `None` passed for Webhook UUID and prints "Custom Webhook" instead. See below, 
image

Fix #3455

### DIFF
--- a/frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
+++ b/frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
@@ -75,15 +75,21 @@
                 <thead>
                   <tr>
                     <th>Timstamp</th>
-                    <th>Webhook ID</th>
+                    <th>Forge-generated Webhook UUID</th>
                     <th>User Agent</th>
                   </tr>
                 </thead>
                 {% for webhook in webhook_history %}
                 <tr>
-                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}})</td>
-                    <td>{{webhook.webhook_uuid}} </td>
-                    <td>{{webhook.user_agent}}</td>
+                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}} ago)</td>
+                    <td>{% if webhook.webhook_uuid is none %} No UUID received
+                        {% else %}{{ webhook.webhook_uuid }}
+                        {% endif %}
+                   </td>
+                    <td>{% if webhook.user_agent is none %} Not Set 
+                        {% else %} {{webhook.user_agent}}
+                        {% endif %}
+                    </td>
                 </tr>
                 {% endfor %}
             </table>
--- a/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
+++ b/frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
@@ -82,7 +82,7 @@ def package_name_required(route):
     return decorated_function
 
 
-def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
+def add_webhook_history_record(webhook_uuid, user_agent=None,
                                builds_initiated_via_hook=None):
     """
     This methods adds info of an intercepted webhook to webhook_history db
@@ -92,14 +92,16 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
         log.debug("No build initiated. Webhook not logged to db.")
         return
 
-    webhookRecord = models.WebhookHistory(created_on=int(time.time()),
+    webhook_record = models.WebhookHistory(created_on=int(time.time()),
                                           webhook_uuid=webhook_uuid,
                                           user_agent=user_agent)
-    db.session.add(webhookRecord)
-    db.session.commit()
+    db.session.add(webhook_record)
+
+    if not isinstance(builds_initiated_via_hook, list):
+        builds_initiated_via_hook = [builds_initiated_via_hook]
 
     for build in builds_initiated_via_hook:
-        build.webhook_history_id = webhookRecord.id
+        build.webhook_history = webhook_record
 
 
 @webhooks_ns.route("/bitbucket/<int:copr_id>/<uuid>/", methods=["POST"])
@@ -346,6 +348,9 @@ def custom_build_submit(copr, package, copr_dir=None):
         log.exception('can not submit build from webhook')
         storage.delete()
         return "BUILD_REQUEST_ERROR\n", 500
+
+    user_agent = flask.request.headers.get('User-Agent')
+    add_webhook_history_record(None, user_agent, build)
 
     # Return the build ID, so (e.g.) the CI process (e.g. Travis job) knows
     # what build results to wait for.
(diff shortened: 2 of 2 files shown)


jaitjacob:
@FrostyX  You may notice redundant `db.session.commit()` in `webhook_general.py`. I couldn't avoid them cause until we explicitly commit, 
1.`id` is not generated and that results in an error
2. Webhook History table does not update with the latest webhook info

frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
```diff
@@ -82,7 +82,11 @@ <h3>Webhook History</h3>
                 {% for webhook in webhook_history %}
                 <tr>
                     <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}})</td>
-                    <td>{{webhook.webhook_uuid}} </td>
+                    <td>{% if webhook.webhook_uuid is none %}
+                        {{"Custom Webhook"}}
```
praiskup, line 86: This goes into the "Webhook ID", I'd probably prefer something like "No UUID generated for this webhook call.".

The double-curly-brackets are not needed.

Also, can you change the header like s/Webhook ID/Forge-generated webhook UUID/g?

praiskup:
Thank you very much for the quick turn-around.  There's no need to send those simple fixes as drafts ;-)

frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
praiskup: Yeah, I've seen people triggering the GitHub/GitLab webhooks manually.  I did that a few times for debugging purposes:
```
curl -X POST -H "X-GitHub-Event: ..." -H "Content-Type: application/json" --data "$(cat some json)" ...
```

And also in the future, we might add webhooks that will not have forge-generated UUIDs (not just custom hooks).

jaitjacob: done
image

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
```diff
@@ -347,6 +351,9 @@ def custom_build_submit(copr, package, copr_dir=None):
         storage.delete()
         return "BUILD_REQUEST_ERROR\n", 500
 
+    user_agent = flask.request.headers.get('User-Agent')
```
nikromen, line 354: this can be `None`. We need to either add here check `if user_agent is None then user_agent = "Not Set"` or maybe better do this check inside add_webhook_history_record and s/user_agent="Not Set"/user_agent=None/

jaitjacob: done. added a check inside `add_webhook_history_record`

nikromen (approved):
:rocket:

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
```diff
@@ -98,8 +101,12 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
     db.session.add(webhookRecord)
     db.session.commit()
 
+    if not isinstance(builds_initiated_via_hook, list):
+        builds_initiated_via_hook = [builds_initiated_via_hook]
+
     for build in builds_initiated_via_hook:
         build.webhook_history_id = webhookRecord.id
+        db.session.commit()
```
praiskup, line 109: you can avoid this commit() actually;  you just have to avoid working with IDs directly, but use the references..

praiskup: example https://github.com/fedora-copr/copr/blob/916f82f059610c2af320d01b1893d59ffd27c725/frontend/coprs_frontend/coprs/logic/coprs_logic.py#L333

jaitjacob: - nice I dont need to use any `commit()` in this method after using refs. 
- renamed `webhookRecord` to `webhook_record` (JS instincts may have kicked in :D )

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
```diff
@@ -88,18 +88,23 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
     This methods adds info of an intercepted webhook to webhook_history db
     along with the initiated build number(s).
     """
+    if user_agent is None:
+        user_agent = 'Not Set'
```
praiskup, line 92: Shouldn't we store `None` to database, instead of `Not set` string?  That would seem nicer, we may decide the default "presentation" string in HTML templates.  Also, what if clients fake the agent to "Not set" :-)

praiskup:
Thank you very much for the update; I'd just store NULLs for the user agent if not known, otherwise LGTM

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
jaitjacob: done. we now store it in DB as `None` and present it in the template as 'Not Set'.

frontend/coprs_frontend/coprs/templates/coprs/detail/settings/integrations.html
```diff
@@ -75,15 +75,21 @@ <h3>Webhook History</h3>
...
                 <tr>
-                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}})</td>
-                    <td>{{webhook.webhook_uuid}} </td>
-                    <td>{{webhook.user_agent}}</td>
+                    <td  class="webhook_timestamp" data-toggle="tooltip" title="{{webhook.created_on|localized_time(g.user.timezone)}}">{{webhook.created_on|localized_time(g.user.timezone)}} ({{webhook.created_on|time_ago()}} ago)</td>
+                    <td>{% if webhook.webhook_uuid is none %} No UUID received
+                        {% else %}{{ webhook.webhook_uuid }}
+                        {% endif %}
+                   </td>
+                    <td>{% if webhook.user_agent is none %} Not Set
```
praiskup, line 89: :+1:

frontend/coprs_frontend/coprs/views/webhooks_ns/webhooks_general.py
```diff
@@ -92,14 +92,16 @@ def add_webhook_history_record(webhook_uuid, user_agent='Not Set',
...
-    webhookRecord = models.WebhookHistory(created_on=int(time.time()),
+    webhook_record = models.WebhookHistory(created_on=int(time.time()),
                                           webhook_uuid=webhook_uuid,
                                           user_agent=user_agent)
-    db.session.add(webhookRecord)
-    db.session.commit()
+    db.session.add(webhook_record)
+
+    if not isinstance(builds_initiated_via_hook, list):
+        builds_initiated_via_hook = [builds_initiated_via_hook]
```
praiskup, line 101: Ah, I missed this before.  Can we simply expect just a list input?  This check is not complete and ideal (what if None comes in), and it makes the "method API" complicated.

jaitjacob: fixed ty :)

praiskup (approved):
Thank you!  Note #3480, but that might be resolved later.

PR #3463, opened 2024-10-07
beaker: migrate to f41

### LINKED ISSUE #3462: beaker: f41+ containers can't start tests
First steps for migration were done in #3463

copr-dev hasn't f41 chroot enabled, thus beaker inside f41 fails:

```
Something went wrong:
Error: chroots: 'fedora-41-x86_64' is not a valid choice for this field.
:: [ 10:09:11 ] :: [   FAIL   ] :: Command 'copr-cli create --chroot fedora-41-x86_64 --disable_createrepo false @copr/TEST1728295748689287368DisableCreaterepoFalse' (Expected 0, got 1)
```

the same is expected for internal copr. Idk what is needed next to make tests work, but I am 100% sure module tests will be broken.

nikromen:
Fixes #3462

### DIFF
--- a/beaker-tests/DockerTestEnv/Dockerfile
+++ b/beaker-tests/DockerTestEnv/Dockerfile
@@ -7,7 +7,7 @@ RUN dnf -y update \
     && dnf -y install dnf-plugins-core wget \
     && dnf -y copr enable @copr/copr \
     && if test "$COPR_PACKAGES" = devel; then dnf -y copr enable @copr/copr-dev; fi \
-    && dnf -y install hostname htop net-tools iputils vim mlocate git sudo \
+    && dnf -y install hostname htop net-tools iputils vim git sudo \
               openssh-server psmisc python-jedi procps-ng findutils tmux \
               expect python3-behave python3-hamcrest \
     && dnf -y install python3-copr rpm-build copr-cli jq \
@@ -15,7 +15,7 @@ RUN dnf -y update \
     && dnf -y clean all
 
 RUN for repo in fedora updates updates-testing; do \
-    dnf -y config-manager --save --setopt $repo.excludepkgs=hello ; done
+    dnf -y config-manager setopt $repo.excludepkgs=hello ; done
 
 RUN echo "LANG=en_US.UTF-8" >> /etc/locale
 RUN setcap cap_net_raw,cap_net_admin+p /usr/bin/ping
--- a/beaker-tests/DockerTestEnv/Makefile
+++ b/beaker-tests/DockerTestEnv/Makefile
@@ -13,7 +13,7 @@ bld: build
 run:
 	$(CONTAINER_ENGINE) run \
 		-dit \
-		-v $(mkfile_path)/../..:/root/copr \
+		-v $(mkfile_path)/../..:/root/copr:z \
 		--name="$(CONTAINER_NAME)" \
 		-h $(CONTAINER_NAME) \
 		-w /root/ \
--- a/beaker-tests/Sanity/copr-cli-basic-operations/runtest-dnf-copr.sh
+++ b/beaker-tests/Sanity/copr-cli-basic-operations/runtest-dnf-copr.sh
@@ -46,24 +46,26 @@ rlJournalStart
         OUTPUT=`mktemp`
 
         # Test `dnf copr search' command
+        # https://github.com/fedora-copr/copr/issues/3531
         rlRun "copr-cli create --chroot $CHROOT ${NAME_PREFIX}DnfCopr"
-        rlRun "dnf copr --hub tested-copr search ${NAME_PREFIX}DnfCopr > $OUTPUT"
+        rlRun "dnf-3 copr --hub tested-copr search ${NAME_PREFIX}DnfCopr > $OUTPUT"
         rlRun "cat $OUTPUT |grep '${NAME_PREFIX}DnfCopr' |grep 'No description given'"
 
         # Test `dnf copr list' command
         rlRun "dnf copr list > $OUTPUT"
-        rlRun "cat $OUTPUT |grep 'copr.fedorainfracloud.org/group_copr/copr'"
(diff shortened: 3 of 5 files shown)


beaker-tests/DockerTestEnv/Makefile
```diff
@@ -13,7 +13,7 @@ bld: build
 run:
 	$(CONTAINER_ENGINE) run \
 		-dit \
-		-v $(mkfile_path)/../..:/root/copr \
+		-v $(mkfile_path)/../..:/root/copr:Z \
```
FrostyX, line 16: Would `:z` work too? I generally try to avoid `:Z` because the bindmount cannot be accessed from other containers. Shouldn't be a problem here but still

nikromen: That was the reason why I put `:Z` here to pretend at least I care about minimal privileges since I know we don't want to access the bind-mount from other containers. But yeah, it doesn;t mater in this use case so `:z` would work too.

beaker-tests/DockerTestEnv/Dockerfile
```diff
@@ -7,15 +7,15 @@ RUN dnf -y update \
     && dnf -y install dnf-plugins-core wget \
     && dnf -y copr enable @copr/copr \
     && if test "$COPR_PACKAGES" = devel; then dnf -y copr enable @copr/copr-dev; fi \
-    && dnf -y install hostname htop net-tools iputils vim mlocate git sudo \
+    && dnf -y install hostname htop net-tools iputils vim plocate git sudo \
```
nikromen, line 10: do we need plocate?

FrostyX:
Is there any other blocker or only the `:z` and `plocate` things needs to be addressed?

beaker-tests/Sanity/copr-cli-basic-operations/runtest-modules.sh
```diff
@@ -99,6 +99,7 @@ rlJournalStart
 
         # Test yaml submit
         PROJECT=module-testmodule-beakertest-$DATE
+        echo "copr-cli create $PROJECT --chroot $CHROOT --chroot fedora-rawhide-i386"
```
FrostyX, line 102: This seems to be a leftover from debugging

beaker-tests/Sanity/copr-cli-basic-operations/runtest-modules.sh
```diff
@@ -85,7 +85,7 @@ function test_successful_packages()
 rlJournalStart
     rlPhaseStartSetup
         setup_checks
-        yum -y install dnf dnf-plugins-core
+        yum -y install dnf dnf-3 dnf-plugins-core
```
praiskup, line 88: I never seen `dnf-3` :D interesting.

Having `dnf install dnf4 dnf3` would be slightly more obvios.

beaker-tests/Sanity/copr-cli-basic-operations/runtest-dnf-copr.sh
```diff
@@ -46,24 +46,26 @@ rlJournalStart
...
         rlRun "dnf -y copr enable --hub tested-copr ${NAME_PREFIX}DnfCopr"
         rlRun "dnf -y install hello"
         rlRun "hello"
-        rlRun "dnf -y erase hello"
+        rlRun "dnf -y remove hello"
 
         # Test `dnf copr list --available-by-user' command
-        rlRun "dnf copr --hub tested-copr list --available-by-user $OWNER > $OUTPUT"
+        # TODO Implement the --available-by-user' parameter in DNF5
+        rlRun "dnf-3 copr --hub tested-copr list --available-by-user $OWNER > $OUTPUT"
```
praiskup, line 68: Do we really need that?  I never used this.

praiskup: I mean, "do we need to implement this in DNF5?"

beaker-tests/Sanity/copr-cli-basic-operations/runtest-dnf-copr.sh
```diff
@@ -46,24 +46,26 @@ rlJournalStart
         OUTPUT=`mktemp`
 
         # Test `dnf copr search' command
+        # TODO Implement the search command in DNF5
```
praiskup, line 49: Again, do we need this in DNF5? If yes, open an issue, if not - let's just leave it with comment that we don't plan to implement it.

FrostyX: I don't know if we necessarily need it, but in comparison to the not-very-useful `--available-by-user` parameter, the search command could be nice to have. I will create a ticket for that. Do you prefer it in the DNF or Copr issue tracker?

beaker-tests/Sanity/copr-cli-basic-operations/runtest-modules.sh
nikromen: on fedora 41:

```
[root@10a47da1ac97 /]# dnf install dnf3
Updating and loading repositories:
Repositories loaded.
No match for argument: dnf3
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
[root@10a47da1ac97 /]#
```

nikromen: s/dnf-3/dnf4

nikromen:
I pushed some fixes, the tests passes for modules and I started all tests against dev, but I think with some changes in the main branch - some of them will probably fail unrelated to this. @FrostyX you tried to fix some problems not related to modules but still related to dnf4, do you think that your changes are enough then?

And I also left the TODOs as they are for you to decide whether to drop them or create issue about them - I don't have strong opinion on that.

FrostyX:
I updated the TODOs, and I am going to merge now. Anything else can be done in a follow-up PR.

Issue #2597, opened 2023-03-28
Build failed in a subdir with make-srpm mothed


pkking:
[logs here](https://download.copr.fedorainfracloud.org/results/lcrpkking/waaagh/srpm-builds/05715548/builder-live.log.gz)
Just for record:
```
Task:
{'appstream': True,
 'background': False,
 'build_id': 5715548,
 'chroot': None,
 'project_dirname': 'waaagh',
 'project_name': 'waaagh',
 'project_owner': 'lcrpkking',
 'repos': [],
 'sandbox': 'lcrpkking/waaagh--lcrpkking',
 'source_json': {'clone_url': 'https://github.com/pkking/copr.git',
                 'committish': '',
                 'spec': 'copr-frontend.spec',
                 'srpm_build_method': 'make_srpm',
                 'subdirectory': 'frontend',
                 'type': 'git'},
 'source_type': 8,
 'submitter': 'lcrpkking',
 'task_id': '5715548'}

Running: git clone https://github.com/pkking/copr.git /var/lib/copr-rpmbuild/workspace/workdir-0ksv7tdd/copr --depth 500 --no-single-branch --recursive

cmd: ['git', 'clone', 'https://github.com/pkking/copr.git', '/var/lib/copr-rpmbuild/workspace/workdir-0ksv7tdd/copr', '--depth', '500', '--no-single-branch', '--recursive']
cwd: .
rc: 0
stdout: 
stderr: Cloning into '/var/lib/copr-rpmbuild/workspace/workdir-0ksv7tdd/copr'...

Running: mock --uniqueext 1679972862.252171 -r /var/lib/copr-rpmbuild/workspace/safe-resultdir-yi7elwsi/mock-source-build.cfg '--plugin-option=bind_mount:dirs=(("/var/lib/copr-rpmbuild/workspace/workdir-0ksv7tdd", "/mnt/workdir-0ksv7tdd"), ("/var/lib/copr-rpmbuild/workspace/safe-resultdir-yi7elwsi", "/mnt/safe-resultdir-yi7elwsi"))' --chroot 'set -x && cd /mnt/workdir-0ksv7tdd/copr/frontend && echo -e "[safe]
[...]

PR #3464, opened 2024-10-07
rpmbuild: make_srpmbuild, set recursive safe.directory

### LINKED ISSUE #3421: The make_srpm method fails with git safe.directory error
Full log: [builder-live.log.gz.txt](https://github.com/user-attachments/files/17113031/builder-live.log.gz.txt)

```
./contrib/fedora/make_srpm.sh --prerelease --output /mnt/safe-resultdir-c3dy38x1
remote: fatal: detected dubious ownership in repository at '/mnt/workdir-a7gx95j6/sssd/.git'        
remote: To add an exception for this directory, call:        
remote: 
remote: 	git config --global --add safe.directory /mnt/workdir-a7gx95j6/sssd/.git        
remote: git upload-archive: archiver died with error
fatal: sent error to the client: git upload-archive: archiver died with error
```

But strangely enough, after the fatal error, we still produce the SRPM?

```
Creating SRPM without extra patches.
setting SOURCE_DATE_EPOCH=1611187200
Wrote: /mnt/workdir-a7gx95j6/sssd/rpmbuild/SRPMS/sssd-2.10.0~beta2-0.240924.083430.git5f5077ac1.src.rpm
```

nikromen:
Relates #3421

### DIFF
--- a/rpmbuild/copr_rpmbuild/providers/scm.py
+++ b/rpmbuild/copr_rpmbuild/providers/scm.py
@@ -17,7 +17,7 @@ log = logging.getLogger("__main__")
 MAKE_SRPM_TEPMLATE = (
     "set -x && "
     'cd {0} && '
-    'echo -e "[safe]\ndirectory = {4}" > ~/.gitconfig && '
+    'echo -e "[safe]\ndirectory = {4}\ndirectory = {4}/*" > ~/.gitconfig && '
     'make -f {1} srpm outdir="{2}" spec="{3}"'
 )
 


rpmbuild/copr_rpmbuild/providers/scm.py
```diff
@@ -17,7 +17,7 @@
 MAKE_SRPM_TEPMLATE = (
     "set -x && "
     'cd {0} && '
-    'echo -e "[safe]\ndirectory = {4}" > ~/.gitconfig && '
+    'echo -e "[safe]\ndirectory = {4}/*" > ~/.gitconfig && '
```
praiskup, line 20: +1

rpmbuild/copr_rpmbuild/providers/scm.py
```diff
@@ -118,7 +118,7 @@ def get_make_srpm_command(self):
 
         makefile_path = os.path.join(mock_repodir, '.copr', 'Makefile')
         make_srpm_cmd_part = MAKE_SRPM_TEPMLATE.format(mock_cwd, makefile_path,
-                mock_resultdir, mock_spec_path, mock_repodir)
+                mock_resultdir, mock_spec_path, mock_workdir)
```
praiskup, line 121: This is weird though; the mock_repodir should be the directory containing `.git` subdir, right?

nikromen: that's just remnant to make it work. Actually I want mock_repodir and do:
```
[safe]
  directory = {4}
  directory = {4}/*
```

nikromen:
@praiskup OK this was reason why we moved repo root as safe directory - https://github.com/fedora-copr/copr/issues/2597#issuecomment-1487901896 so I don't think we should revert this change. We probably want the `/*` feature

praiskup:
I still don't get why :sweat_smile: I experimented with permissions like:

    # I'm "praiskup"
    $ cd /tmp
    $ git clone ~/rh/projects/copr/copr
    $ cd /tmp/copr
    # sudo su
    # chown postgres:postgres cli
    # git status
    On branch main
    Your branch is up to date with 'origin/main'.

    nothing to commit, working tree clean
    # chown postgres:postgres /tmp/copr
    # git status
    fatal: detected dubious ownership in repository at '/tmp/copr'
    To add an exception for this directory, call:

            git config --global --add safe.directory /tmp/cop

IOW, there's no issue with the git tree sub-directory.  Only if the top-level
directory is mis-owned.  But I agree we don't want to revert the change;
the problem there clearly fixes the problem if subdirectory is used as
`cwd`.

praiskup:
We talked about this PR on f2f meeting, and while the `<dir>` to `<dir>/*` change is not necessarily needed now (a different fix was needed), some users in the future might consider it useful (they might be doing manual git clone in sub-dirs, and there's no need to bother them with tweaking the safe.directory configuration).

praiskup:
Jirka promised to update the commit message.

nikromen:
right, I forgot about that, sorry

nikromen:
done

praiskup (approved):
Seems OK, we talk about gitconfig file **in a mock chroot**. Thank you!

PR #3465, opened 2024-10-07
backend: fix copr-backend-unknown-resalloc-tickets

praiskup:
And add a test-case now.


backend/tests/test_unknown_resalloc_tickets.py
```diff
@@ -0,0 +1,57 @@
...
+        with contextlib.redirect_stdout(f):
+            SCRIPT._main()  # pylint: disable=protected-access
+            return f.getvalue()
+
+
+@patch_path_run
+@patch_path_fake_executables("unknown_tickets")
+@patch_getpwuid("resalloc")
+def test_unknown_tickets():
+    subprocess.call("ps")
```
FrostyX, line 34: Would it be the same to just read the `ps` file and print it to STDOUT? So that we don't have to run a script.

praiskup: This one is a noop and useless debugging thingy - going to drop.  `ps` is executed by the `SCRIPT` itself.

FrostyX:
There is an empty file called `backend/fake_executables`. Is that intentional? If yes, can you please put some docstring to it?

backend/tests/test_unknown_resalloc_tickets.py
FrostyX: Yeees, that makes much more sense now. Thank you.
(We fake the `ps` binary, provide PATH to the faked binary, and the tested script has no idea it was mocked)

praiskup:
`backend/fake_executables` dropped

FrostyX (approved):

PR #3467, opened 2024-10-08
Store payload in Webhook History & make webhook ID clickable to show corresponding payload

### LINKED ISSUE #634: Provide webhook payload for triggered builds
Original issue: https://pagure.io/copr/copr/issue/634
Opened: 2019-04-03 20:14:03
Opened by: frostyx

jaitjacob:
In this PR, 

1. Webhook Payload is stored in Webhook History table. 
2. The Webhook ID is rendered as a link to open the payload in new tab. The JSON data is passed as a HTML Data Attribute along with the markup.
image
3. Payload opened in new tab using javascript function `showJson`,
image

Reference,
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

Fixed #634

### DIFF
--- a/frontend/coprs_frontend/alembic/versions/046fb5ed2cf3_add_new_column_to_store_webhook_payload.py
+++ b/frontend/coprs_frontend/alembic/versions/046fb5ed2cf3_add_new_column_to_store_webhook_payload.py
@@ -1,0 +1,24 @@
+"""
+add new column to store Webhook payload
+
+Revision ID: 046fb5ed2cf3
+Create Date: 2024-10-07 05:53:50.891339
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '046fb5ed2cf3'
+down_revision = 'bb52d9f878f5'
+branch_labels = None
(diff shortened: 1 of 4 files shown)


praiskup:
Thank you very much for your work on this!

Team: Could someone start a discussion on the Fedora Devel mailing list
regarding whether it's considered "safe enough" to make the payload publicly
available?  My initial feeling is that payloads from Forges **should** be safe
to publish, yet Forges don't typically do so.  It would be helpful to get input
from the involved parties and let them help us decide.

praiskup:
> They don't allow user+password authentication and require a token. Then curl or gh can be used. But this is more setup steps than I would like

If we provide documentation on how to do the boilerplate setup (token?), and the table self-links this documentation, I don't think it is toooo bad. :+1:

jaitjacob:
sorry marked it ready for review by mistake. please ignore.

praiskup:
@jaitjacob hello, how is this PR going?  Do you think we could document the way(s) how to "to get the payload"?

xsuchy:
Closing due to inactivity. Please reopen if you want to continue.

Issue #3471, opened 2024-10-10
Packit 400 errors


praiskup:
Follow up for #3372 and #3443.  Now we get error 400 sometimes.

praiskup:
Hot-fixed in production.

PR #3472, opened 2024-10-10
frontend: avoid a duplicated "exists" check

praiskup:
The commit 609d36988fb307193c77c872b726323bfca0647b only partly resolved the TOCTOU effects (which we are not fixing here either).  The problem being fixed now is that we duplicated the duplication check,
- once in the `validate_on_submit()`, then
- using the `add(check_for_duplicates=True)`. The later check is wrong though, because it raises an exception that is not handled properly (exist_ok=True ignored).

Complements: fdff0c826fe8ffa407f50d4006e617961a6fe046
Complements: 609d36988fb307193c77c872b726323bfca0647b
Fixes: #3471

### DIFF
--- a/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
+++ b/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
@@ -242,7 +242,6 @@ class ProjectAdd(Resource):
                 selected_chroots=form.selected_chroots,
                 description=form.description.data,
                 instructions=form.instructions.data,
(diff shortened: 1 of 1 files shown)


praiskup:
I mean, this frontend's code deserves a major cleanup IMVHO - namely I think this commit brings in some dead code.  But the commit here seems relatively simple so we could do more things in a separate PR.

FrostyX (approved):

FrostyX:
> @FrostyX do you think I can hot-fix this in prod?

We discussed it off-list but just FTR, I ACKed this

PR #3473, opened 2024-10-10
Change the pyp2spec invocation

befeleme:
pyp2spec 0.10.0 changed the API around the specfile generation. For fully-automated environments as copr is, the automode newly includes convenient defaults to make the specfiles as close to buildability as possible.

FYI: pyp2spec 0.10.0 is currently in bodhi, so during the next 7 days it'll get to all active Fedoras.

### DIFF
--- a/rpmbuild/copr_rpmbuild/providers/pypi.py
+++ b/rpmbuild/copr_rpmbuild/providers/pypi.py
@@ -56,7 +56,7 @@ class PyPIProvider(Provider):
             "pyp2spec",
             self.pypi_package_name,
             "--fedora-compliant",
-            "--top-level",
+            "--automode",
         ]
         if self.pypi_package_version:
             cmd += ["-v", self.pypi_package_version]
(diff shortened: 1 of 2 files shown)


praiskup:
Let's wait till this is in fedora and merge.  Thank you very much for the update!

praiskup:
one missing thing, @befeleme can you add a minimal pyp2spec version in Requires: in the corresponding spec file?

nikromen (approved):

befeleme:
JFYI: In Fedora 41 pyp2spec will be shipped tomorrow, it was stopped during the freeze.

rpmbuild/copr_rpmbuild/providers/pypi.py
```diff
@@ -56,7 +56,7 @@ def _produce_srpm_pyp2spec(self):
             "pyp2spec",
             self.pypi_package_name,
             "--fedora-compliant",
-            "--top-level",
```
praiskup, line 59: Why don't we need --top-level?

befeleme: It's included/hidden as implementation detail in the `automode` now.

PR #3479, opened 2024-10-15
rpmbuild: basic sanity check for rhsm

praiskup:
Check that we can actually import the script module, and that we call the script with correct arguments.

### DIFF
--- a/rpmbuild/tests/test_rhsm.py
+++ b/rpmbuild/tests/test_rhsm.py
@@ -1,0 +1,43 @@
+import os
+import sys
+import io
+import importlib
+from unittest import mock, skipIf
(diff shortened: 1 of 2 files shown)


rpmbuild/tests/test_rhsm.py
```diff
@@ -0,0 +1,44 @@
+
```
FrostyX, line 1: Nit: A blank line here

FrostyX (approved):

rpmbuild/tests/test_rhsm.py
praiskup: fixed

PR #3484, opened 2024-10-17
Add more logging to the copr-builder-ready script

FrostyX:
So it doesn't have an empty stdout when everything went fine, e.g.

```
[2024-10-16 22:31:55,344][  INFO][PID:3774651] Checking that builder machine is OK
[2024-10-16 22:31:56,157][  INFO][PID:3774651] 
[2024-10-16 22:31:56,158][  INFO][PID:3774651] Filling build.info file with builder info
```

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -255,6 +255,7 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
         # generalize it into a separate package that we could eventually use
         # here.
         cmd = "copr-builder-ready " + self.job.chroot
+        self.log.info("Running remote command: %s", cmd)
         rc, stdout, stderr = self.ssh.run_expensive(
             cmd, subprocess_timeout=660)
         self.log.info(stdout)
--- a/rpmbuild/bin/copr-builder-ready
+++ b/rpmbuild/bin/copr-builder-ready
@@ -88,6 +88,9 @@ def main():
     check_mock_config(chroot)
     if subscription_required(chroot):
         wait_for_subscription()
+    else:
+        print("Red Hat subscription not needed for {0}".format(chroot))
+    print("Builder is ready to be used")
 
 
 if __name__ == "__main__":


FrostyX:
Now the logs looks like this for Fedora

```
[2024-10-17 10:30:58,347][  INFO][PID:4101539] Red Hat subscription not needed for fedora-40-x86_64
Builder is ready to be used
```

And RHEL:

```
[2024-10-17 10:30:48,699][  INFO][PID:4101535] Checking Red Hat subscription (attempt #1)
Red Hat subscription active
Builder is ready to be used
```

backend/copr_backend/background_worker_build.py
```diff
@@ -255,6 +255,7 @@ def _check_vm(self):
         # generalize it into a separate package that we could eventually use
         # here.
         cmd = "copr-builder-ready " + self.job.chroot
+        self.log.info("Running remote command: %s", cmd)
```
praiskup, line 258: nit: ton't you want to show stderr, if provided? might give us some warnings in advance

praiskup (approved):
Anyway, looks good. Thanks!

backend/copr_backend/background_worker_build.py
FrostyX: We do

```
if rc:
    self.log.info(stderr)
```

which should IMHO work fine.

praiskup: but if rc == 0 and there's some output, we just swallow it

FrostyX: I did it this way because for some reason my stderr is polluted by this line when testing this locally

```
Shared connection to builder closed.
```

So I intentionally wanted to swallow it. But we haven't seen that message in STG so I am confused.

praiskup: :shrug: if you want to hear my opinion, I don't think the logs need to be pretty - but must be complete (minus credentials, haha). But it is up to you, I already approved this PR.

Issue #3450, opened 2024-10-03
Configure a production Pulp instance


FrostyX:

FrostyX:
We currently use https://pulp.stage.devshift.net as our STG instance. According to @dkliban we won't be able to use Basic Auth but will need to switch to client certificates for authentication. The good thing is, we will be able to do the same for STG to test it.

This is how we can work with the certificates in `python-requests` - https://docs.python-requests.org/en/latest/user/advanced/#client-side-certificates

I also have some links on how to obtain the certificates but I'd rather not post them publicly. Please ping me if interested.

FrostyX:
I think we can now close this:

- We prepared our production S3 bucket and AWS credentials for it
- The Pulp team generated crt+key pair for us, so that we can authenticate against Pulp
- I implemented support for the crt+key authentication in PR #3491 and hotfixed it in production
- We deploy the certs and Pulp config through ansible - https://pagure.io/fedora-infra/ansible/blob/main/f/roles/copr/backend/tasks/pulp.yml
- And we have our first Copr project with results stored in Pulp - https://copr.fedorainfracloud.org/coprs/g/copr/pulp-test/

FrostyX:
The only problem we have is that STG instance uses authentication through user+password and stores all results on a storage provided by the Pulp instance, and production uses crt+key and we need to provide our own S3 bucket. This difference is too much, so we should work on unifying this.

PR #3491, opened 2024-10-25
Allow using crt+key pair for Pulp authentication

FrostyX:
See #3450

This will be necessary for the production Pulp instance. But we still want to keep the possibility of username+password auth for our local development.

### DIFF
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -50,6 +50,14 @@ class PulpClient:
         """
         return (self.config["username"], self.config["password"])
 
+    @property
+    def cert(self):
+        """
+        See Client Side Certificates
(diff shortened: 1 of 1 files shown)


backend/copr_backend/pulp.py
```diff
@@ -74,7 +82,12 @@ def request_params(self):
         """
         Default parameters for our requests
         """
-        return {"auth": self.auth, "timeout": self.timeout}
+        params = {"timeout": self.timeout}
+        if all(self.cert):
+            params["cert"] = self.cert
+        else:
+            params["auth"] = self.auth
```
praiskup, line 89: Just curious, can we add a check that only `copr` has read access to the key file?
If not, we can make sure at deployment time.

praiskup (approved):

PR #3494, opened 2024-11-01
backend: manpage for copr-backend-resultdir-cleaner && logger fix

praiskup:
Build the manual page from an updated --help output.  The logger needs to be initialized in `_main`, otherwise the log-file can mistakenly be created under the `root` user before the `script_requires_user()` check is done.

### DIFF
--- a/backend/run/copr-backend-resultdir-cleaner
+++ b/backend/run/copr-backend-resultdir-cleaner
@@ -171,14 +171,13 @@ def clean_in(resultdir, dry_run=True):
 
 def _main():
     logging.basicConfig(level=logging.DEBUG)
-    setup_script_logger(LOG, "/var/log/copr-backend/resultdir-cleaner.log")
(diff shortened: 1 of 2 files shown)


backend/run/copr-backend-resultdir-cleaner
```diff
@@ -162,7 +166,9 @@ def clean_in(resultdir, dry_run=True):
 
 
 def _main():
-    args = parser.parse_args()
+    logging.basicConfig(level=logging.DEBUG)
+    setup_script_logger(LOG, "/var/log/copr-backend/resultdir-cleaner.log")
```
FrostyX, line 174: Can we do `os.path.join(opts["log_dir"], "resultdir-cleaner.log")` or you intentionally wanted to setup the logger before parsing the config?

praiskup: I just moved the lines (c&p), but using opts seems possible. PTAL.

FrostyX (approved):

PR #3509, opened 2024-11-11
Add script to migrate existing build results to Pulp

### LINKED ISSUE #3503: Script to migrate existing project (all of its data) to Pulp

FrostyX:
Fix #3503


FrostyX:
I successfully used the script to migrate data from the Copr STG instance to the Pulp STG instance.
https://copr.stg.fedoraproject.org/coprs/frostyx/pulp-migration/

It woks for CoprDirs as well:

```
docker run -it quay.io/fedora/fedora:40 bash
dnf install 'dnf-command(copr)'
dnf copr enable copr.stg.fedoraproject.org/frostyx/pulp-migration:custom:bar
dnf install hello
```

This successfully installs the package from Pulp.

backend/run/copr-change-storage
```diff
@@ -0,0 +1,119 @@
...
+    if args.src == args.dst:
+        print("The source and destination storage is the same, nothing to do.")
+        return
+
+    if args.src == "pulp":
+        print("Migration from pulp to somewhere else is not supported")
+        sys.exit(1)
+
+    if args.delete:
+        print("Data removal is not supported yet")
```
praiskup, line 70: Since we have a pre-initialized logger, I think we should avoid using `print()` command.

backend/run/copr-change-storage
```diff
@@ -0,0 +1,119 @@
...
+
+                # We cannot check return code here
+                storage.upload_build_results(chroot, resultdir, None)
+
+                result = storage.publish_repository(chroot)
+                if not result:
+                    log.error("Failed to publish a repository: %s", resultdir)
+                    break
+
+                log.info("OK: %s", resultdir)
```
praiskup, line 115: I suppose we can not make this in a transactional manner.. (if error happens, rollback).  But would it be possible to first analyze the situation and gather the tasks that need to be done, fail if some problem happens, and only if no problems happen - start the processing?

Also, I'm curious if whether we need a project lock (for building and other modification).

backend/run/copr-change-storage
```diff
@@ -0,0 +1,119 @@
...
+                continue
+
+            appstream = None
+            devel = None
+            storage = PulpStorage(
+                owner, subproject, appstream, devel, config, log)
+
+            for builddir in os.listdir(chrootdir):
+                resultdir = os.path.join(chrootdir, builddir)
+                if not os.path.isdir(resultdir):
```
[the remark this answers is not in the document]
FrostyX: You are right, that would cause problems. Updated.

backend/run/copr-change-storage
FrostyX: > But would it be possible to first analyze the situation and gather the tasks that need to be done, fail if some problem happens, and only if no problems happen - start the processing?

Sooo, I am not really sure how helpful this would be. Gathering tasks beforehand would probably avoid issues like the script trying to access a directory it doesn't have permissions to and then failing. Or something like this. But I suppose the majority of failures that can/will happen are going to happen due to networking issues or something else when actually uploading things to Pulp. And having a calculated list of tasks wouldn't IMHO help.

I would probably only remember or maybe pre-calculate the number of RPM files we are uploading and after everything is done, query Pulp to find out if we have the same number. Or maybe compare names of RPMs if we wanted to be more precise. If it doesn't match, we can either re-try several times, or just log it and manually review all failures.

> Also, I'm curious if whether we need a project lock (for building and other modification).

That dumping a lockfile in this script would be easy but changing our build-related code, action code, cron jobs, etc to respect the lock, sounds like a bigger problem.

If such a locking feature would be generally useful, then sure. But if the only purpose would be for the Pulp migration, I hope we could figure something easier.

For initial migrations of test users, I think we would be fine with "please don't submit new builds until the migration is finished". And the mass migration of everything, will be done in batches. So maybe we can just put an ugly hack into our build/action scheduler to temporarily hide all jobs that fall in the currently migrated batch.

praiskup (approved):

PR #3513, opened 2024-11-13
Skip only if all macros in ExclusiveArch and ExcludeArch evaluated

### LINKED ISSUE #3244: ExclusiveArch skipping something it shouldn't
We have for example this build which was skipped because of an `ExclusiveArch`
https://copr.fedorainfracloud.org/coprs/sentry/zig-tools/build/7399386/

It uses `ExclusiveArch:  %{zig_arches}` which on my laptop evaluates as

```
[jkadlcik@zeratul ~]$ rpm --eval "%{zig_arches}"
x86_64 aarch64 riscv64 mips64 mips64el mips64r6 mips64r6el
```

so it IMHO should be built for x86_64 chroots - but it was skipped.

<details>
<summary>Full spec file as a backup:</summary>

```
%global         debug_package   %{nil}

Name:           zls
Version:        0.11.0
Release:        1%{?dist}
Summary:        A Zig language server with features like autocomplete and goto definition 

License:        MIT
URL:            https://github.com/zigtools/zls
Source0:        %{url}/archive/%{version}.tar.gz#/%{name}-%{version}.tar.gz

ExclusiveArch:  %{zig_arches}

BuildRequires:  zig >= %{version}
BuildRequires:  zig-rpm-macros

%description
The Zig Language Server (ZLS) is a tool that implements Microsoft's Language
Server Protocol for Zig in Zig. In simpler terms: it'll provide you with
completions, go-to definition, etc. when you write Zig code!

%prep
%setup -q

%build
%zig_build

%install
%zig_install

%check
%zig_test

%files
%license LICENSE
%doc README.md
%{_bindir}/zls

%changelog
* Thu May 02 2024 Jan200101 <sentrycraft123@gmail.com> - 0.11.0-1
- Update to 0.11

* Sat Mar 26 2022 Jan Drögehoff <sentrycraft123@gmail.com> - 0.9.0-1
- Update to 0.9.0

* Tue Sep 14 2021 Jan Drögehoff <sentrycraft123@gmail.com> - 0.1.0-1
- Bump release
```

</details>

FrostyX:
Fix #3244

It can happen that our builder doesn't understand some architecture macro, e.g. `%{zig_arches}`. In such a case when it is used in `ExclusiveArch`, we shouldn't skip all chroots like we do now. A better approach is to be safe and avoid any skipping. Builds will fail in some chroots that should have been skipped but that can be workarounded by the user by manually submitting the build only for a subset of chroots.

### DIFF
--- a/rpmbuild/copr_rpmbuild/helpers.py
+++ b/rpmbuild/copr_rpmbuild/helpers.py
@@ -428,14 +428,24 @@ class Spec:
         """
         Evaluated %{exclusivearch} as a list
         """
-        return self.safe_attr("exclusivearch").split()
+        values = self.safe_attr("exclusivearch").split()
+        unknown = " ".join([x for x in values if x.startswith("%")])
+        if unknown:
+            log.warning("Unknown macros in ExclusiveArch: %s", unknown)
+            return []
+        return values
 
     @property
     def excludearch(self):
         """
         Evaluated %{excludearch} as a list
         """
-        return self.safe_attr("excludearch").split()
+        values = self.safe_attr("excludearch").split()
+        unknown = " ".join([x for x in values if x.startswith("%")])
+        if unknown:
+            log.warning("Unknown macros in ExcludeArch: %s", unknown)
+            return []
+        return values
 
     def safe_attr(self, name):
         """
--- a/frontend/coprs_frontend/tests/test_logic/test_builds_logic.py
+++ b/frontend/coprs_frontend/tests/test_logic/test_builds_logic.py
@@ -655,3 +655,43 @@ class TestBuildsLogic(CoprsTestCase):
         build = models.Build.query.get(1)
         assert build.source_state == "succeeded"
         assert not os.path.exists(storage)
+
+    @pytest.mark.parametrize(
+        "exclusivearch,states", [
+            (["x86_64", "i386"], ["waiting", "waiting"]),
+            (["x86_64"], ["waiting", "skipped"]),
+            ([], ["waiting", "waiting"]),
+        ]
+    )
+    @pytest.mark.usefixtures(
+        "f_users", "f_coprs", "f_builds", "f_mock_chroots", "f_db")
+    def test_skipping_chroots(self, exclusivearch, states):
+        build = BuildsLogic.add(self.u2, "foo", self.c2)
+        self.db.session.commit()
+        assert len(build.chroots) == 0
+        data = {
+            "builds": [{
+                "id": 5,
+                "task_id": 5,
+                "srpm_url": "http://foo",
+                "status": 1,
+                "pkg_name": "foo",
+                "pkg_version": 1,
+                "chroot": "srpm-builds",
+                "results": {
+                    "epoch": None,
+                    "excludearch": [],
+                    "exclusivearch": exclusivearch,
+                    "name": "biosdevname",
+                    "release": "17",
+                    "version": "0.7.3"
+                },
+            }]
+        }
+        response = self.tc.post(
+            "/backend/update/",
+            headers=self.auth_header,
+            json=data,
+        )
+        assert response.status_code == 200
+        assert [x.state for x in build.build_chroots] == states


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1132,15 +1132,19 @@ def finish(chroot, status):
 
                     exclusivearch = upd_dict["results"].get("exclusivearch")
                     if exclusivearch and arch not in exclusivearch:
-                        chroot.status_reason = \
-                            "This chroot was skipped because of ExclusiveArch"
-                        finish(chroot, StatusEnum("skipped"))
+                        # Skip only if all macros in ExclusiveArch evaluated
+                        if not [x for x in exclusivearch if x.startswith("%")]:
```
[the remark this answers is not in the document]
praiskup: And, ... coul give user a hint that something weird is happening? in the sources' builder-live.log/backend.log?

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1132,15 +1132,19 @@ def finish(chroot, status):
...
+                                "This chroot was skipped because of ExclusiveArch"
+                            finish(chroot, StatusEnum("skipped"))
 
                     excludearch = upd_dict["results"].get("excludearch")
                     if arch in excludearch:
-                        chroot.status_reason = \
-                            "This chroot was skipped because of ExcludeArch"
-                        finish(chroot, StatusEnum("skipped"))
+                        # Skip only if all macros in ExcludeArch evaluated
+                        if not [x for x in excludearch if x.startswith("%")]:
```
praiskup, line 1144:
```
In [1]: if [False, False]:
   ...:     print("true")
   ...: 
true

Comment (FrostyX) [reply frontend/coprs_frontend/coprs/logic/builds_logic.py:1144]: But this is a different case

```python
>>> [x for x in ["foo", "bar", "baz"] if x.startswith("%")]
[]

>>> [x for x in ["foo", "%bar", "baz"] if x.startswith("%")]
['%bar']
```

The first one is `False`, the second one is `True`.

frontend/coprs_frontend/coprs/logic/builds_logic.py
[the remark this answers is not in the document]
FrostyX: > Macros may look like %foo a b %c, so probably we should just check for '%' in x

True, but if I understand correctly, this shouldn't matter. For example, in our beaker tests we use this specfile
https://src.fedoraproject.org/rpms/biosdevname/blob/rawhide/f/biosdevname.spec

It has `ExclusiveArch: %{ix86} x86_64` ... If it was reversed and it was `ExclusiveArch: x86_64 %{ix86}`, that would be the case you are talking about?

It shouldn't be a problem because in `copr-rpmbuild`, we produce a list out of them:

```
Output: ['exclusivearch-test-1-1.src.rpm']
Running SRPMResults tool
Package info: {
    "name": "exclusivearch-test",
    "epoch": null,
    "version": "1",
    "release": "1",
    "exclusivearch": [
        "noarch",
        "x86_64",
        "aarch64"
    ],
    "excludearch": []
}
SRPMResults finished
```

So here, in this frontend code, our `exclusivearch` variable is `["noarch", "x86_64", "aarch64"]`. If there was an unevaluated macro, it would be `["noarch", "x86_64", "aarch64", "%{zig_arches}"]`. So IMHO `startswith` should work as expected.

That being said, I have no issue with `'%' in x` and we can use it if you prefer.

> And, ... coul give user a hint that something weird is happening? in the sources' builder-live.log/backend.log?

That would be great but ... probably not? This is a frontend code, so no user-logs. And we cannot display an UI banner here either. The only indication is that at the end of the SRPM builder-live.log, there will be unevaluated macro in the `exclusivearch` list.

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: ah, this is a check for emptiness (not obvious), then good!

frontend/coprs_frontend/coprs/logic/builds_logic.py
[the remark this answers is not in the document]
praiskup: You are correct. Awesome analysis. Can we have a test for this case?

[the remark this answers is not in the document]
FrostyX: > What would happen if we produced an empty list on copr-rpmuibuild then?

That would probably be the easiest but I thought of this more as a workaround.
Anyway, I added some warnings to the builder-live.log and also tests. WDYT?

rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -428,14 +428,22 @@ def exclusivearch(self):
         """
         Evaluated %{exclusivearch} as a list
         """
-        return self.safe_attr("exclusivearch").split()
+        values = self.safe_attr("exclusivearch").split()
+        for macro in values:
+            if macro.startswith("%"):
+                log.warning("Unknown macro in ExclusiveArch: %s", macro)
+        return values
```
praiskup, line 435: I'm not complaining, just curious ... When we are doing this already, why do we have to report back to backend && frontend useless set of macros (some unexpanded)?  Wouldn't it be better to return just empty list?

FrostyX: Updpated, PTAL

praiskup (approved):
:rocket: thank you!

PR #3519, opened 2024-11-19
docker: create /var/lock/copr-dist-git directory

FrostyX:
Otherwise the import command:

    copr-distgit-process-import --build-id 7252586 --worker-id import_worker:7252586

Fails with the following error:

    FileNotFoundError: [Errno 2] No such file or directory: '/var/lock/copr-dist-git/_@_var_@_lib_@_dist-git_@_cache_@_lookaside_@_pkgs_@_frostyx_@_test-devel_@_copr-cli.lock'


praiskup:
Thank you!

Issue #3270, opened 2024-05-20
Upgrade builders to Fedora 40


praiskup:
It is out now: https://fedorapeople.org/groups/schedule/f-40/f-40-key-tasks.html

praiskup:
F41 out now: https://fedorapeople.org/groups/schedule/f-41/f-41-key-tasks.html

hroncok:
As of today, at least 146 rawhide specfiles cannot be SRPM'ed on Fedora \< 41.

 - 9 use a BuildSystem
 - 137 use `%autosetup -C`

(The groups are distinct.)

praiskup:
Note for myself: `ansible` updated on copr-be-dev, do the same on copr-be.

praiskup:
I was able to generate AWS images, but libvirt images are blocked on: `https://pagure.io/cloud-sig/issue/433`

praiskup:
Nah, spent some time on getting things work (for some reaons, selinux was breaking VM log-in, and sshd server).

Then I realized that we have too old ansible on RHEL 8 hypervisors :-( to handle DNF5 -based fedora:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'Base' object has no attribute 'load_config_from_file'

```

praiskup:
I was able to generate the golden images finally, and tests are passing - except for the problems with pyp2spec #3517.

praiskup:
Discussed on today's meeting... and except for the pyp2spec problems, everything seems fine -> and the consensus in #3517 seems to be we should push the move to F41 forward.  Doing so now.

praiskup:
Done: https://lists.fedorahosted.org/archives/list/copr-devel@lists.fedorahosted.org/thread/LTJXSOYFX56QA37F6JDQPCDM2G3C2FTC/

praiskup:
Uh, one P9 machine is missing.

hroncok:
Thanks.

praiskup:
There's also the known problem with old specfiles using `%patchN` commands.  Users should migrate to `%patch -PN`, `%patch N` or `%autosetup` in the best case.

PR #3526, opened 2024-11-20
doc: updates after F39 → F41 builder update

praiskup:
Fixes: #3270

### DIFF
--- a/doc/how_to_upgrade_builders.rst
+++ b/doc/how_to_upgrade_builders.rst
@@ -3,17 +3,19 @@
 How to upgrade builders
 =======================
 
-This article explains how to upgrade the Copr builders images in
+This article explains how to upgrade generate the Copr builder images (so called
+"golden images") and how to migrate the Fedora Copr to them.  Namely for
 
 - :ref:`AWS <prepare_aws_source_images>` (x86_64 and aarch64),
-- :ref:`LibVirt/OpenStack <prepare_libvirt_source_images>` (x86_64 and ppc64le),
-- :ref:`IBM Cloud <prepare_ibmcloud_source_images>` (s390x),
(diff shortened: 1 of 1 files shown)


doc/how_to_upgrade_builders.rst
```diff
@@ -76,7 +77,9 @@ Repeat the previous steps.
 
 The remaining step is to configure ``copr_builder_images.aws.{aarch64,x86_64}``
 options in `Ansible git repo`_, in file ``inventory/group_vars/copr_dev_aws``
-and reprovision the ``copr-be-dev`` instance, see :ref:`Testing`.
+and reprovision the ``copr-be-dev`` instance, see :ref:`Testing`.  Document the
```
FrostyX, line 80: I suppose "Document" means adjusting this comment?

    # n-1: aarch64: ami-0f93da3f4f368afb4
    # n-1: x86_64: ami-062f0015f85924f2f
    # n-0: aarch64: ami-0a2abb60d2e09f6fa
    # n-0: x86_64: ami-0460b8805bf9781cf

`n-0` is confusing because `n-0 = n` ...

FrostyX: I jumped with the comment too early. I am now reading the next line which says 

    previous ``ami-*`` variants as ``n-1``

but I would expect to take the current `x86_64: ami-0c430e4bc139d93ec` put the new value there, and move the old one to `n-0: x86_64` not `n-1` like you are suggesting.

FrostyX (commented):
Otherwise LGTM

doc/how_to_upgrade_builders.rst
praiskup: The `n-0` variants are just confusing, I'm going to remove those right away.  I just commented images I generated last week so I don't forget to remove them.  And then I forgot to drop the comment.

PR #3527, opened 2024-11-22
Implement a check for every storage whether a repository is available

FrostyX:
Otherwise, builds in Pulp projects will fail with this:

    Waiting for copr_base repository
    Waiting for copr_base repository
    Backend process error: Giving up waiting for copr_base repository

So far it worked only because we didn't create Pulp projects directly but rather migrated existing projects into Pulp and therefore this path existed on the backend.

### DIFF
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -195,7 +195,7 @@ class PulpStorage(Storage):
         self.client = PulpClient.create_from_config_file()
 
     def init_project(self, dirname, chroot):
-        repository = self._repository_name(chroot)
+        repository = self._repository_name(chroot, dirname)
         response = self.client.create_repository(repository)
         if not response.ok and "This field must be unique" not in response.text:
             self.log.error("Failed to create a Pulp repository %s because of %s",
@@ -206,7 +206,7 @@ class PulpStorage(Storage):
         # mentioned by its href, not name
         repository = self._get_repository(chroot)
 
-        distribution = self._distribution_name(chroot)
+        distribution = self._distribution_name(chroot, dirname)
         response = self.client.create_distribution(distribution, repository)
         if not response.ok and "This field must be unique" not in response.text:
             self.log.error("Failed to create a Pulp distribution %s because of %s",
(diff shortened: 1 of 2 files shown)


backend/copr_backend/storage.py
```diff
@@ -172,6 +180,11 @@ def delete_builds(self, dirname, chroot_builddirs, build_ids):
                         self.log.debug("can't remove %s", log_path)
         return result
 
+    def repository_exists(self, chroot):
+        repodata = os.path.join(self.opts.destdir, self.owner, self.project,
```
praiskup, line 184: Is `self.project` actually a `self.coprdir`?

FrostyX: It's actually this:

https://github.com/fedora-copr/copr/blob/e123e4e16c0d1d3dfb17ff4610fc35d781292365/backend/copr_backend/actions.py#L100

which comes from frontend

https://github.com/fedora-copr/copr/blob/e123e4e16c0d1d3dfb17ff4610fc35d781292365/backend/copr_backend/actions.py#L100

so it should only be a project name, not CoprDir name. Which should IMHO be what we want because CoprDirs don't have their own repositories available (see #3369) and they use the main repo.

praiskup: Separate CoprDirs have their own repositories, and they should be created before build. We previously used `self.chroot_dir`, see #3571 how the variable looks like.

I believe we are changing the semantics now :shrug:

It is truth that we do not necessarily need the CoprDir's metadata, because these are not used at build-time, but that is probably a "temporary" issue.

FrostyX: Just to confirm that you are right. Previously, we were checking if this exists

```
/var/lib/copr/public_html/results/frostyx/test-pulp-devel-7:custom:foo/fedora-rawhide-x86_64/repodata/repomd.xml
```

and we check

```
/var/lib/copr/public_html/results/frostyx/test-pulp-devel-7/fedora-rawhide-x86_64/repodata/repomd.xml
```

I will fix it, so we don't change the behavior.

> It is truth that we do not necessarily need the CoprDir's metadata, because these are not used at build-time, but that is probably a "temporary" issue.

This is the root of the problem. I didn't treat it as a "temporary issue" but rather "this is how it is". But I agree with you, we eventually want to get this fixed.

FrostyX: I hope I addressed the issue correctly. PTAL.
There is also one more commit fixing semi-related issue.

Issue #3460, opened 2024-10-07
Move infra from Fedora 39 to 41


praiskup:
Don't forget to go through hot-fixes:
https://github.com/fedora-copr/copr/issues?q=label%3Ahot-fixed+

FrostyX:
After the update, make sure the `iptables-legacy` wasn't pulled back in (see #3439)

FrostyX:
We are currently working on upgrading the STG instances

- [ ] Frontend (@FrostyX)
- [ ] DistGit (@FrostyX)
- [ ] Backend (@praiskup)
- [ ] Keygen (@praiskup)

FrostyX:
The package `golang-github-prometheus-node-exporter` is now dead in Fedora. I tried to rebuild it in Copr but had some issues with. Even for F39 which should still support it, so I am not sure if it could work for F41 or not. Temporarily removing it from the playbooks:
https://pagure.io/fedora-infra/ansible/c/476d71fb8acc639c254015898e296347a4289ccb

praiskup:
We need that package for the prometheus monitoring in Red Hat network (grafana metrics).  I think the package was renamed to `node-exporter`, per changelog:
```
* so led 13 2024 Mikel Olasagasti Uranga <mikel@olasagasti.info> - 1.6.1-1
- Initial package, rename from golang-github-prometheus-node-exporter -
  Closes rhbz#2250145
```

FrostyX:
When upgrading the database via `reindexdb --all`, it spammed the output with hundreds of lines like this:

```
WARNING:  database "coprdb" has a collation version mismatch
DETAIL:  The database was created using collation version 2.38, but the operating system provides version 2.40.
HINT:  Rebuild all objects in this database that use the default collation and run ALTER DATABASE coprdb REFRESH COLLATION VERSION, or build PostgreSQL with the right library version.
```

I did as suggested, to fix the warnings:

```
bash-5.2$ psql coprdb
coprdb=# ALTER DATABASE coprdb REFRESH COLLATION VERSION;
coprdb=# ALTER DATABASE postgres REFRESH COLLATION VERSION;
bash-5.2$ reindexdb --all
```

No warnings now.

FrostyX:
A lot of errors in the httpd log:

```
[root@copr-fe-dev ~][STG]# tail -f /var/log/httpd/error_log
[...]

PR #3529, opened 2024-11-23
Upgrade containers for local development to F41

FrostyX:
See #3460

Several issues appeared during the upgrade, and we are addressing them here:

- To my dismay, `python3-ipdb` is not available on F41. See https://bugzilla.redhat.com/show_bug.cgi?id=2280968 Hopefully, the upstream issue will get fixed and the pacakge will get unretired in Fedora. I love using it.
- The `dist-git-client` is not a part of Copr codebase anymore and it was moved to dist-git. See https://github.com/release-engineering/dist-git As a consequence, the name of the configuration directory changed.
- Removing mlocate. We discussed it several times and IIRC we don't need it

Also, please recreate the whole environment through `docker-compose down --rmi all --volumes` and `docker-compose up -d`.


praiskup (approved):
thank you

Issue #3495, opened 2024-11-01
Perform Backend's resultdir cleanup


praiskup:
There's the `/usr/bin/copr-backend-resultdir-cleaner` utility that we probably 
never executed.  Except that it removes `chroot_scan` sub-directories, it also
log-out some directories that we could remove ("manually", after reviewing the
list of occurences).

This ticket is the "administrator" task that needs to be done (babysit the script, etc.).

praiskup:
Still running, but some effect is already visible:
```
Mon Nov 25 04:25:30 AM UTC 2024
Filesystem                                          Inodes IUsed IFree IUse% Mounted on
/dev/mapper/c[copr@copr-be ~][PROD]$ date && df -i /var/lib/copr/public_html
Mon Nov 25 04:26:23 AM UTC 2024
Filesystem                                             Inodes    IUsed     IFree IUse% Mounted on
/dev/mapper/copr--backend--data-copr--backend--data 664791040 44762982 620028058    7% /var/lib/copr/public_html
opr--backend--data-copr--backend--data   634M   43M  592M    7% /var/lib/copr/public_html

**Comment by praiskup** [comment]
```
TODO DIR APPSTREAMFAIL: /var/lib/copr/public_html/results/ellana/spotify-qt/fedora-rawhide-x86_64/appdata
WARNING:__main__:TODO DIR APPSTREAMFAIL: /var/lib/copr/public_html/results/ellana/spotify-qt/fedora-rawhide-x86_64/appdata

real    4281m50.238s
user    23m47.398s
sys     56m24.556s

$ df -i /var/lib/copr/public_html
Filesystem
/dev/mapper/copr--backend--data-copr--backend--data 664791040 43585645 621205395    7% /var/lib/copr/public_html

```

praiskup:
59639988 -> 43585645 => only ~27% of inodes down, I thought it would be more close to ~50% :-( but still good.

praiskup:
Run, 43.6M => 43.2M.  Done.

PR #3537, opened 2024-11-28
resultdir-cleaner: remove old appstream failures

praiskup:
These are no longer that frequent as we disabled appstream-builder by default long time ago.

Relates: #3495

### DIFF
--- a/backend/run/copr-backend-resultdir-cleaner
+++ b/backend/run/copr-backend-resultdir-cleaner
@@ -94,9 +94,7 @@ def clean_in(resultdir, dry_run=True):
                 continue
 
             if subdir in ["tmp", "cache", "appdata"]:
-                # tmp => 95k * 7 inodes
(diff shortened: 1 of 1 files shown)


backend/run/copr-backend-resultdir-cleaner
```diff
@@ -94,9 +94,7 @@ def clean_in(resultdir, dry_run=True):
                 continue
 
             if subdir in ["tmp", "cache", "appdata"]:
-                # tmp => 95k * 7 inodes
-                # cache => 95k * 2 inodes
-                todo_directory(chroot_subdir_path, "APPSTREAMFAIL")
+                remove_old_dir(chroot_subdir_path, dry_run)
```
praiskup, line 97: There's 150k of such directories;  I'd like to clean them once this is merged (hot-fix).

nikromen (approved):

PR #3540, opened 2024-12-02
docker: remove legacy link and version

nikromen:
version: https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-top-level-element-obsolete

---

link: https://docs.docker.com/engine/network/links/

having link in compose file result in error:
Error response from daemon: bad parameter: link is not supported Error: executing /usr/libexec/docker/cli-plugins/docker-compose up -d: exit status 1

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -13,8 +13,6 @@
 # Other ports:
 #   5009:5432 -- PostgreSQL database
 
-version: '3'
-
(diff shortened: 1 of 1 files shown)


nikromen:
We already have `depends_on` database which should be enough in this case so `link` or its alternative - creating network  between those - isn't needed, right?

FrostyX (approved):

PR #3549, opened 2024-12-06
backend, frontend: implement support for Pulp manual createrepo

### LINKED ISSUE #3498: manual createrepo (devel sub-directory)

FrostyX:
Fix #3498


FrostyX:
I just wanted to push a proof of concept. This still needs to fix some quirks and refactor the code. Worth mentioning:

- This approach of having two separate repositories and "copying packages from one to another" was recommended by Dennis
- Currently we use `fedora-rawhide-x86_64/devel/` directories on backend, but Pulp has a problem with that because the `devel` directory would encroach an existing repository. So that I used `fedora-rawhide-x86_64_devel/` naming. I don't like it very much.

backend/copr_backend/storage.py
```diff
@@ -379,16 +427,16 @@ def repository_exists(self, dirname, chroot):
         return response.ok
 
     def _repository_name(self, chroot, dirname=None):
+        # On backend we use /devel but Pulp has a problem with that
```
praiskup, line 430: Will the CDN -> Backend -> PULP redirect work?

backend/copr_backend/pulp.py
```diff
@@ -187,6 +187,19 @@ def delete_content(self, repository, artifacts):
...
 
+    def copy_content(self, repository, units):
+        """
+        Copy a list of content units (RPM packages) into this repository
+        The "copy" isn't the right word because the actual data isn't touched
+        at all. Only a new database entry pointing to the data is created.
+        So regardless of the number of units, this operation should be fast.
+        https://pulpproject.org/pulp_rpm/restapi/#tag/Repositories:-Rpm/operation/repositories_rpm_rpm_modify
+        """
+        path = os.path.join(repository, "modify/")
```
praiskup, line 198: This `modify/` suffix deserves an in line comment :-)

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -610,9 +610,11 @@ def generate_build_config(cls, copr, chroot_id):
             repos[0]["module_hotfixes"] = True
 
         if not copr.auto_createrepo:
+            suffix = "_devel" if copr.storage == StorageEnum.pulp else "/devel"
+            baseurl = copr.repo_url + "/{0}{1}/".format(chroot_id, suffix)
```
praiskup, line 614: Can we document the PULP issues with `/devel` here as well as we do for the backend side?

praiskup:
Have you considered to have two distributions, one for normal repo and one for the devel repo? https://pulpproject.org/pulpcore/docs/user/learn/lifecycle-promotion/

backend/copr_backend/pulp.py
nikromen: the endpoint (with the suffix) is mentioned in the documented link 

```
http://localhost:24817{rpm_rpm_repository_href}modify/
```

Conan-Kudo:
What about `fedora-rawhide-x86_64-staged/` instead?

FrostyX:
Sorry for the delay, I finally updated the PR. It now uses two distributions (aka the "Lifecyle Promotion Support") instead of two repositories, as you've all suggested to do.

PTAL

backend/copr_backend/storage.py
```diff
@@ -308,8 +339,8 @@ def publish_repository(self, chroot, **kwargs):
                 "Pulp task {0} didn't create any resources".format(task))
 
         publication = resources[0]
-        distribution_name = self._distribution_name(chroot)
-        distribution = self._get_distribution(chroot)
+        distribution_name = self._distribution_name(chroot, devel=self.devel)
+        distribution = self._get_distribution(chroot, devel=self.devel)
```
praiskup: It just seems weird to call `self.foo(some_arg=self.bar)`, `self.foo()` should be enough- is this needed?

backend/copr_backend/storage.py
```diff
@@ -205,15 +205,46 @@ def init_project(self, dirname, chroot):
...
+        # is finished. We don't want to copy the packages there. This
+        # `init_project` method is called when a project is created (for which
+        # the copying does nothing) but also when a "Regenerate" button is
+        # clicked (for which we copy the data). We should probably separate
+        # these two concepts.
+        if not self.devel:
+            response = self.client.get_distribution(devel_distribution_name)
+            if not response.ok:
+                self.log.error("ERRR")
+                return False
```
praiskup: This is extremely hard reading :-)

We seem to be syncing the public repo with devel when `not self.devel`.  I think `self.devel` is not a per-repository config, but the current action option (i.e. self.devel can be false even though copr.devel = true).

So consider we a) create normal repository, b) than we toggle the "manual createrepo" feature.  Do we sync the newly created `devel` package with the old `public` one?

praiskup:
I'd prefer having a review session :-) on (post) today's meeting

backend/copr_backend/storage.py
FrostyX: That's a fair point. Fixed.

backend/copr_backend/storage.py
FrostyX: > This is extremely hard reading :-)

I realize that, sorry. Hopefully a meeting will make it clearer

>  I think self.devel is not a per-repository config, but the current action option (i.e. self.devel can be false even though copr.devel = true).

Exactly. Which explains the next question

> We seem to be syncing the public repo with devel when not self.devel

When you have a `self.devel` project, you update the devel distribution after builds finish. In such a project, when a "Regenerate" button is clicked, it generates an action with `not self.devel`.

 > So consider we a) create normal repository, b) than we toggle the "manual createrepo" feature. Do we sync the newly created devel package with the old public one?

That's a case that I didn't consider at all when developing this, and it turns out, it somehow syncs it. At the moment I don't understand why or how but I can figure it out.

But shout it even matter? During a build, both devel and public repositories are enabled and therefore even if the devel one was empty, the packages would be installed from the public one.

praiskup: Summary from the talk we had;  there's this table: https://docs.pagure.org/copr.copr/createrepo.html
While the old Storage doesn't care if devel repo exists or not (full createrepo is run automatically reading all the RPMs, including those that were "devel"-only), the new cares.

FrostyX: The bug we found on our call wasn't a bug, but we misunderstood the code. Proving your point that it needs more comments. I added more comments, PTAL.

backend/copr_backend/storage.py
```diff
@@ -205,15 +216,74 @@ def init_project(self, dirname, chroot):
...
-        response = self.client.create_publication(repository)
-        return response.ok
+        # If a project enabled the manual createrepo mode, we need to create a
+        # devel distribution to be consumed by other builds within the project
+        if self.devel:
+            response = self.client.create_distribution(devel_distribution_name, repository)
+            if not response.ok and "This field must be unique" not in response.text:
+                self.log.error("Failed to create a Pulp distribution %s because of %s",
+                               devel_distribution_name, response.text)
+                return False
```
praiskup, line 328: This is still very tricky :sweat_smile:  If create_distribution() fails, we return False, but if it succeeds, we continue to the next section, which is `if not self.devel`.

Perhaps we should just do `if self.devel else:`?

praiskup: Are we taking into account that, when `devel=True`, there are actually two createrepo actions to be handled by Backend?  Per this Frontend's request:
https://github.com/fedora-copr/copr/blob/6e7b6d4c4da042dafc2028f6cca1de248c9955f4/frontend/coprs_frontend/coprs/logic/actions_logic.py#L120-L124

FrostyX:
Once again, this PR proves to be too complicated because I tested it once again while rebasing to `main` and found a bug.

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
```diff
@@ -495,8 +496,11 @@ class RegenerateRepos(Resource):
     @staticmethod
     def _common(copr):
         with db_session_scope():
-            ActionsLogic.send_createrepo(copr, devel=False)
-
+            ActionsLogic.send_createrepo(
+                copr,
+                devel=False,
+                reason=CreaterepoReason.manual_createrepo_evenet,
```
praiskup: evenet => event

common/copr_common/enums.py
```diff
@@ -142,3 +142,23 @@ class StorageEnum(metaclass=EnumType):
...
+    we run it differently.
+
+    See when do we run createrepo:
+    https://docs.pagure.org/copr.copr/createrepo.html#when-do-we-run-createrepo
+    """
+    vals = {
+        "new_project": 0,
+        "new_chroot": 1,
+        "new_coprdir": 2,
+        "manual_createrepo_toggle": 3,
```
praiskup, line 159: is it enough to have "toggle?"  shouldn't have on/off?

praiskup: hm, there's hte "acr" flag

praiskup:
We should bump: https://github.com/fedora-copr/copr/blob/235810c90dfe47f59bde96d1cce2bb1cf32aa338/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py#L28

backend/copr_backend/storage.py
```diff
@@ -193,26 +204,95 @@ def __init__(self, *args, **kwargs):
...
+            if not response.ok:
+                self.log.error("Failed to update Pulp distribution %s for because %s",
+                               public_distribution, response.text)
+
+        # When a project is being created. We do nothing special.
+        elif reason == CreaterepoReason.new_project:
+            pass
+
+        # We want to disable the manual createrepo feature on a project
+        elif reason == CreaterepoReason.manual_createrepo_toggle:
```
praiskup, line 319: Ad the comment, disable, or enable/disable?

praiskup:
:+1: for having the enum

backend/copr_backend/storage.py
FrostyX: Only disable because there is a condition `if self.devel:` above this one, which already eats up the enable action.

backend/copr_backend/storage.py
```diff
@@ -511,9 +587,12 @@ def _repository_name(self, chroot, dirname=None):
             chroot,
         ])
 
-    def _distribution_name(self, chroot, dirname=None):
+    def _distribution_name(self, chroot, dirname=None, devel=None):
+        # On backend we use /devel but in Pulp we cannot create subdirectories
         repository = self._repository_name(chroot, dirname)
-        if self.devel:
+        if devel is None:
+            devel = self.devel
```
praiskup, line 594: nit: this if a is None => then a is b is pretty tricky, considering that we don't know when b true

backend/copr_backend/storage.py
praiskup: Yeah, this still feels shady, it means that when `devel == True`, the `reason` never affects this method (this feels like we could have yet another state in the enum long term?).

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -614,9 +614,12 @@ def generate_build_config(cls, coprdir, chroot_id):
             repos[0]["module_hotfixes"] = True
 
         if not copr.auto_createrepo:
+            # On backend we use /devel but in Pulp we cannot create subdirectories
+            suffix = "-devel" if copr.storage == StorageEnum.pulp else "/devel"
+            baseurl = copr.repo_url + "/{0}{1}/".format(chroot_id, suffix)
```
praiskup, line 619: Long term we should have utility function for this, like `copr_dir.repo_url(chroot_id, devel=False)`.
Seems like we intentionally ignore copr-dirs?

praiskup (approved):
Seems good, thank you for the work!
There are a few nits/wishes, but these should not block the release.

Issue #3574, opened 2025-01-06
Broken per-arch backend limit


praiskup:
The `builds_max_workers_arch_per_owner=s390x=4` limit applies to all architectures:

```
[2025-01-06 11:00:34,128][ DEBUG][PID:646316][backend.build_dispatcher][worker_manager.py:run:466] Task '8474088-fedora-rawhide-i386' skipped, limit info: 'arch_s390x_owner', counter: x86_64_@rocm-packagers-sig=4, s390x_@fedora-llvm-team=4, aarch64_@fedora-llvm-team=1, x86_64_packit=4, x86_64_averbyts=1, x86_64_xxmitsu=4, x86_64_@theforeman=1, x86_64_jwakely=4, aarch64_jwakely=4, ppc64le_jwakely=4, s390x_jwakely=4, s390x_packit=1, ppc64le_packit=1, aarch64_packit=1, aarch64_solopasha=3, x86_64_solopasha=3, x86_64_maximizerr=1, x86_64_nucleo=1, x86_64_psimovec=2, aarch64_psimovec=2, x86_64_eclipseo=4, x86_64_lbalhar=4, x86_64_osloup=4, s390x_osloup=4, aarch64_osloup=2, ppc64le_osloup=4, x86_64_whitehara=4, ppc64le_whitehara=4, s390x_whitehara=4, aarch64_whitehara=4, s390x_@yggdrasil=1, x86_64_stsp=2
```

PR #3575, opened 2025-01-06
backend: per-arch/per-owner limit fixed

praiskup:
Prior this patch, the per-arch limit was applied to all architectures.

Fixes: #3574

### DIFF
--- a/backend/copr_backend/rpm_builds.py
+++ b/backend/copr_backend/rpm_builds.py
@@ -143,8 +143,15 @@ class ArchitectureUserWorkerLimit(HashWorkerLimit):
     Copr owner (user or group).
     """
     def __init__(self, architecture, limit):
+        def _hasher(job):
+            if job.requested_arch != architecture:
+                # When _hasher(job) returns None, no limit applies, see
+                # HashWorkerLimit._groups.add() for more info.
+                return None
+            return f"{job.requested_arch}_{job.owner}"
+
         super().__init__(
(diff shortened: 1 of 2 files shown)


backend/copr_backend/rpm_builds.py
```diff
@@ -143,8 +143,15 @@ class ArchitectureUserWorkerLimit(HashWorkerLimit):
     Copr owner (user or group).
     """
     def __init__(self, architecture, limit):
+        def _hasher(job):
+            if job.requested_arch != architecture:
+                # When _hasher(job) returns None, no limit applies, see
+                # HashWorkerLimit._groups.add() for more info.
+                return None
+            return f"{job.requested_arch}_{job.owner}"
```
FrostyX, line 151: Returning `None` seems strange. Aren't we supposed to return some string, like this ^^ and something outside this class counts how many same strings are there and compare them to the limit?

praiskup: If we return None for a job, that particular one is not calculated; HashWorkerLimit uses StringCounter.add():
https://github.com/fedora-copr/copr/blob/9d0d6f333898bf2fb80fcfe090bd2525149085ea/common/copr_common/worker_manager.py#L111-L114

FrostyX: I am reading the issue again, and I probably understand now. The `None` makes sense.

praiskup:
OK to merge?

PR #3576, opened 2025-01-06
releng: tito --test builds to have higher NVRs than released versions

praiskup:
Historically, Tito only modified the Release tag for `--test` builds. Now, we also bump the Version field to ensure that Mock built from Git revisions always has a higher NVR than (previously) released packages.

https://github.com/rpm-software-management/tito/releases/tag/tito-0.6.27-1

### DIFF
--- a/.tito/tito.props
+++ b/.tito/tito.props
@@ -4,3 +4,4 @@ tagger = tito.tagger.VersionTagger
 changelog_do_not_remove_cherrypick = 0
 changelog_format = %s
 fetch_sources = True
+test_version_suffix = .test


.tito/tito.props
```diff
@@ -4,3 +4,4 @@ tagger = tito.tagger.VersionTagger
 changelog_do_not_remove_cherrypick = 0
 changelog_format = %s
 fetch_sources = True
+test_version_suffix = .test
```
FrostyX, line 7: Nit: would `.git` be better?

praiskup: The difference would be:

```diff
/tmp/tito/copr-backend-2.1.test-1.git.38.3719a4f.fc41.src.rpm
/tmp/tito/copr-backend-2.1.git-1.git.38.3719a4f.fc41.src.rpm
```

Nobrainer, just note there's the double-`git` part `git-1.git`.

FrostyX (approved):

PR #3585, opened 2025-01-15
ci: enable automatic merging of pull requests

jamacku:
PR that meets the following criteria will be automatically merged:

- Passing CI
- Approved by at least one reviewer
- No Draft or WIP title
- No conflicts

The merging workflow runs on schedule every 45 minutes. It can be executed manually on demand from the Actions tab. It internally uses several GitHub Actions to perform the merge.

- https://github.com/redhat-plumbers-in-action/gather-pull-request-metadata to gather PR metadata
- https://github.com/redhat-plumbers-in-action/pull-request-validator to check reviews and CI status
- https://github.com/redhat-plumbers-in-action/auto-merge to perform the merge
- https://github.com/redhat-plumbers-in-action/issue-commentator to comment on PRs about the merge status

---

I have tested the workflow on https://github.com/jamacku/copr/pull/1

A working example can be found at https://github.com/redhat-plumbers/systemd-rhel10/pull/45#issuecomment-2410935343. It contains a few more pieces, but merging logic is also present.

---

There are a few limitations/requirements:

- When merging PR this way, it doesn't trigger `on: push` workflows - https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
- Maintainers who approve PRs need to have member visibility set to public; otherwise, GitHub Action will not see that they are members eligible to review PRs.

/cc @xsuchy

### DIFF
--- a/.github/workflows/auto-merger.yml
+++ b/.github/workflows/auto-merger.yml
@@ -1,0 +1,100 @@
+name: Auto Merger
+on:
+  schedule:
+    # Workflow runs every 45 minutes
+    - cron: '*/45 * * * *'
+  workflow_dispatch:
+    inputs:
+      pr-number:
+        description: 'Pull Request number/s ; when not provided, the workflow will run for all open PRs'
+        required: true
+        default: '0'
+
+permissions:
+  contents: read
+
+jobs:
+  # Get all open PRs
+  gather-pull-requests:
+    if: github.repository == 'fedora-copr/copr'
+    runs-on: ubuntu-latest
+
+    outputs:
+      pr-numbers: ${{ steps.get-pr-numbers.outputs.result }}
+      pr-numbers-manual: ${{ steps.parse-manual-input.outputs.result }}
+
+    steps:
+      - id: get-pr-numbers
+        if: inputs.pr-number == '0'
+        name: Get all open PRs
+        uses: actions/github-script@v6
+        with:
+          # !FIXME: this is not working if there is more than 100 PRs opened
+          script: |
+            const { data: pullRequests } = await github.rest.pulls.list({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: 'open',
+              per_page: 100
+            });
+            return pullRequests.map(pr => pr.number);
+
+      - id: parse-manual-input
+        if: inputs.pr-number != '0'
+        name: Parse manual input
+        run: |
+          # shellcheck disable=SC2086
+          echo "result="[ ${{ inputs.pr-number }} ]"" >> $GITHUB_OUTPUT
+        shell: bash
+
+  validate-pr:
+    name: 'Validation of Pull Request #${{ matrix.pr-number }}'
+    needs: [ gather-pull-requests ]
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


mfocko:
Just out of interest, what's the reason for using GH Action instead of the GitHub's built-in functionality?

At the same time, from the reviewer's PoV, if I review the PR and click on the big green button denoting _merge when ready_, it has basically the same effect, cause in the branch protection rules I can require specific CI checks to pass, and also there cannot be any merge conflicts.

jamacku:
> Just out of interest, what's the reason for using GH Action instead of the GitHub's built-in functionality?

This is a valid question. We (rhel systemd) had this automation before GitHub introduced all the Auto-merge stuff. We have other automation pieces related to Jira and Upstream that work together with auto-merge functionality. It allows us to have more customization and flexibility regarding the behavior. For example, we can waive the CI, all CI checks block us, and we pick non-blocking checks; when PR is merged, we transition Jira into a new state, etc.

For the copr project, it might be sufficient to use rules and auto-merge. But I know the copr project has a policy about merging approved PRs only after 24 hours from submission. This is currently not possible by using only rules. It is currently not possible by our action either, but it can be easily extended to support this requirement.

praiskup:
I'm pretty happy with the approach we currently have, there are no delays in `fedora-copr/copr` project caused by manual merges; getting a proper review is much more expensive than hitting "rebase and merge" button (if so, I'm completely fine to resolve concerns).

Also, somewhat nitpicking - I like if "Committer: " value != bot account.

Then there are often situations when somebody reviews, and since we are not in hurry and we don't merge immediately - another team member has a chance to review even after more than 24 hours (once merged, nobody reviews).

Note though that this is an opinion of just one team-member, others may feel differently, so let's give them a chance to speak.
Also note that automatic merges after ~24h or maybe even  ~7days would be **very welcome** in other, rather low-traffic projects (this one though in particular is under a pedantic set of humans' radars).

praiskup:
We discussed this yesterday on the team meeting; summary is that the team seems to be **for this change**. I see no benefits, just risks, and path to a code-quality decrease (maybe maintenance problems). I'm confident with my :-1: vote, sorry.

One of the ideas was to have a rule of "at least two reviewers", would that be possible?

 @nikromen promised to make a proper review.

nikromen (approved):
I say let's try it and see, mayble we'll like it, maybe not :D but yup let's discuss this on planing

praiskup:
Two post-merge remarks:
- the tool seems to label all the pull requests with its own labels, not sure it is anyhow useful, but we'll see long term
- what is very bad is that it "touches" each PR periodically, bumping the timestamp; when I sort the PRs per `Recently updated`, it seems all the PRs are updated all at the same time (so I don't see which PR got recent updates by human)
Screenshot_20250218_113752

jamacku:
Thank you Pavel for the feedback.

>     * the tool seems to label all the pull requests with its own labels, not sure it is anyhow useful, but we'll see long term

The labels could be disabled if you don't find them useful.

>     * what is very bad is that it "touches" each PR periodically, bumping the timestamp; when I sort the PRs per `Recently updated`, it seems all the PRs are updated all at the same time (so I don't see which PR got recent updates by human)

This is new to me. I have never used sorting by recently updated, but I will try to improve this so it won't touch PRs that frequently.

Issue #3328, opened 2024-07-12
"Batch-build after" doesn't take effect when doing Packages -> Rebuild all


WOnder93:
The builds just start right away, which is a bug.

praiskup:
Indeed, thank you for the report.

PR #3636, opened 2025-02-18
frontend: use after/with-build-id fields when "rebuilding all"

praiskup:
Fixes: #3328

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -798,11 +798,19 @@ class BuildsLogic(object):
         return build
 
     @classmethod
-    def _setup_batch(cls, batch, after_build_id, with_build_id, user):
-        # those three are exclusive!
-        if sum([bool(x) for x in
-                [batch, with_build_id, after_build_id]]) > 1:
-            raise BadRequest("Multiple build batch specifiers")
+    def setup_batch(cls, after_build_id, with_build_id, user,
+                    always_create=False):
+        """
+        According to "after_build_id" and "with_build_id" arguments (they are
+        mutually exclusive), get a pre-existing (or create a new) batch and
+        return it.  If neither of those args is specified, None is returned,
+        unless always_create=True - then a completely separate batch is crated
+        and returned.
+        """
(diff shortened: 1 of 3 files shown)


frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -798,11 +798,19 @@ def create_new(cls, user, copr, source_type, source_json, chroot_names=None, pkg
...
+                    always_create=False):
+        """
+        According to "after_build_id" and "with_build_id" arguments (they are
+        mutually exclusive), get a pre-existing (or create a new) batch and
+        return it.  If neither of those args is specified, None is returned,
+        unless always_create=True - then a completely separate batch is crated
+        and returned.
+        """
+        batch = None
+        if bool(with_build_id) and bool(after_build_id):
```
nikromen, line 811: Are the bools here necessary? They are just numbers or None so the if statement should work either way

e.g.:

```
>>> bool(1 and None)
False
>>> bool(1 and 1)
True
>>> bool(None and None)
False
>>>
```

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -798,11 +798,19 @@ def create_new(cls, user, copr, source_type, source_json, chroot_names=None, pkg
...
-                [batch, with_build_id, after_build_id]]) > 1:
-            raise BadRequest("Multiple build batch specifiers")
+    def setup_batch(cls, after_build_id, with_build_id, user,
+                    always_create=False):
+        """
+        According to "after_build_id" and "with_build_id" arguments (they are
+        mutually exclusive), get a pre-existing (or create a new) batch and
+        return it.  If neither of those args is specified, None is returned,
+        unless always_create=True - then a completely separate batch is crated
+        and returned.
```
nikromen, line 808: thanks for the docs :pray:

FrostyX (approved):

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: updated

Issue #3483, opened 2024-10-17
Fedora instance: please switch authentication from OpenID to OIDC


abompard:
Hey folks! The Fedora instance of Copr is currently using OpenID (and GSSAPI) for authentication.
We are looking to remove OpenID from the authentication options, because we'd like to switch the authentication provider from Ipsilon to Keycloak, which only supports OIDC.
If I understand correctly, Copr is already capable of OIDC authentication. Would it be possible to switch Copr's authentication to OIDC? (still with Ipsilon for now)

You'll need the following info:
- provider metadata: https://id.fedoraproject.org/openidc/.well-known/openid-configuration
- client_id: copr
- client_secret: just use the `copr_oidc_prod_client_secret` and `copr_oidc_stg_client_secret` ansible variables.

And I'll need the `redirect_uri` that you're going to use in the OIDC process.

I'm happy to help with the switch, ping me on Matrix (I'm in #buildsys and #infrastructure and #apps)

praiskup:
Thank you for the report.

Does the removal of OpenID mean removal of GSSAPI?

abompard:
That seems like a different thing, no? Is there a link between OpenID authentication and GSSAPI authentication in Copr ?
Does your GSSAPI auth go trough Ipsilon?

praiskup:
What is the ETA for killing OpenID in Fedora?  Do you have some tracker?

abompard:
> Probably **it is** a different thing? And I hope. :) you will know better than me, that's why I am asking 😅 "plain" GSSAPI is supported both in Copr cli and web-ui separately, plus gssapi is also accepted through Ipsilon (OID).

Yeah so it's independant from Ipsilon/OpenID :-)

> What is the ETA for killing OpenID in Fedora? Do you have some tracker?

We haven't decided on that yet, but we would prefer migrating all apps to OIDC as soon as reasonably possible, if they are already capable of it.

praiskup:
I have been pinged in https://pagure.io/fedora-infrastructure/issue/10241 , that seems lika a tracker.

The overall Fedora Copr movement is not a high priority task, though.  Do you think we should hurry up?

nirik:
Well, the only things left using openid (that we directly know of) are copr and anitya. There's going to be a new anitya release soon and then copr will be the last one. ;)

We want to move all these so we can retire ipsilon and move to keycloak, so sooner would be better for us.

praiskup:
[...]

PR #3642, opened 2025-02-24
frontend: change order of login menu, sign up should be the latest

nikromen:
changes the login menu from 

menu1

to

menu2

to keep the logins together

Relates to #3483

### DIFF
--- a/frontend/coprs_frontend/coprs/context_processors.py
+++ b/frontend/coprs_frontend/coprs/context_processors.py
@@ -56,6 +56,12 @@ def login_menu():
                 'desc': 'log in',
             })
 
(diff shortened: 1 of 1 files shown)


FrostyX (approved):

praiskup:
While on it, can you move the "OIDC login" link more "left, closer to the 'log in' option"?  Practically, we do not plan to have enabled both at the same time, so it basically replaces the normal OID login.  And it deserves to be the first and preferred method.

nikromen:
done

praiskup (approved):

praiskup:
Seems like hot-fixed; adding label.

PR #3225, opened 2024-04-16
frontend, backend: run rawhide-to-release for all CoprDirs

FrostyX:
Fix #3016

PR #3649, opened 2025-03-01
frontend: 404 when build_chroot is not found in database from task_id

### LINKED ISSUE #3457: Error 500 instead of 404?
```
ERROR:coprs:Admin-only exception
Request: GET https://copr.fedorainfracloud.org/backend/get-build-task/6742605-fedora-39-aarch64:
User: None
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/views/backend_ns/backend_general.py", line 368, in get_build_task
    task = BuildsLogic.get_build_task(task_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/logic/builds_logic.py", line 463, in get_build_task
    build_chroot = BuildChrootsLogic.get_by_build_id_and_name(build_id, chroot_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/logic/builds_logic.py", line 1501, in get_by_build_id_and_name
    mc = MockChrootsLogic.get_from_name(name).one()
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/sqlalchemy/orm/query.py", line 2870, in one
    return self._iter().one()
           ^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/result.py", line 1522, in one
    return self._only_one_row(
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/result.py", line 562, in _only_one_row
    raise exc.NoResultFound(
sqlalchemy.exc.NoResultFound: No row was found when one was required
ERROR:coprs:Response error: 500 Database error, contact admin
```

nikromen:
Addresses concern raised in
https://github.com/fedora-copr/copr/issues/3457#issuecomment-2692289898

Providing nonexisting chroot will result in NoResultFound too.
This change returns 404 instead of 500.

Fix #3457

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -460,8 +460,8 @@ class BuildsLogic(object):
         except ValueError:
             raise MalformedArgumentException("Invalid task_id {}".format(task_id))
 
-        build_chroot = BuildChrootsLogic.get_by_build_id_and_name(build_id, chroot_name)
         try:
+            build_chroot = BuildChrootsLogic.get_by_build_id_and_name(build_id, chroot_name)
             return build_chroot.join(models.Build).one()
         except NoResultFound as ex:
             raise ObjectNotFound("Specified task ID not found") from ex
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -365,10 +365,11 @@ def pending_jobs():
     return streamed_json(_stream())
 
 
-@backend_ns.route("/get-build-task/<task_id>/")
-@backend_ns.route("/get-build-task/<task_id>")
-def get_build_task(task_id):
+@backend_ns.route("/get-build-task/<int:build_id>-<chroot>/")
+@backend_ns.route("/get-build-task/<int:build_id>-<chroot>")
+def get_build_task(build_id, chroot):
     try:
+        task_id = f"{build_id}-{chroot}"
         task = BuildsLogic.get_build_task(task_id)
         build_record = get_build_record(task)
         return flask.jsonify(build_record)


nikromen:
tested on stg and it returns 404 as it should instead of 500:

```
$ curl -w "Code: %{http_code}\n" -X GET https://copr.stg.fedoraproject.org/backend/get-build-task/8711675-fedora-43-x86_64
{"msg":"Specified task ID not found"}
Code: 404
```

also instead of messy log, we get in logs

```
2025-03-03 17:55:51,010 [ERROR][/usr/share/copr/coprs_frontend/coprs/error_handlers.py:42|error_handlers:handle_error][ANON] Response error: 404 The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.
```

FrostyX:
Thank you for testing @nikromen,
but I am still worried. Sorry, I couldn't point you to the exact issue on the meeting.

There are a lot of layers of abstractions on the backend, but IMHO we will hit this error handling in case something is wrong:

https://github.com/fedora-copr/copr/blob/e81a5241a4c7c69957dbb42268db6afe23eae316/common/copr_common/request.py#L84-L96

And it behaves differently for `4xx` and `5xx` status codes. So I really wanted to make sure that backend won't start doing anything crazy once it wants to build something and we now return a different status code.

nikromen:
Ok, I can see your point now.

tldr;let's close this PR since it works, the only inconvenience is a rare occasion of bad HTTP code in error logs

Honestly, the original error is a bit of a mystery for me then :D I originally thought the build https://copr.fedorainfracloud.org/coprs/petersen/haskell-language-server/build/6742605/ was happening during f39 branching and thus we got the database error (which should be addressed by https://github.com/fedora-copr/copr/pull/3225 I think)... but looking to copr's branching history - F39 was branched on Aug 11, 2023 so it's some weird glitch :/

Anyway, yes, with the change in this PR, the task above would fail at the backend side - since it was successful after the second retry (see the logs of the f39-aarch build) and because `BuildsLogic.get_by_build_id_and_name` is used only for backend-related code right now it should really return 5xx in this case...

My issue with this is - we use the BuildsLogic and CoprsLogic methods not only for backend-related code but, in general, everywhere else - even inside `apiv3_ns` when parsing users' input... and sometimes we return or parse 4xx error, sometimes 5xx. This is hard for the maintainer to process when the methods return 4xx and when 5xx to use them for user or backend-related tasks (e.g. adding/editing some functionality and using the methods can result in different retval codes than expected).

But to add more headache to this, it would be hard to change the behaviour so backend receives 5xx when it should, and the rest gets 4xx, meanwhile not to break anything else :D And since what we have now works fine and nobody complains (except confusing HTTP code in our logs), I don't want to touch it anymore :D

WDYT about not accepting this change and closing the issue that we are fine with it?

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -460,8 +460,8 @@ def get_build_task(cls, task_id):
         except ValueError:
             raise MalformedArgumentException("Invalid task_id {}".format(task_id))
 
-        build_chroot = BuildChrootsLogic.get_by_build_id_and_name(build_id, chroot_name)
         try:
+            build_chroot = BuildChrootsLogic.get_by_build_id_and_name(build_id, chroot_name)
```
praiskup, line 464: This is a valid code movement.

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -1498,13 +1498,18 @@ def new(cls, build, mock_chroot, **kwargs):
...
-        )
+            return (
+                BuildChroot.query
+                .filter(BuildChroot.build_id == build_id)
+                .filter(BuildChroot.mock_chroot_id == mc.id)
+            )
+        except DataError as ex:
+            raise MalformedArgumentException(
+                f"Invalid build_id: {build_id} or name: {name}"
+            ) from ex
```
praiskup, line 1512: I'm not sure where the DataError comes from.  Do we *really* need this part at all?

nikromen: reproducer: https://copr.fedorainfracloud.org/backend/get-build-task/6742s605-fedora-39-aarch64

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: true, that's maybe the only thing I would want to accept right now from this PR

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: Are you sure?  If I apply the change above, I'm getting this locally:
```
curl http://127.0.0.1:55555/backend/get-build-task/6742s605-fedora-39-aarch64
{"msg":"Specified task ID not found"}
```

nikromen:
```
curl https://copr.fedorainfracloud.org/backend/get-build-task/871s1675-fedora-42-x86_64
```

produces in logs:

```
sqlalchemy.exc.DataError: (psycopg2.errors.InvalidTextRepresentation) invalid input syntax for type integer: "871s1675"
2025-03-17 16:32:04,543 [ERROR][/usr/share/copr/coprs_frontend/coprs/error_handlers.py:42|error_handlers:handle_error][ANON] Response error: 9h9h Database error, contact admin
```

the same goes for staging instance

praiskup: Right, I can reproduce now when f39 aarch64 exists.

praiskup:
This whole `/backend/` status isn't nice, I am not comfortable with the old code :shrug:  we shouldn't e.g. send exceptions over the wire, briefly discussed in https://github.com/fedora-copr/copr/pull/3626 .  This deserves some non-trivial reorg, which we don't have time for.

The exception handling dance in this PR, (re)raising, etc., doesn't make the situation better.  Would you consider something like this instead?
```python
-@backend_ns.route("/get-build-task/<task_id>/")
-@backend_ns.route("/get-build-task/<task_id>")
-def get_build_task(task_id):
+@backend_ns.route("/get-build-task/<int:build_id>-<chroot>/")
+@backend_ns.route("/get-build-task/<int:build_id>-<chroot>")
+def get_build_task(build_id, chroot):
+    task_id = f"{build_id}-{chroot}"
```

praiskup (approved):

Issue #3644, opened 2025-02-24
Message for 404 project does not exist is badly decoded


nikromen:
e.g. https://copr.fedorainfracloud.org/coprs/mcjesus/scores

![Image](https://github.com/user-attachments/assets/c1bfe2cb-19d1-4e03-b1b9-354b32ba255f)

the bytes are printed as raw string

Issue #3072, opened 2023-12-20
ISE - sqlalchemy.exc.NoResultFound: No row was found when one was required


xsuchy:
I see this in production log:
```
[Tue Dec 19 18:38:26.705835 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276] --- Logging error ---
[Tue Dec 19 18:38:26.709463 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276] Traceback (most recent call last):
[Tue Dec 19 18:38:26.709782 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]   File "/usr/share/copr/coprs_frontend/coprs/logic/complex_logic.py", line 201, in get_copr
[Tue Dec 19 18:38:26.709798 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]     return CoprsLogic.get(user_name, copr_name, **kwargs).filter(Copr.group_id.is_(None)).one()
[Tue Dec 19 18:38:26.709807 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[Tue Dec 19 18:38:26.709827 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/orm/query.py", line 2870, in one
[Tue Dec 19 18:38:26.709837 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]     return self._iter().one()           
[Tue Dec 19 18:38:26.709845 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]            ^^^^^^^^^^^^^^^^^^
[Tue Dec 19 18:38:26.709861 2023] [wsgi:error] [pid 2595633:tid 2595638] [remote 18.236.247.178:2276]   File "/usr/lib64/python3.12/site-packages/sqlalchemy/engine/result.py", line 1522, in one
[...]

PR #3076, opened 2023-12-20
convert eror message to binary text before emitting

xsuchy:
The query comes from users and does not need to conform the regexp that we uses when creating a project. In fact the string may contain unicode characters that cannot be converted to ascii stream.

Fixes: #3072

Addressing:

```
Traceback (most recent call last):
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1823, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/views/misc.py", line 244, in wrapper
    copr = ComplexLogic.get_copr(username, coprname, with_mock_chroots=True)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/logic/complex_logic.py", line 203, in get_copr
    raise ObjectNotFound(
coprs.exceptions.ObjectNotFound: Project t0xic0der/nvid\xe2\x80\xa6 does not exist. During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/lib64/python3.12/logging/__init__.py", line 1163, in emit
    stream.write(msg + self.terminator)
UnicodeEncodeError: 'ascii' codec can't encode character '\\u2026' in position 167: ordinal not in range(128) Call stack:
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 2552, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 2529, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1825, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/lib/python3.12/site-packages/flask_restx/api.py", line 674, in error_router
    return original_handler(e)
  File "/usr/lib/python3.12/site-packages/flask/app.py", line 1676, in handle_user_exception
[...]

PR #3651, opened 2025-03-03
Revert "convert eror message to binary text before emitting" 

nikromen:
This reverts commit https://github.com/fedora-copr/copr/commit/57fcaf089b1a3e6f5f6aa25ebc67f856f996f631.

https://github.com/fedora-copr/copr/pull/3651#issuecomment-2710932004
explains why probably this issue was resolved and we no longer need this
fix. Also this commit caused bad formatting - b' prefix and 'suffix.
If similar issue like https://github.com/fedora-copr/copr/issues/3072 appears, we should probably parse the string
and leave the unparsable characters as their byte value.
See https://github.com/fedora-copr/copr/commit/4c8b3ec5bdc9965def756c95879a451510154cc3

Fix https://github.com/fedora-copr/copr/issues/3644

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/complex_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/complex_logic.py
@@ -204,7 +204,7 @@ class ComplexLogic(object):
             raise ObjectNotFound(
                 message="Project @{}/{} does not exist.".format(
                     group_name, copr_name
-                ).encode()
+                )
             ) from exc
 
     @staticmethod
@@ -219,7 +219,7 @@ class ComplexLogic(object):
             raise ObjectNotFound(
                 message="Project {}/{} does not exist.".format(
                     user_name, copr_name
-                ).encode()
+                )
             ) from exc
 
     @staticmethod


nikromen:
the https://github.com/fedora-copr/copr/pull/3651/commits/043086c6a407f62e05c200a04f276760a2e94c9a applies second approach - it tries to parse the message to utf-8, but it is stored as a string nevertheless... this way it works around for every case some user puts weird characters as any input in API. 

e.g. https://copr.fedorainfracloud.org/coprs/g/<nasty_characters> -> the non-parsable characters will be parsed as their byte representation but as string (`some\x01\x02string`)

nikromen:
aha, it is true that utf-8 characters like "â€¦" work now, you are right.

I've been researching why this is, and I've concluded that we must have had the locale set to US.US-ASCII or something similar on the old F39 machines at the time. Now we have US.UTF-8. That's probably why this error used to occur, because logging uses by default [io.text_encoding](https://docs.python.org/3/library/io.html#io.text_encoding), which defaults to [Python's UTF-8 mode](https://docs.python.org/3/library/os.html#utf8-mode) and returns either locale or UTF-8 if the locale is not set. In that case it was trying to log to ASCII instead of UTF-8.

In any case, it just reduced the set of characters on which this error occurs. Non-UTF-8 characters (such as UTF-16) will fall with the same error.

So there are two possibilities:
  - accept this PR as a solution that reverses https://github.com/fedora-copr/copr/pull/3076 while outputting all non-UTF-8 characters as a pure string of their byte value
  - remove the parsing part from this PR and just revert https://github.com/fedora-copr/copr/pull/3076 . If someone will send us requests with non-utf8 characters and we get errors, then do the parsing afterwards.
  
what do you think about it?

praiskup:
The old reproducer was:

```python
b'Project t0xic0der/nvid\xe2\x80\xa6 does not exist.'
```

That is in utf:

```
'Project t0xic0der/nvid… does not exist.'
```

This doesn't fail the logger on F41 (current prod).  It is ugly, yes, the logging code still dumps the message using the `\xe2\x80...` instead of `\u2026`, not sure why (does logging want ascii? if yes, where/why? because they don't want to ignore the encoding errors?)

But at least it doesn't try to encode elpisis into ASCII.

> When I simply revert the code from PR https://github.com/fedora-copr/copr/pull/3076, I get:

Yes, please, revert.  I tested with str=$(dd if=/dev/urandom bs=8 count=1) and I got:

```
[remote 77.92.220.242:48662] ERROR:coprs:Response error: 404 Project frostyx/\xef\xbf\xbd|\xef\xbf\xbd\xef\xbf\xbd|\xef\xbf\xbdO\x02 does not exist.
```

We should be safe;  no tracebacks.  But if we miss something, we'll at least get the reproducer.

nikromen:
> not sure why (does logging want ascii? if yes, where/why? because they don't want to ignore the encoding errors?)

<citing my comment above> I've concluded that we must have had the locale set to US.US-ASCII or something similar on the old F39 machines at the time. Now we have US.UTF-8. That's probably why this error used to occur, because logging uses by default [io.text_encoding](https://docs.python.org/3/library/io.html#io.text_encoding), which defaults to [Python's UTF-8 mode](https://docs.python.org/3/library/os.html#utf8-mode) and returns either locale or UTF-8 if the locale is not set. In that case it was trying to log to ASCII instead of UTF-8.

> We should be safe; no tracebacks. But if we miss something, we'll at least get the reproducer.

this will still mean that we got logging error when someone sends e.g. utf32 only characters

praiskup:
> I've concluded that we must have had the locale set to US.US-ASCII 

I honestly doubt that.  But something changed between F39 and F41 :+1:

> this will still mean that we got logging error when someone sends e.g. utf32 only characters

No.  How comes?  Can you elaborate (provide reproducer)?

nikromen:
> No. How comes? Can you elaborate (provide reproducer)?

I don't have utf16 encoding on my machine so I cant :/ but my logic behind is that if the logs are in utf8 what will happen if somebody will send something in utf16? This is the similar situation as the original issue:
we had logging only in ascii -> somebody sent utf8 character -> error
now we have utf8 -> somebody sends utf16 only char -> error

nikromen:
anyway I drop the str -> bytes -> str parsing code and only reverted https://github.com/fedora-copr/copr/commit/57fcaf089b1a3e6f5f6aa25ebc67f856f996f631 as recommended. PTAL

praiskup:
> my logic behind is that if the logs are in utf8 what will happen if somebody will send something in utf16

I wasn't able to reproduce the traceback with a random binary blob, so I doubt utf-16 would trigger the problem.This is 

> the similar situation as the original issue:

That's the issue;  I don't think we are able to reproduce the old issue, who knows what was happening back then...

praiskup (approved):

praiskup:
Thank you for the updated.

PR #3652, opened 2025-03-03
custom method docs fixes

praiskup:
I intentionally keep the explanation out, because it was misleading.  The underlying code generates source RPM itself, and transfers the SRPM to dist-git.   We don't need src.rpm though, hence the https://github.com/release-engineering/dist-git/pull/72  dance.

### DIFF
--- a/doc/custom_source_method.rst
+++ b/doc/custom_source_method.rst
@@ -77,13 +77,25 @@ Since this all is in user's hands, it is not technically incorrect to have empty
 hook payload, e.g. it is valid to call `curl -X POST <THE_CUSTOM_HOOK_URL>` to
 trigger the custom source build method.
 
-Src.RPM
--------
 
-Copr expects that `script` creates SPEC file, tar ball, and patches in the working
-directory. We cannot process SRC.RPM. Because some chroots can use technology
-which our server cannot recognize. E.g., in the past `rpm` changed compression and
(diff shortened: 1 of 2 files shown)


doc/custom_source_method.rst
```diff
@@ -77,13 +77,25 @@ Since this all is in user's hands, it is not technically incorrect to have empty
...
+
+Copr expects that when the provided `script` is executed, it provides a working
+SPEC file, source archive(s) referenced by Source statements (typically a
+tarball), and optionally patches referenced by Patch statements.  These files
+should be placed in the script’s `resultdir` directory (which defaults to the
+current working directory).
+
+This method does not accept pre-generated source RPMs.  If you already have an
+existing tool that generates one, you can use it in the script, but you must
+extract its contents within the script itself.  For example::
```
FrostyX, line 92: > This method does not accept pre-generated source RPMs

I understand the point, but it is a bit confusing what it could mean - the method accepts pre-generated SRPMs a in the sense that you can wget a SRPM and extract it.

I would go with something like:

"The output of this method cannot be a SRPM. If you already have one an ..."

praiskup: Very good point, I used this `The script's output in this method cannot be an SRPM.`

xsuchy (approved):
+1

PR #3654, opened 2025-03-03
frontend: fix SAWarnings about autoflush

nikromen:
Since there was a lot of these warnings, there are 3 kind of fixes to these:

- 1st -> Just adds objects created and manipulated in tests to session to silence warnings
- 2nd -> Shifting `db.session.add`: These warnings occur because SQLAlchemy tries to manage relationships affecting an object (e.g. CoprDir) that has not yet been added to the session, and a flush is triggered by some automatic magic inside SQLalchemy, hence the warning. Moving `session.add` before tinkering/adding relationships to it silence the warning
- 3rd -> If for some reason 2nd fix can't be applied due to complicated relationships creations, `no_autoflush` can be applied to tell SQLAlchemy not to automatically flush (we do it anyway with `db.session.flush()`)

The second solution was inspired by https://github.com/sqlalchemy/sqlalchemy/issues/4491#issuecomment-462091501 since this is similar situation

### DIFF
--- a/frontend/coprs_frontend/coprs/logic/builds_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/builds_logic.py
@@ -958,13 +958,13 @@ class BuildsLogic(object):
             isolation=isolation,
             ssh_public_keys=ssh_public_keys,
         )
+        db.session.add(build)
 
         if timeout:
             build.timeout = timeout
         else:
             build.timeout = package.timeout if package else app.config["DEFAULT_BUILD_TIMEOUT"]
 
-        db.session.add(build)
         cls.assign_buildchroots(
             build,
             chroots,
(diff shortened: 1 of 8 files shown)


frontend/coprs_frontend/coprs/logic/coprs_logic.py
```diff
@@ -285,6 +285,7 @@ def get_multiple_fulltext(cls, fulltext=None, projectname=None,
         return query
 
     @classmethod
+    @no_autoflush
```
praiskup, line 288: This disables just one warning :shrug:

praiskup:
There are still a few of similar issues waiting for the fix.  Do you plan to fix them here, or in a new PR?

praiskup:
> Do you plan to fix them here, or in a new PR?

I have looked;  I have a work-in-progress patch.  Going to submit as a separate PR ASAP.

frontend/coprs_frontend/coprs/logic/coprs_logic.py
praiskup: I think we can drop the `no_autoflush` entirely in favor of https://github.com/fedora-copr/copr/pull/3699

PR #3657, opened 2025-03-06
Use CDN baseurl for Pulp projects

### LINKED ISSUE #3632: Use CDN baseurl for Pulp projects
In repofiles we use

```
baseurl=https://console.redhat.com/api/pulp-content/public-copr/@copr/pulp-test/fedora-$releasever-$basearch/
```

and we would like to use our CDN.

FrostyX:
Fix #3632

The whole frontend part of this PR is only a revert of a previously merged change.

### DIFF
--- a/backend/copr_backend/actions.py
+++ b/backend/copr_backend/actions.py
@@ -147,8 +147,8 @@ class Createrepo(Action):
 
     def add_http_redirect(self):
         """
-        Create a HTTP redirect for this project
-        See https://pagure.io/fedora-infra/ansible/blob/main/f/roles/copr/backend/templates/lighttpd/pulp-redirect.lua.j2
+        Create a HTTP redirect for this project.  See:
+        https://pagure.io/fedora-infra/ansible/blob/main/f/roles/copr/backend/templates/lighttpd/pulp-redirect.lua.j2
         """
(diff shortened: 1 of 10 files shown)


FrostyX:
The lint errors are in the reverted code. Not sure if I should fix them (and therefore alter the reverted changes) or rather keep them

nikromen (approved):
lgtm, let's merge it and see how it works on stg with the rest

FrostyX:
Unfortunately, this did break building in Pulp projects locally. It fails with

```
Failed to download metadata (baseurl: "http://backend_httpd:5002/results/frostyx/test-pulp-network-error-3/fedora-41-x86_64/") for repository "copr_base"
 Librepo error: Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
```

because obviously, the repo doesn't exist on backend.

I was about to change our `backend_httpd` container from nginx to lighttpd so that we can add the redirect, but the lua redirect script lives in the Fedora Infra ansible repo. We would have to either move or duplicate it. Alternatively, we could revert part of the reverted code back and have `PULP_CONTENT_URL` pointing to our CDN but locally pointing to the correct container.

Issue #2789, opened 2023-06-26
Provide users/groups via `/usr/lib/sysusers.d/`


praiskup:
This is going to be handled by RPM: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/NEFOV236FJYS2RED2SEOV5YHDFLDX7DK/#KB6KS3U7RO4AYANGSUNVSG7UF5AW52Q2

At the early stages (I think) `systemd` will be automatically installed as a runtime requirement (even if not necessarily needed by us).  But this in particular shouldn't be an issue for our use-cases.

FrostyX:
Triage: We need to revert a revert :D

xsuchy:
This should be revert of ad7551f0b98dcd094365af95827405ebb0e88485

praiskup:
https://github.com/rpm-software-management/mock/commit/ad7551f0b98dcd094365af95827405ebb0e88485

praiskup:
I believe we wanted to fix this in Copr packages.  Reopening.

praiskup:
This has been done downstream-only by @keszybz for all of our components, and because we use Tito it will be reverted next week because we plan to have release ASAP.  Please consider raising a priority, and use it as tracker:

- [ ] copr-backend
- [ ] copr-keygen
- [ ] copr-dist-git
- [ ] copr-frontend
- [ ] dist-git

nikromen:
what does this mean? this issue was originally to provide the sysuser macro for mock. How this relates to copr?

praiskup:
Dist-git (external project) remains.

PR #3663, opened 2025-03-12

xsuchy:
forwarded from downstream copr-frontend with added if-conditions
Related: https://github.com/fedora-copr/copr/issues/2789

### DIFF
--- a/frontend/conf/copr-frontend.sysusers.conf
+++ b/frontend/conf/copr-frontend.sysusers.conf
@@ -1,0 +1,2 @@
+u copr-fe - 'COPR frontend user' /usr/share/copr/coprs_frontend /bin/bash
+g  copr-fe -
--- a/frontend/copr-frontend.spec
+++ b/frontend/copr-frontend.spec
@@ -296,16 +296,12 @@ EOF
 %py_byte_compile %{__python3} %{buildroot}%{_datadir}/copr/coprs_frontend/coprs
 %py_byte_compile %{__python3} %{buildroot}%{_datadir}/copr/coprs_frontend/alembic
 
+install -m0644 -D conf/copr-frontend.sysusers.conf %{buildroot}%{_sysusersdir}/copr-frontend.conf
+
 %check
 %if %{with check}
 ./run_tests.sh -vv --no-cov
 %endif
-
-%pre
-getent group copr-fe >/dev/null || groupadd -r copr-fe
-getent passwd copr-fe >/dev/null || \
-useradd -r -g copr-fe -G copr-fe -d %{_datadir}/copr/coprs_frontend -s /bin/bash -c "COPR frontend user" copr-fe
-usermod -L copr-fe
 
 
 %post
@@ -360,7 +356,7 @@ usermod -L copr-fe
 %{_libexecdir}/copr_dump_db.sh
 %exclude_files flavor
 %exclude_files devel
-
+%{_sysusersdir}/copr-frontend.conf
 
 %files fedora
 %license LICENSE


frontend/copr-frontend.spec
```diff
@@ -296,16 +296,22 @@ EOF
 %py_byte_compile %{__python3} %{buildroot}%{_datadir}/copr/coprs_frontend/coprs
 %py_byte_compile %{__python3} %{buildroot}%{_datadir}/copr/coprs_frontend/alembic
 
+%if 0%{?fedora} && 0%{?fedora} > 41
```
praiskup, line 299: Are you sure we have to special-case F40?  If yes, Mock on F40 is broken :shrug:

nikromen: I think this adresses https://src.fedoraproject.org/rpms/copr-frontend/pull-request/3#comment-252345 ?

frontend/copr-frontend.spec
```diff
@@ -296,16 +296,22 @@ EOF
...
+install -m0644 -D conf/copr-frontend.sysusers.conf %{buildroot}%{_sysusersdir}/copr-frontend.conf
+%endif
+
 %check
 %if %{with check}
 ./run_tests.sh -vv --no-cov
 %endif
 
+%if 0%{?fedora} && 0%{?fedora} > 41
```
nikromen, line 308: did you mean 0%{?fedora} < 41 perhaps? since this way we will end without copr-fe user and group for f41 and less

frontend/conf/copr-frontend.sysusers.conf
```diff
@@ -0,0 +1,2 @@
+u! copr-fe - 'COPR frontend user' /usr/share/copr/coprs_frontend /bin/bash
```
nikromen, line 1: rpm 4.20.1 just landed to fedora41 (and will land to f42 in matter of hours) https://bodhi.fedoraproject.org/updates/?packages=rpm could you please then edit the conditions to include also f41?

also I tried this to install on rawhide yesterday and I ended installing it without copr-fe user or group... did you managed to get copr-fe user and group?

nikromen: also I tried this to install on rawhide yesterday and I ended installing it without copr-fe user or group... did you managed to get copr-fe user and group?

frontend/copr-frontend.spec
praiskup: Aha, good to document at least!  Anyway, in practice we do not use locked users.  We do things like `su - copr-fe` on a daily basis.

praiskup: I would keep the copr-fe user (and other PRs too) unlocked.  If we disagree on this, we can do `sed -i 's/u!/u/' copr-frontend.conf` if on `%if 0%{?fedora} && 0%{?fedora} <= 41`.

Otherwise, since we only support Fedora, the use of sysusers RPM feature should be unconditional.

frontend/copr-frontend.spec
```diff
@@ -296,16 +296,22 @@ EOF
...
 %pre
 getent group copr-fe >/dev/null || groupadd -r copr-fe
 getent passwd copr-fe >/dev/null || \
 useradd -r -g copr-fe -G copr-fe -d %{_datadir}/copr/coprs_frontend -s /bin/bash -c "COPR frontend user" copr-fe
 usermod -L copr-fe
+%endif
```
praiskup, line 314: And we can drop this manual user/group creation entirely.

frontend/copr-frontend.spec
xsuchy: Why have it unlocked? We have it locked now (see the line with `usermod -L copr-fe`) so this is no modification of current behaviour. You can still `su` to locked account. You just cannot log in (in terminal or ssh) to that account.

xsuchy: > Otherwise, since we only support Fedora, the use of sysusers RPM feature should be unconditional.

What about Euler? 

Additionally if we remove this condition that we have to add requires on specific version of `rpm` because on older version of Fedora you can have rpm from GA release that does not support this.

frontend/copr-frontend.spec
xsuchy: no. because for fedora <41 the old behaviour will be used.

frontend/conf/copr-frontend.sysusers.conf
xsuchy: yes. it worked for me.

frontend/copr-frontend.spec
praiskup: > Why have it unlocked? 

Because that's what we practically do (we create users deterministically via ansible, and execute e.g. `copr-frontend create-chroot` manually, etc.).

praiskup: > What about Euler?

Good point, @pkking how far is openEuler with [Rpm >= 4.19 support](https://rpm-software-management.github.io/rpm/manual/users_and_groups.html)?

praiskup: @pkking of course if you create users differently (not via an RPM installation, %post scriptlets, etc.), you don't need to have RPM 4.19 at all - then, whatever we do in this PR can not affect your openEuler deployments; nb. our Copr deployments are neither affected by this change.

praiskup: Hmm, now I read better, sorry - learning something new :sweat_smile: What is the practical effect of `usermod -L`?  I thought it basically does `/sbin/nologin`, but as I understand it now - it only ever makes users' passwords non-working (modifying etc/shadow).  If that's what it does, I don't think we need that at all (we don't create users with passwords).

xsuchy:
Discussed on mtg, that the old part will be removed and I will leave only the sysuser part. Unconditionally.

cc @pkking  any objections?

xsuchy:
Updated.

frontend/copr-frontend.spec
pkking: > @pkking of course if you create users differently (not via an RPM installation, %post scriptlets, etc.), you don't need to have RPM 4.19 at all - then, whatever we do in this PR can not affect your openEuler deployments; nb. our Copr deployments are neither affected by this change.

Hi, the latest openEuler LTS release is using the rpm `version 4.18`, the next LTS release will be at 2026/03, and yes, we run copr in k8s cluster and create user using a `useradd` command:
```shell
RUN groupadd -r copr-fe -g 997
RUN useradd -r copr-fe -u 997 -g 997 -d /usr/share/copr/coprs_frontend
```

frontend/copr-frontend.spec
```diff
@@ -296,17 +296,13 @@ EOF
...
 %py_byte_compile %{__python3} %{buildroot}%{_datadir}/copr/coprs_frontend/alembic
 
+install -m0644 -D conf/copr-frontend.sysusers.conf %{buildroot}%{_sysusersdir}/copr-frontend.conf
+
 %check
 %if %{with check}
 ./run_tests.sh -vv --no-cov
 %endif
 
-%pre
```
pkking, line 304: can you please explain why to remove the entire `%pre`?
that means downstream without `rpm > 4.19` should create user manually

praiskup: That's right.  Fedora 41+ is on RPM 4.20+, so that is non-issue for us - and since you pre-create usernames in kube deployments it should be no problem for openEuler Copr, too.

frontend/conf/copr-frontend.sysusers.conf
praiskup: I'm afraid we need to `s/u!/u/` to make this work on Fedora <= 41.

praiskup (approved):

PR #3664, opened 2025-03-12

xsuchy:
forwarded from downstream copr-dist-git with added if-conditions
Related: https://github.com/fedora-copr/copr/issues/2789


nikromen (commented):
lgtm, if you could just update the condition as mentioned in https://github.com/fedora-copr/copr/pull/3663#discussion_r2005443063 it would be nice!

praiskup (changes requested):
We should sync with #3663

xsuchy:
Updated.

praiskup (approved):

PR #3667, opened 2025-03-13
unify bindir and sbindir

xsuchy:
In Fedora 42 /usr/bin and /usr/sbin has been merged. We do not have real reason to insist on /usr/sbin and we can do this change for older distributions too.

Fixes: #3635

Addressing:
RPM build warnings:
    File listed twice: /usr/bin/copr-assure-permissions
    File listed twice: /usr/bin/copr-aws-s3-hitcounter
    File listed twice: /usr/bin/copr-backend-analyze-results
    File listed twice: /usr/bin/copr-backend-generate-graphs
    File listed twice: /usr/bin/copr-backend-process-action
    File listed twice: /usr/bin/copr-backend-process-build
    File listed twice: /usr/bin/copr-backend-resultdir-cleaner
    File listed twice: /usr/bin/copr-backend-service
    File listed twice: /usr/bin/copr-backend-unknown-resalloc-tickets.py
    File listed twice: /usr/bin/copr-change-storage
    File listed twice: /usr/bin/copr-compress-live-logs
    File listed twice: /usr/bin/copr-rename-chroot
    File listed twice: /usr/bin/copr-repo
    File listed twice: /usr/bin/copr-run-dispatcher-backend
    File listed twice: /usr/bin/copr_find_obsolete_builds.sh
    File listed twice: /usr/bin/copr_find_wrong_chroot_artifacts.py
    File listed twice: /usr/bin/copr_fix_gpg.py
    File listed twice: /usr/bin/copr_log_hitcounter.py
    File listed twice: /usr/bin/copr_print_results_to_delete.py
    File listed twice: /usr/bin/copr_prune_results.py
    File listed twice: /usr/bin/copr_prune_srpms.py
    File listed twice: /usr/bin/copr_run_logger.py
    File listed twice: /usr/bin/copr_sign_unsigned.py
    File listed twice: /usr/bin/print_queues.py

PR #3668, opened 2025-03-13

xsuchy:
forwarded from downstream copr-backend with added if-conditions
Related: #2789

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -108,6 +108,8 @@ Requires(post): systemd
 Requires(preun): systemd
 Requires(postun): systemd
 
+%{?fedora:Requires(pre): group(lighttpd)}
+
 %description
 COPR is lightweight build system. It allows you to create new project in WebUI,
 and submit new builds and COPR will create yum repository from latest builds.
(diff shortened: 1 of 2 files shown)


xsuchy:
This built on top of https://github.com/fedora-copr/copr/pull/3667 that should be merged before anyone tries to review this one.

nikromen (commented):
lgtm, if you could just update the condition as mentioned in https://github.com/fedora-copr/copr/pull/3663#discussion_r2005443063 it would be nice!

also could you please rebase so we don't have dupe of https://github.com/fedora-copr/copr/pull/3667 in this PR?

praiskup (changes requested):
We should sync with #3663

backend/copr-backend.spec
```diff
@@ -185,15 +188,20 @@ cp -a docs/build/html %{buildroot}%{_pkgdocdir}/
...
 install -p -m 644 copr-backend-resultdir-cleaner.1 %{buildroot}/%{_mandir}/man1/
 
+%if 0%{?fedora} && 0%{?fedora} > 41
+install -m0644 -D conf/copr-backend.sysusers.conf %{buildroot}%{_sysusersdir}/copr-backend.conf
+%endif
 
 %check
 ./run_tests.sh -vv --no-cov
 
+%if 0%{?fedora} && 0%{?fedora} < 42
```
pkking, line 198: So this line will omit the user creation also in openEuler and RHEL?
I think we still need the code block in this scenario

praiskup: The existing Copr instances (four driven by us, one driven by you) - these all prepare the users/groups manually, so it is not an issue.  If this causes any problems to real deployments, we may consider partial revert.

xsuchy:
updated.

praiskup (approved):

backend/copr-backend.spec
```diff
@@ -108,6 +108,8 @@ Requires(post): systemd
 Requires(preun): systemd
 Requires(postun): systemd
 
+Requires(pre): group(lighttpd)
```
praiskup, line 111: Don't we need the %{?fedora:} condition here?

praiskup (changes requested):
the requires(pre) needs to be in %{?fedora:} condition

backend/copr-backend.spec
xsuchy: Why?
We will always need a package that provides this group.

praiskup: I actually had to install this to make the release happen :-( ee171546d58c7d21cf0a9e57962a049753df703c

Issue #3685, opened 2025-03-31
Make sure we release client packages into EPEL 10 branched + stable


praiskup:
https://github.com/fedora-copr/copr/blob/7b0da1c11588d73b33d0cab0a7d2444d35ec35a0/.tito/releasers.conf#L8

We should be using epel-all (that should include both epel10 and epel10.0).

PR #3686, opened 2025-03-31
releng: release clients into all EPEL chroots

praiskup:
This includes EPEL 10.0 and EPEL 10.1 (next) at this point in time.

Resolves: #3685


praiskup:
Self-merging to get a new commit in main, so I can test PULP builds (we normally push these changes without PRs anyway).

PR #3688, opened 2025-04-01
frontend: drop no longer unused PULP_CONTENT_URL config

praiskup:
This config option was just confusing in Fedora's ansible.git repo, and also here upstream.


FrostyX:
I think we will actually need this to fix the container in #3632

praiskup:
See also #3731

PR #3716, opened 2025-05-01
Use ThreadPoolExecutor to parallelize RPM uploads to Pulp.

dkliban:

### DIFF
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -2,6 +2,7 @@
 Support for various data storages, e.g. results directory on backend, Pulp, etc.
 """
 
+from concurrent.futures import ThreadPoolExecutor, as_completed
 import os
 import json
 import shutil
@@ -58,7 +59,7 @@ class Storage:
         """
         raise NotImplementedError
 
-    def upload_build_results(self, chroot, results_dir, target_dir_name):
+    def upload_build_results(self, chroot, results_dir, target_dir_name, max_workers=1):
         """
         Add results for a new build to the storage
         """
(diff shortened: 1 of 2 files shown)


backend/copr_backend/storage.py
```diff
@@ -216,42 +217,63 @@ def init_project(self, dirname, chroot):
...
+        # This involves a lot of unnecessary waiting until every
+        # RPM content is created. Once we can reliably label Pulp
+        # content with Copr build ID, we should drop this code and stop
+        # creating the `pulp.json` file
+        task = response.json()["task"]
+        response = self.client.wait_for_finished_task(task)
+        return response
+
     def upload_build_results(self, chroot, results_dir, target_dir_name):
+        max_workers = min(32, (os.cpu_count() or 1) * 5) # This could probably be hardcoded
```
praiskup, line 237: This would be `min(32, 80)` in our case, and we can expect ~100+ processes uploading stuff to Pulp, i.e. 3200 parallel threads.

praiskup:
Thank you, @dkliban, awesome contribution!  I just think we should limit the concurrency "by default" (parallelize just the migration script).  Otherwise looks good.

dkliban:
@praiskup I updated the PR with your suggestions.

backend/copr_backend/storage.py
```diff
@@ -216,42 +217,62 @@ def init_project(self, dirname, chroot):
...
         return response.ok
 
-    def upload_build_results(self, chroot, results_dir, target_dir_name):
+    def upload_rpm(self, repository, path):
+        response = self.client.create_content(repository, path)
+
+        if not response.ok:
+            self.log.error("Failed to create Pulp content for: %s, %s",
+                           path, response.text)
+            return
```
praiskup, line 226: Linter is right; we should probably return failed response to be able to uniformly read it with others.

praiskup:
Ideally please provide the docstrings (our preferred style), or prefix "internal" methods with `_`.  One of those makes linter happy.

backend/copr_backend/storage.py
dkliban: I'll update it

backend/copr_backend/storage.py
```diff
@@ -58,10 +59,17 @@ def init_project(self, dirname, chroot):
...
+        """
+        Add an RPM to the storage
+        """
+        raise NotImplementedError
+
+    def upload_build_results(self, chroot, results_dir, target_dir_name, max_workers=1):
         """
         Add results for a new build to the storage
         """
+        raise NotImplementedError
```
praiskup, line 72: I don't think we can afford this exception here;  it was meant to be a no-op for the other storage.

praiskup (approved):

PR #3720, opened 2025-05-13
Adds build_id pulp_label to RPMs uploaded to Pulp.

dkliban:


FrostyX:
Oh, this is epic. I didn't test how well it works, but it looks like exactly what I wanted to do but didn't know how.

praiskup (approved):
Thank you!

PR #3722, opened 2025-05-13
Remove by build id

dkliban:


dkliban:
@praiskup @FrostyX I left the code that waits for upload tasks to finish. I can remove it if that is your preference.

PR #3723, opened 2025-05-15
backend: more log "checkpoints" for build worker

praiskup:
There was an outage when it was rather hard to realize that resalloc API isn't responding.  These new log entries should help with this.  The outage was caused by https://github.com/praiskup/resalloc/issues/173


FrostyX (approved):

xsuchy (approved):
+1

PR #3729, opened 2025-05-20
Small fixes for Pulp labels

FrostyX:
This should fix our failing beaker tests and also make sure we won't have inconsistent labels because of leading zeros.

### DIFF
--- a/backend/run/copr-change-storage
+++ b/backend/run/copr-change-storage
@@ -149,14 +149,15 @@ def main():
             for builddir_entry in os.scandir(chrootdir):
                 if not builddir_entry.is_dir():
                     continue
-                builddir = builddir_entry.name
-                build_id = builddir.split('-')[0]
-                resultdir = os.path.join(chrootdir, builddir)
 
+                builddir = builddir_entry.name
+                resultdir = os.path.join(chrootdir, builddir)
 
                 if not is_valid_build_directory(builddir):
                     log.info("Skipping: %s", resultdir)
                     continue
+
+                build_id = str(int(builddir.split("-")[0]))
 
                 # TODO Fault-tolerance and data consistency
                 # Errors when creating things in Pulp will likely happen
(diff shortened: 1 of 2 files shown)


dkliban (approved):

backend/run/copr-change-storage
```diff
@@ -150,7 +150,7 @@ def main():
                 if not builddir_entry.is_dir():
                     continue
                 builddir = builddir_entry.name
-                build_id = builddir.split('-')[0]
+                build_id = builddir.split('-')[0].lstrip("0")
```
praiskup, line 153: Even though, we could use `str(int(...))` pattern to make the script fail in case of unexpected value.

praiskup (approved):

backend/run/copr-change-storage
FrostyX: This actually caught an error!

```
Uploaded to Pulp: hello-2.8-1.fc41.src.rpm
Uploaded to Pulp: hello-2.8-1.fc41.x86_64.rpm
Uploaded to Pulp: hello-debuginfo-2.8-1.fc41.x86_64.rpm
Uploaded to Pulp: hello-debugsource-2.8-1.fc41.x86_64.rpm
OK: /var/lib/copr/public_html/results/frostyx/test-pulp-20/fedora-41-x86_64/08561800-hello
  File "/opt/copr/backend/run/copr-change-storage", line 217, in <module>
    main()
    ~~~~^^
  File "/opt/copr/backend/run/copr-change-storage", line 153, in main
    build_id = str(int(builddir.split("-")[0]))
                   ~~~^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: 'repodata'
```

FrostyX: It wouldn't be a problem in production because we `if not is_valid_build_directory(builddir)` just below this line and the directory would be skipped but I moved it up.

praiskup (approved):

PR #3754, opened 2025-06-15
backend: fix waiting for Pulp copr_base repository

FrostyX:
Locally, we use the same Pulp hostname for accessing its API and its results. However, in production and on STG we use https://mtls.internal.console.redhat.com to access the Pulp API, and https://console.redhat.com/api/pulp-content/public-copr-stage/ to fetch the results.

With @dkliban we discovered that this code for checking the availability of a Pulp repository returns 403 error because we are asking on a wrong URL.


FrostyX:
CI fails only in Rawhide for unrelated reasons.

PR #3756, opened 2025-06-18
frontend: after edit, redirect to the package

### LINKED ISSUE #3511: [RFE] After editing package in Copr, redirect back to the package page at /package/foo instead of generic /packages/
I want to edit a package's default build method. Click Edit. Go there: `https://copr.fedorainfracloud.org/coprs/g/python/python3.14/package/python-setuptools/edit/distgit`. Change what I want. Click Save.

Now:
I am redirected to https://copr.fedorainfracloud.org/coprs/g/python/python3.14/packages/

Ideally:
I'd like to be redirected to that package's page: https://copr.fedorainfracloud.org/coprs/g/python/python3.14/package/python-setuptools

Rationale: in 99% of cases I want to spawn a build right after that.

FrostyX:
Fix #3511


praiskup (approved):
thank you!

PR #3761, opened 2025-06-19
backend: make PULP work more verbose in logs

praiskup:
Relates: #3689


praiskup:
@dkliban can you confirm this is safe to log publicly?  These log entries are (in most of the cases) exposed to end-users.

PR #3763, opened 2025-06-19
frontend: show coprdirs in the builds table

### LINKED ISSUE #3192: Request for Enhancement: In package view, in list of builds, show directories
Hello. When I see this https://copr.fedorainfracloud.org/coprs/g/python/python3.13/package/python3.13 there are many builds. It is not visible that some of them are from custom directories.

Could the builds from directories be marked as such (or categorized, like when looking at https://copr.fedorainfracloud.org/coprs/g/python/python3.13/builds/)?

FrostyX:
Fix #3192

Showing the column only when a project has some CoprDirs to avoid confusing users that don't care about them. Also calling them "subprojects" to match the documentation.
https://docs.pagure.org/copr.copr/user_documentation.html#subprojects


FrostyX:
Some benchmarking is required. The builds table is one of the known performance bottlenecks and a missing JOIN would cause a havoc.

praiskup:
Screenshot would be nice here, too.

FrostyX:
Attaching screenshots here. Also, I didn't notice any performance issues, so I think we can merge if you think the code is ok.

Screenshot_2025-08-19_17-18-08

---

Screenshot_2025-08-19_17-18-25

praiskup:
Looks great!

PR #3778, opened 2025-07-08
Update documentation for builder images to use image-buidler and bootc

FrostyX:
Absorbing the temporary documentation from here
https://github.com/fedora-copr/copr-image-builder/blob/main/README-bootc.md


xsuchy (approved):
+1

PR #3779, opened 2025-07-09
Switch RPM upload to the synchronous API from pulp_rpm

dkliban:
Prior to this commit, a synchronous upload API was only available in pulp_service plugin. This API has now been released as part of pulp_rpm. This change will ensure COPR continues to be able to upload RPM packages when this API is removed from pulp_service plugin.


FrostyX (approved):

xsuchy (approved):
+1

PR #3780, opened 2025-07-09
python-copr: Stop using deprecated %py3_build/%py3_install macros

xsuchy:
First remove el7 cruft.
Then replace %py3_build/%py3_install macros.

Fixes: RHBZ#2378523


xsuchy:
One of the failure is for epel-7-x86_64 - I would argue that we should remove this from CI.

Issue #3743, opened 2025-06-04
Storage consumption charts are not working


praiskup:
Since 10th month 2024? Uh.

https://copr-be.cloud.fedoraproject.org/stats/distro.html

PR #3781, opened 2025-07-10
backend: drop the 'Z' suffix from timestamp

praiskup:
ParserError: Unknown string format: 2024-10-08T03:07:02.693262+00:00Z

Fixes: #3743


xsuchy (approved):
+1

PR #3782, opened 2025-07-10
backend: Stop using deprecated %py3_build/%py3_install macros

xsuchy:
remove el7 cruft 
and stop using deprecated %py3_build/%py3_install macros

Fixes: RHBZ#2377231

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -1,7 +1,3 @@
-%if 0%{?rhel} <= 7 && 0%{?rhel} > 0
-%global _pkgdocdir %{_docdir}/%{name}-%{version}
-%endif
-
 %global prunerepo_version 1.20
 %global tests_version 5
 %global tests_tar test-data-copr-backend
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -31,7 +31,6 @@ BuildRequires: systemd
 BuildRequires: util-linux
 
 BuildRequires: python3-devel
-BuildRequires: python3-setuptools
 
 BuildRequires: python3-copr
 BuildRequires: python3-copr-common >= %copr_common_version
@@ -105,6 +104,9 @@ Requires(postun): systemd
 
 %{?fedora:Requires(pre): lighttpd-filesystem}
 
+%generate_buildrequires
+%pyproject_buildrequires
+
 %description
 COPR is lightweight build system. It allows you to create new project in WebUI,
 and submit new builds and COPR will create yum repository from latest builds.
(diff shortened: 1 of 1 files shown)


backend/copr-backend.spec
```diff
@@ -1,7 +1,3 @@
-%if 0%{?rhel} <= 7 && 0%{?rhel} > 0
```
praiskup, line 1: I don't think we build copr-backend for RHEL.  Seems safe to abandon %py3_* scripts entirely here.

xsuchy: Ok. updated.

praiskup (approved):

backend/copr-backend.spec
```diff
@@ -109,6 +104,9 @@ Requires(postun): systemd
 
 %{?fedora:Requires(pre): lighttpd-filesystem}
```
hroncok, line 105: Just an observation: Should this be unconditional, considering "you don't think you build copr-backend for RHEL"?

praiskup: Good question, thanks @hroncok.  Too late (merged) but ... we know people build this for openEuler at least, which reminds me that we should discuss these changes with @pkking.

pkking: Hi @praiskup, i think this `requires` is needed in openEuler, and also, openEuler is still using %py3_* macros :(

praiskup: Ahah :-( would you mind taking care of this @pkking and add those conditions for components that do not build?  We don't have the needed capacity.

Also, these macros have serious issues, that's why Fedora drops the support.  What is the ETA for getting %pyproject_* into openEuler?

pkking: > Ahah :-( would you mind taking care of this @pkking and add those conditions for components that do not build? We don't have the needed capacity.
> 
> Also, these macros have serious issues, that's why Fedora drops the support. What is the ETA for getting %pyproject_* into openEuler?

Okay, i'd add some conditions when we upgrade copr.

Besides, old openEuler release keep using py3_* macros, they will EOL after 2027, i think RHEL will still need this macros too

And would you please provides the `py3_*` macros serious issues for reference?

hroncok: issues for reference:

> The upstream maintainers of the setuptools project have decided to remove the `setup.py install` command [in October 2025](https://github.com/pypa/setuptools/commit/4566569dc9f48adf970962277286615ab50df881). Both the `install` and `build` commands have been deprecated for 5 years.

From https://fedoraproject.org/wiki/Changes/DeprecateSetuppyMacros#Detailed_Description

Issue #3777, opened 2025-07-07
Leftovers in redis-db in Fedora Copr production


praiskup:
```
127.0.0.1:6379> keys *
...
430) "createrepo_batched::/var/lib/copr/public_html/results/@python/python3.14/fedora-rawhide-x86_64::2645885"
431) "createrepo_batched::/var/lib/copr/public_html/results/@copr/PyPI/epel-9-x86_64::3666469"
432) "createrepo_batched::/var/lib/copr/public_html/results/@python/python3.14-b1/fedora-rawhide-x86_64::1402552"
433) "createrepo_batched::/var/lib/copr/public_html/results/jplesnik/perl-test/fedora-rawhide-ppc64le::2448666"
434) "rpm_build_worker:9245644-fedora-42-s390x"
435) "createrepo_batched::/var/lib/copr/public_html/results/@copr/PyPI/fedora-42-x86_64::2905238"
436) "createrepo_batched::/var/lib/copr/public_html/results/@python/python3.14/fedora-rawhide-x86_64::2401902"
437) "createrepo_batched::/var/lib/copr/public_html/results/@python/python3.14/fedora-rawhide-x86_64::1865806"
438) "createrepo_batched::/var/lib/copr/public_html/results/torsava/automatic-obsoletes/fedora-rawhide-x86_64::2044008"
439) "createrepo_batched::/var/lib/copr/public_html/results/torsava/automatic-obsoletes/fedora-42-x86_64::1376177"
440) "createrepo_batched::/var/lib/copr/public_html/results/torsava/automatic-obsoletes/fedora-42-x86_64::1806927"
441) "createrepo_batched::/var/lib/copr/public_html/results/torsava/automatic-obsoletes/fedora-rawhide-x86_64::1725536"
442) "createrepo_batched::/var/lib/copr/public_html/results/@python/python3.14/fedora-rawhide-x86_64::13466"
443) "createrepo_batched::/var/lib/copr/public_html/results/jplesnik/perl-test/fedora-rawhide-s390x::754509"
444) "createrepo_batched::/var/lib/copr/public_html/results/@copr/PyPI/fedora-41-x86_64::2800263"
445) "createrepo_batched::/var/lib/copr/public_html/results/huakim/kde-plasma/fedora-42-s390x::663100"
446) "createrepo_batched::/var/lib/copr/public_html/results/jplesnik/perl-test/fedora-rawhide-ppc64le::3157009"
[...]

Issue #3770, opened 2025-06-25
Error from json.loads(task_dict["task"]) when deleting builds


FrostyX:
@jplesnik is repeatedly experiencing issues when deleting builds. They don't disappear from her results directory. Not a CDN issue.

The problem is here:

https://github.com/fedora-copr/copr/blob/d15a42fff96d5e0f791879ae067cafc9af2e97e7/backend/copr_backend/storage.py#L159-L167

See the logs:

```
[2025-06-25 07:25:00,693][  INFO][PID:2745612][/usr/bin/copr-backend-process-action.managed.pid-2745612][helpers.py:_info:107] Finished after 4 seconds with exit code 1 (copr-repo --batched /var/lib/copr/public_html/results/jplesnik/perl-test/fedora-rawhide-ppc64le --delete 09199802-perl-Fsdb --no-appstream-metadata)
stdout:

stderr:
Unexpected exception
Traceback (most recent call last):
  File "/usr/bin/copr-repo", line 445, in main
    main_try_lock(opts, batch)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/bin/copr-repo", line 397, in main_try_lock
    main_locked(opts, batch, opts.log)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/bin/copr-repo", line 323, in main_locked
    (batch_full, batch_add, batch_delete, batch_rpms_to_remove) = batch.options()
                                                                  ~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/copr_backend/createrepo.py", line 156, in options
    task_opts = json.loads(task_dict["task"])
                           ~~~~~~~~~^^^^^^^^
KeyError: 'task'
Traceback (most recent call last):
  File "/usr/bin/copr-repo", line 455, in <module>
    sys.exit(main())
             ~~~~^^
  File "/usr/bin/copr-repo", line 445, in main
    main_try_lock(opts, batch)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/bin/copr-repo", line 397, in main_try_lock
    main_locked(opts, batch, opts.log)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/bin/copr-repo", line 323, in main_locked
    (batch_full, batch_add, batch_delete, batch_rpms_to_remove) = batch.options()
[...]

PR #3783, opened 2025-07-11
backend: fix race condition in BatchedCreaterepo

praiskup:
Previously we observed two race conditions.

For #3770, this used to happen:

1. process A adds the `task A` entry to DB
2. process B takes the `task A`, processes it, and commits() it
3. process C lists the task A (due to absence of fair locking!)
4. process A (without lock) detects the `task A` is committed by `process B`, and removes the task from DB to optimize waiting (see #1423 why we do this)
5. process C reads the `task A` content by hgetall(), but receives an empty dict.

Previously we continued, and failed the createrepo run:

6. process C tries to process task A, but fails with KeyError (if this happens, `task B` was already dropped from DB so nobody re-tried).

The problem of #3770 was that `task B` sometimes requested a build removal, but did not finish.  In such case, Frontend just blindly removes the build from it's own database (trusting backend to finish the removal), and never gives the user chance to remove the build again. Solution for #3770 is to ignore the empty `task A` in 6.

Also even weirder situation happened in #3777:

2. process B takes the `task A`, processes it, and commits() for the first time (and still exists)
3. process A (without lock) detects the task is committed, and removes itself from DB to speedup (see why we do this in #1423)
4. process B commits() the `task A` once more (double-commit()), but this key was never removed from DB.
5. since the `status=succeeded`, any other process later on ignores this entry as finished.

This situation isn't known to cause any damages though.  To fix this issue we simply never double-commit() now.  And we have to clean the (innocent) DB entries manually.

Folow-up-for: 7442ec4303f2e0ba2e31a6555c4da34b96de0ee1
Fixes: #3770
Fixes: #3777
Relates: #1423

### DIFF
--- a/backend/copr_backend/createrepo.py
+++ b/backend/copr_backend/createrepo.py
@@ -110,6 +110,10 @@ class BatchedCreaterepo:
         finally:
             # This is atomic operation, other processes may not re-start doing this
             # task again.  https://github.com/redis/redis/issues/9531
+            #
+            # The 'delete_if_not = True' is not a TOCTOU case.  Our caller holds
+            # the lock, hence no other machine may take our task (would require
+            # lock) and re-process it.
             if status or delete_if_not:
                 self.redis.delete(self.key)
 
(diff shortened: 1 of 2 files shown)


backend/copr_backend/createrepo.py
```diff
@@ -147,17 +151,25 @@ def options(self):
             assert key != self.key
 
             task_dict = self.redis.hgetall(key)
+            if not task_dict:
```
praiskup, line 154: Race 1 closed here.  We simply ignore this nonexisting task, and it never comes back again.

backend/run/copr-repo
```diff
@@ -358,9 +358,6 @@ def main_locked(opts, batch, log):
     # possible we should move those two things before 'delete_builds' call.
     add_appdata(opts)
 
-    # while we still hold the lock, notify others we processed their task
-    batch.commit()
```
praiskup, line 362: Race 2 closed here; we don't do double-commit -- the second one possibly bringing the task ID back into the loop for no reason.

PR #3788, opened 2025-07-14
frontend: drop support for OID auth

### LINKED ISSUE #3769: Remove Openid login and keep just OIDC
Because of FTI
https://bugzilla.redhat.com/show_bug.cgi?id=2373920

And doing this may obsolete 
https://github.com/fedora-copr/copr/issues/3659

FrostyX:
Fix #3769

Don't worry, it removes OID not OIDC.


frontend/coprs_frontend/coprs/auth.py
```diff
@@ -22,38 +19,48 @@ class UserAuth:
     encapsulated within this class.
     """
 
-    @staticmethod
-    def logout():
+    @classmethod
+    def next_url(cls):
+        """
+        Where should user be redirected after a successful login?
```
praiskup, line 25: Ideally to the same page where he comes from (when clicking on the login button).

frontend/coprs_frontend/coprs/auth.py
```diff
@@ -22,38 +19,48 @@ class UserAuth:
...
+
+        At this moment, we redirect to the user profile but it would also make
+        sense to redirect back at the page where user clicked the login button.
+
+        We used to get the next URL through `oid.get_next_url()` but there is
+        no such equivalent for our current OIDC client. We need to set and get
+        it manually.
```
praiskup: If there really isn't a support for referrer, we probably need to handle this via a "hop" pre-login page ourselves?

praiskup: But it is weird there's no built-in support to return back :-(

praiskup: Bodhi uses something like this URL for log-in:
`https://bodhi.fedoraproject.org/login?came_from=https://bodhi.fedoraproject.org/updates/?status=testing&critpath=True`

praiskup (approved):

PR #3791, opened 2025-07-17
modernize spec file and drop old stuff

xsuchy:

### DIFF
--- a/cli/copr-cli.spec
+++ b/cli/copr-cli.spec
@@ -1,11 +1,4 @@
-%if 0%{?rhel} && 0%{?rhel} <= 7
-%global __python %_bindir/python2
-%global with_python2 1
-%else
 %global __python %_bindir/python3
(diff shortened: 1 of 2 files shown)


cli/copr-cli.spec
```diff
@@ -105,11 +72,10 @@ install -m 755 copr_cli/package_build_order.py %{buildroot}/%{_bindir}/package-b
 
 
 %check
-%{__python} -m pytest -vv tests
+%{__python3} -m pytest -vv tests
```
mfocko, line 75: Can you do that? I see `__python` declared at the top 👀

xsuchy: I left the macro on top because without it, rpmbuild yells "uversioned python usage detected". And yes, __python3 macro exists.

praiskup (approved):

PR #3801, opened 2025-07-23
backend: make sure every PULP api call is logged

praiskup:


FrostyX (approved):
This is a very good idea.

PR #3807, opened 2025-07-25
Adds batching to add_content() and delete_content() in Pulp client.

dkliban:
The locking logic is borrowed from the implementation in the BackendStorage.

Each call to PulpClient.add_content() and PulpClient.delete_content() now works like this:

1. Writes the request to Redis.
2. Check if another process processed the request. Return if so.
2. Try to acquire a lock.
3. If lock is acquired a. Collect all requests to add and remove content and make one request to Pulp. b. Create a publication If lock is not acquire, go back to step 2.
4. Notify other processes that their request was processed.

The PulpStorage.publish_repository() is now a no-op.

The distribution is no longer updated after a publication is created. The distribution always points to the latest publication for a particular repository.

This reduces the number of new repository versions and publications that are created for large repositories.

### DIFF
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -10,6 +10,145 @@ import tomllib
 from urllib.parse import urlencode
 import requests
 
+from copr_common.lock import lock, LockTimeout
+from copr_common.redis_helpers import get_redis_connection
+
+
+class BatchedAddRemoveContent:
+    """
+    Group a set of `add_and_remove` tasks for a single Pulp repository.
+    """
+    def __init__(self, repo_href, rpms_to_add, rpms_to_remove, backend_opts=None, log=None, noop=False):
+        self.repo_href = repo_href
+        self.rpms_to_add = rpms_to_add or []
+        self.rpms_to_remove = rpms_to_remove or []
+        self.log = log
+        self.noop = noop
+
+        if not backend_opts:
+            self.log.error("can't get access to redis, batch disabled")
+            self.noop = True
+            return
+
+        self._pid = os.getpid()
+        self._json_redis_task = json.dumps({
+            "add_content_units": self.rpms_to_add,
+            "remove_content_units": self.rpms_to_remove,
+        })
+
+        self.notify_keys = []
+        self.redis = get_redis_connection(backend_opts)
+
+    @property
+    def key(self):
+        """ Our instance ID (key in Redis DB) """
+        return "add_remove_batched::{}::{}".format(
+            self.repo_href, self._pid)
+
+    @property
+    def key_pattern(self):
+        """ Redis key pattern for potential tasks we can batch-process """
+        return "add_remove_batched::{}::*".format(self.repo_href)
+
+    def make_request(self):
+        """ Request the task into Redis DB.  Run _before_ lock! """
+        if self.noop:
+            return None
+        self.redis.hset(self.key, "task", self._json_redis_task)
+        return self.key
+
+    def check_processed(self, delete_if_not=True):
+        """
+        Drop our entry from Redis DB (if any), and return True if the task is
+        already processed.  When 'delete_if_not=True, we delete the self.key
+        from Redis even if the task is not yet processed (meaning that caller
+        plans to finish the task right away).
+        """
+        if self.noop:
+            return False
+
+        status = self.redis.hget(self.key, "status") == "success"
+        self.log.debug("Has already a status? %s", status)
+
+        try:
+            if not status:
+                # not yet processed
+                return False
+        finally:
+            # This is atomic operation, other processes may not re-start doing this
+            # task again.  https://github.com/redis/redis/issues/9531
+            #
+            # The 'delete_if_not = True' is not a TOCTOU case.  Our caller holds
+            # the lock, hence no other machine may take our task (would require
+            # lock) and re-process it.
+            if status or delete_if_not:
+                self.redis.delete(self.key)
+
+        return status
+
+    def options(self):
+        """
+        Get the options from other _compatible_ (see below) Redis tasks, and
+        plan the list of tasks in self.notify_keys[] that we will notify in
+        commit().
+        """
+        rpms_to_add = set(self.rpms_to_add)
+        rpms_to_remove = set(self.rpms_to_remove)
+
+        if self.noop:
+            return {"add_content_units": rpms_to_add, "remove_content_units": rpms_to_remove}
+
+        for key in self.redis.keys(self.key_pattern):
+            assert key != self.key
+
+            task_dict = self.redis.hgetall(key)
+            if not task_dict:
+                # TOCTOU: The key might no longer exist in DB, even though
+                # _our process_ holds the lock!  See the
+                # 'self.redis.delete(self.key)' above in check_processed(); the
+                # _original process_ removes key without locking, if _third_
+                # process commit()ted the status=succeeded.
+                # Prior fixing #3770/#3777 we tried to process these deleted
+                # tasks and ended up with KeyError on task_dict["task"] below.
+                self.log.info("Key %s already processed (by other process), "
+                              "removed by the original process (without lock) "
+                              "since we listed it.", key)
+                continue
+
+            if task_dict.get("status") is not None:
+                # skip processed tasks
+                self.log.info("Key %s already processed, skip", key)
+                continue
+
+            task_opts = json.loads(task_dict["task"])
+
+            # we can process this task!
+            self.notify_keys.append(key)
+
+            # append "add" tasks, if that makes sense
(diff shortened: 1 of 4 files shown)


backend/copr_backend/pulp.py
```diff
@@ -10,6 +10,146 @@
...
+        Drop our entry from Redis DB (if any), and return True if the task is
+        already processed.  When 'delete_if_not=True, we delete the self.key
+        from Redis even if the task is not yet processed (meaning that caller
+        plans to finish the task right away).
+        """
+        if self.noop:
+            return False
+
+        status = self.redis.hget(self.key, "status") == "success"
+        self.log.debug("Has already a status? %s", status)
```
praiskup, line 71: I am getting PyLint errors:
>  self.log.error("can't get access to redis, batch disabled") # E: Instance of 'BatchedAddRemoveContent' has no 'log' member 

Not sure why pylint here @ gitlab is green, probably related to #3809.

backend/copr_backend/pulp.py
```diff
@@ -195,27 +336,115 @@ def create_content(self, path, labels):
...
+
+            if batch.check_processed(delete_if_not=False):
+                self.log.info("Task processed by other process (no-lock)")
+                return
+
+            try:
+                lockdir = os.environ.get(
+                    "COPR_TESTSUITE_LOCKPATH", "/var/lock/copr-backend")
+                with lock(self.opts.directory, lockdir=lockdir, timeout=5, log=self.log):
+                    response = self.modify_repository_content_locked(batch)
```
praiskup: This takes two arguments

backend/copr_backend/pulp.py
```diff
@@ -195,27 +336,115 @@ def create_content(self, path, labels):
...
+                    # If no exception happened, we are done (break).
+                    self.log.debug("Repository version and Publication created by this process")
+                    break
+            except LockTimeout:
+                continue  # Try again...
+
+            # we never loop, only upon timeout
+            assert False
+
+    def modify_repository_content(self, repository, add_content_units=[], remove_content_units=[]):
```
praiskup: Valid PyLint's warning: # W: Dangerous default value [] as argument

praiskup:
@dkliban once you are back working on this PR, please just rebase on top of current `main`

backend/copr_backend/pulp.py
dkliban: Yep, once i ran this code i discovered it also.

backend/copr_backend/pulp.py
dkliban: i fixed it

backend/copr_backend/pulp.py
dkliban: I change this to just be None.

backend/copr_backend/pulp.py
```diff
@@ -195,27 +321,119 @@ def create_content(self, path, labels):
...
+        batch = BatchedAddRemoveContent(repository, artifacts, None, self.opts, self.log)
+        # Put our task to Redis DB and allow _others_ to process our
+        # own task.  This needs to be run _before_ the lock() call.
+        batch.make_request()
+        try:
+            self.try_lock(repository, batch)
+            self.log.info("Pulp: add_content: %s (%s)", repository, artifacts)
+            return True
+        except Exception as e:
+            self.log.info("Pulp: Failed to add content: %s %s %s", repository, artifacts, e)
```
praiskup: The log goes to `backend.log` (visible by user), could we by any chance leak the PULP key here through str(e)?

backend/copr_backend/pulp.py
```diff
@@ -195,27 +338,120 @@
...
+            import traceback;self.log.info(traceback.format_exc())
```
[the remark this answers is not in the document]
praiskup: Seems like a leftover.

backend/copr_backend/pulp.py
```diff
@@ -195,27 +338,120 @@
...
-        return requests.post(url, json=data, **self.request_params)
+        batch = BatchedAddRemoveContent(repository, None, artifacts, self.opts, self.log)
+        # Put our task to Redis DB and allow _others_ to process our
+        # own task.  This needs to be run _before_ the lock() call.
+        batch.make_request()
+        try:
+            self.try_lock(repository, batch)
+            self.log.info("Pulp: delete_content: %s (%s)", repository, artifacts)
+            return True
+        except Exception:
```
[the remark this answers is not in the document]
praiskup: We typically respect this PyLint warning; I think it is fair to let this exception bomb out?  Or is there any exception that we expect to come "normally"?

backend/copr_backend/storage.py
```diff
@@ -276,49 +276,10 @@ def create_repository_version(self, chroot, package_hrefs):
...
-
-        # Do we want to update the distribution to point to a specific
-        # publication? When not doing so, the distribution should probably
-        # automatically point to the latest publication
-        response = self.client.update_distribution(distribution, publication)
-        if not response.ok:
-            self.log.error("Failed to update Pulp distribution %s for because %s",
-                           distribution_name, response.text)
-            return False
+        # Publishing occurs after each repository version is created.
```
praiskup, line 282: How do we achieve this?

backend/copr_backend/pulp.py
```diff
@@ -10,6 +10,148 @@
...
+                continue
+
+            task_opts = json.loads(task_dict["task"])
+
+            # we can process this task!
+            self.notify_keys.append(key)
+
+            # append "add" tasks, if that makes sense
+            rpms_to_add.update(task_opts["add_content_units"])
+            rpms_to_remove.update(task_opts["remove_content_units"])
```
praiskup, line 130: Note three's a chance that the same prn may appear in both sets, so perhaps we shouldn't request the same thing to be added and removed in one API call?

Also, you can use set operators like `rpms_to_add |= set(task_opts["add_content_units"])`, and

```
common = rpms_to_add & rpms_to_del
rpms_to_add -= common
rpms_to_del -= common
```
```

File: backend/copr_backend/pulp.py [hunk 9]
```diff
@@ -10,6 +10,148 @@
...
+
+            if len(self.notify_keys) >= MAX_IN_BATCH - 1:  # one is ours!
+                self.log.info("Batch add-remove limit %s reached, skip the rest",
+                              MAX_IN_BATCH)
+                break
```
praiskup: I think we can drop this "MAX" thing.  The reason we do that for backend storage is that we use subprocess & execve() that has some commandline limits, but I doubt we have limits for json API payload size?

The thing is, especially for removals, that you get a very large json here... consider user selects thousands of builds for deletion (single button), each build has many chroot-builds, and each chroot build many prns.

backend/copr_backend/pulp.py
dkliban: I'll remove the e. It's a leftover from my debugging.

backend/copr_backend/pulp.py
dkliban: i'll add some logic to make sure we are not doing unnecessary work.

backend/copr_backend/pulp.py
dkliban: i don't remember what the maximum size for a POST request payload is, but we should probably not worry about that right now. if we run into it at some point we can add a max limit here.

backend/copr_backend/pulp.py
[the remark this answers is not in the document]
dkliban: You are right ... we should just let the exception happen. I don't know of any exceptions that we are expecting.

backend/copr_backend/storage.py
dkliban: It's now done as part of the `modify_repository_content_locked` method.

backend/copr_backend/pulp.py
```diff
@@ -10,6 +10,148 @@
...
+
+
+MAX_IN_BATCH = 100
+
+
+class BatchedAddRemoveContent:
+    """
+    Group a set of `add_and_remove` tasks for a single Pulp repository.
+    """
+    def __init__(self, repo_href, rpms_to_add, rpms_to_remove, backend_opts=None, log=None, noop=False):
```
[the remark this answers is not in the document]
FrostyX: @dkliban you said you are still fighting the linter. This is the only one issue remaining, and I would just ignore this one. You can put this as the first line of the function and linter will be happy.

```
# pylint: disable=too-many-positional-arguments
```

FrostyX:
Thank you very much for the PR @dkliban. I skimmed it over and didn't see any obvious errors. I discovered a small issue when testing it locally and submitted a fix in PR #3816. Other than that, it works. I think we can merge this and do any follow-ups in separate PRs if needed.

Issue #3643, opened 2025-02-24
Setup development/testing environment for OIDC


nikromen:
Setup this for our local development so we can log-in inside containerized copr. This could be achieved by https://github.com/fedora-infra/tiny-stage

Follow up from https://github.com/fedora-copr/copr/issues/3483

pkking:
We are trying to pack copr as a helm chart to simplify the local development, tiny stage seems a interesting way :)

PR #3808, opened 2025-07-25
Fix docker compose after sysusers.conf and auth changes

FrostyX:
With @dkliban, we discovered that our docker compose is completely broken, failing the builds at every step of the process. 

See the commit description from `explicitly set home directory and shell for the user`, that was the most surprising issue.

### DIFF
--- a/docker/frontend/files/etc/copr/copr.conf
+++ b/docker/frontend/files/etc/copr/copr.conf
@@ -47,7 +47,7 @@ SEND_EMAILS = False
 ## possible options: None, "http", "https"
 # ENFORCE_PROTOCOL_FOR_BACKEND_URL = "https"
 ## post-process url leading to frontend
-# ENFORCE_PROTOCOL_FOR_FRONTEND_URL = "https"
+ENFORCE_PROTOCOL_FOR_FRONTEND_URL = "http"
 
 # URLs for fedmenu resources
 #FEDMENU_URL = "http://threebean.org/fedmenu/"


praiskup:
The feature is RPM based, no longer systemd. And.. note this https://github.com/rpm-software-management/mock/pull/1604 ... i realized sysusers work in F42+ :(

dist-git/conf/copr-dist-git.sysusers.conf
```diff
@@ -1,7 +1,7 @@
 #Type Name     ID             GECOS                 Home directory Shell
 g     packager -
 g     copr-dist-git -
-u     copr-dist-git -         'copr-dist-git user'  -              -
+u     copr-dist-git -         'copr-dist-git user'  /home/copr-dist-git  /bin/bash
```
praiskup: Is it correct to create these tmpfiles on non-tmpfs filesystem?  That's the first time I see this being done :shrug: people normally executed `mkhomedir_helper`

docker/frontend/files/usr/bin/init-database.sh
```diff
@@ -3,3 +3,8 @@
 cd /usr/share/copr/coprs_frontend/ && copr-frontend create-db --alembic alembic.ini
 copr-frontend create-chroot \
     $(ls /etc/mock/{fedora,epel}-*-{i386,x86_64}.cfg |xargs -I{} -n1 basename {} .cfg)
+
+# OIDC and Kerberos don't support auth on local developer instances without
+# workarounds. Let's have a default user for easier onboarding.
+copr-frontend add-user jdoe jdoe@example.com
+copr-frontend alter-user jdoe --admin
```
nikromen: we could get rid of it in the future by https://github.com/fedora-copr/copr/issues/3643

could you please link the issue there? I think I will implement it in the future if I'll see the todo too much often :D

docker/frontend/files/etc/copr/copr.conf
```diff
@@ -47,7 +47,7 @@ SEND_EMAILS = False
 ## possible options: None, "http", "https"
 # ENFORCE_PROTOCOL_FOR_BACKEND_URL = "https"
 ## post-process url leading to frontend
-# ENFORCE_PROTOCOL_FOR_FRONTEND_URL = "https"
+ENFORCE_PROTOCOL_FOR_FRONTEND_URL = "http"
```
nikromen, line 50: why http here? :thinking:

FrostyX: Because this affects how the API config is generated:

```
[copr-cli]
login = ...
username = frostyx
token = ...
copr_url = http://localhost:5000
encrypted = False
# expiration date: 2025-08-05
```

The `http` URL there and also the `encrypted = False`. Otherwise, `copr-cli` doesn't work against the local instance. Or at least it never worked for me without this.

docker/frontend/files/usr/bin/init-database.sh
FrostyX: Good point, thank you. Updated.

FrostyX:
Thank you for the test @dkliban.

We should probably let @pkking, @TommyLike, and @Bob131 know about this PR, just to make sure they are not surprised by the changes. Maybe the default user could be problematic for them if they use the containers in production.

dist-git/conf/copr-dist-git.sysusers.conf
FrostyX: In my understanding, this has nothing to do with tmpfs.

I think there are two different concepts. The "dynamic user allocatoin"
https://docs.fedoraproject.org/en-US/packaging-guidelines/UsersAndGroups/#_dynamic_allocation
which creates `/etc/passwd` entries from a `.conf` file like this ^^

And tmpfiles.d which can create files and directories from a `.conf` file, including user home directories
https://docs.fedoraproject.org/en-US/packaging-guidelines/Tmpfiles.d/

I don't understand at all, why dynamic user allocation doesn't automatically create the home directories, and we need to add another entry. And I couldn't make it work, so I workarounded it by creating the directories manually in dockerfiles.

However, the home directory and shell values that I set here in this line, they work. With the previous `-` values, the home directory was set to `/` and shell to `/bin/nologin`.

pkking:
> Thank you for the test @dkliban.
> 
> We should probably let @pkking, @TommyLike, and @Bob131 know about this PR, just to make sure they are not surprised by the changes. Maybe the default user could be problematic for them if they use the containers in production.

IIUC, this PR does something similar with https://github.com/fedora-copr/copr/pull/3663#discussion_r2009603924 but in `docker-compose`, when merging, the docker image will create user rely on the rpm itself, and if openEuler keep using `rpm < 4.19`, thats means we should still create user and set shell manully?

On the other hand, the only impact for me when we upgrade to `rpm >= 4.19` is the uid is not fixed, but in our prod cluster, we pin the copr family user in some uid before 1000, any suggestion?

FrostyX:
> when merging, the docker image will create user rely on the rpm itself

@pkking there are multiple changes related to users, so I should have been more precise, sorry :-)

There is a change in `dist-git/conf/copr-dist-git.sysusers.conf` which only specifies the home directory and login shell for the `copr-dist-git` user. I don't think that should negatively affect you.

What I wanted to be careful about was the change in `init-database.sh` which creates a default user and makes them an admin. Very useful for development, but if you use this script in production it would be a problem. Additionally, the change in `docker/frontend/files/entrypoint` allows to log in as that user without using a password.

dist-git/conf/copr-dist-git.sysusers.conf
praiskup: Argh, sorry.  I thought I'm commenting on the tmpfiles.d snippet.
Yes, we probably need to specify homedir explicitly, just not sure if we want /home sub-dir.

dist-git/conf/tmpfiles.d/copr-dist-git.conf
```diff
@@ -1 +1,2 @@
 D /var/lock/copr-dist-git 0700 copr-dist-git copr-dist-git -
+d /home/copr-dist-git 0755 copr-dist-git copr-dist-git -
```
praiskup: This is what I wanted to comment on ...
Btw, I tried `cat /usr/lib/tmpfiles.d/*.conf | grep /home` (no hit)
Isn't /home suspicious?

dist-git/conf/copr-dist-git.sysusers.conf
FrostyX: I will test `/var/lib/copr-dist-git` that you suggested on Slack.

FrostyX: Worked fine, I updated the PR.

pkking:
> What I wanted to be careful about was the change in `init-database.sh` which creates a default user and makes them an admin. Very useful for development, but if you use this script in production it would be a problem. Additionally, the change in `docker/frontend/files/entrypoint` allows to log in as that user without using a password.

I think this hack will be temperary due to https://github.com/fedora-copr/copr/issues/3643 ？

dist-git/conf/tmpfiles.d/copr-dist-git.conf
```diff
@@ -1 +1,2 @@
 D /var/lock/copr-dist-git 0700 copr-dist-git copr-dist-git -
+d /var/lib/copr-dist-git 0755 copr-dist-git copr-dist-git -
```
praiskup:
```
[root@copr-dist-git ~][PROD]# rpm -qf /var/lib/copr-dist-git/
copr-dist-git-1.1-1.fc41.noarch
```

Commit message:
```
Also, the user home directory is not created automatically.
```

It is, because it is a [part of the copr-dist-git.rpm's payload](https://github.com/fedora-copr/copr/blob/62d0f7f8c8710b1cdd7bd0ba2d3761c8ab7f6ae8/dist-git/copr-dist-git.spec#L120)?

docker/backend/Dockerfile
```diff
@@ -79,8 +79,10 @@ RUN chown copr:root /etc/sign.conf && \
     chmod 0660 /etc/sign.conf
 
 # Not sure why the directory wasn't created automatically
-RUN mkdir /var/lock/copr-backend
-RUN chown copr:copr /var/lock/copr-backend
+RUN mkdir -p /var/lock/copr-backend /var/lib/copr/public_html/results
```
praiskup: > Not sure why the directory wasn't created automatically

The /var/lock is a tmpfs

```
$ df -h /var/lock/
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            11G  2.9M   11G   1% /run
```

 normally, when VM is started, tmpfiles.d create those:
```
$ cat /usr/lib/tmpfiles.d/legacy.conf | grep /var/lock
L /var/lock - - - - ../run/lock
```

Podman runtime doesn't seem to pre-create /var/lock, my bet is that it is considered legacy by FHS, hence the `legacy.conf` ?? (but I'm lazy to check now).

It might be a good idea to move from /var/lock to some /run sub-directory long-term :shrug: (if /var/lock is legacy).  But for now, we have to create /var/lock/copr-backend manually :+1: 

The ` /var/lib/copr/public_html/results` is a different story.  We probably bind-mount the parent directory **at runtime** into the container, overriding what has been installed with copr-backend.rpm payload at **build time**.  I doubt that `mkdir -p $resultdir` has any effect.

docker/distgit/Dockerfile
```diff
@@ -34,8 +34,12 @@ RUN sed -i 's/Listen 80/Listen 5001/' /etc/httpd/conf/httpd.conf
...
 RUN chown copr-dist-git:packager /tmp/copr-dist-git
 
-RUN mkdir /var/lock/copr-dist-git/
-RUN chown copr-dist-git:copr-dist-git /var/lock/copr-dist-git/
+# These should be created by the RPM package and tmpfiles.d but for some reason
+# it doesn't work. Workarounding it here.
+RUN mkdir /var/lock/copr-dist-git/ /home/copr-dist-git
+RUN chown copr-dist-git:copr-dist-git \
+    /var/lock/copr-dist-git/ \
+    /home/copr-dist-git
```
praiskup: We are moving from `/home` to `/var/lib` in different commit.

backend/conf/copr-backend.sysusers.conf
```diff
@@ -1,3 +1,3 @@
 #Type Name     ID             GECOS                 Home directory Shell
-u     copr     -              'COPR user'           -              /bin/bash
+u     copr     -              'COPR user'           /home/copr              /bin/bash
```
praiskup, line 2: I _think_ this deserves cleanup long-term, too.  `/home/copr` is sub-optimal for service accounts (we do the same right now for copr-dits-git).  FTR, also `copr-frontend` uses `/usr/share/copr/coprs_frontend`.

But I don't think we want to change this now in this PR - moving to `/usr/share/<something>` .  Opening https://github.com/fedora-copr/copr/issues/3812

docker/distgit/Dockerfile
FrostyX: Good catch, updated.

docker/backend/Dockerfile
FrostyX: I see your point about the lock directory but since `RUN mkdir /var/lock/copr-backend` was already there, I'd keep it as is.

> The  /var/lib/copr/public_html/results is a different story. We probably bind-mount the parent directory at runtime into the container

True

>  I doubt that mkdir -p $resultdir has any effect.

I doubted too but I think it indeed helped. I will try removing it, to double check.

dist-git/conf/tmpfiles.d/copr-dist-git.conf
FrostyX: > Also, the user home directory is not created automatically.

I came to that conclusion from
https://www.freedesktop.org/software/systemd/man/latest/sysusers.d.html

> systemd-sysusers only sets the home directory record in the user
> database. To actually create the directory, consider adding a
> corresponding tmpfiles.d(5) fragment.

but it may be a wrong conclusion. 

We have `/var/lib/copr-dist-git` now but previously I had `/home/copr-dist-git` ... that's the problem. The RPM package probably created `/var/lib/copr-dist-git` through 

```
%dir %attr(0755, copr-dist-git, copr-dist-git) %{_sharedstatedir}/copr-dist-git/
```

but I didn't care for it because I wanted `/home/copr-dist-git`. Sooo, I think we can remove this tmpfiles.d line.

docker/distgit/Dockerfile
FrostyX: Actually, it may be a bad idea.

I was creating the `/home/copr-dist-git` directory because the very next Dockerfile line is

```
RUN echo " [user]" >> /home/copr-dist-git/.gitconfig && \
```

I just checked, and in production we have this

```
[copr-dist-git@copr-dist-git ~][PROD]$ echo $HOME
/home/copr-dist-git
```

I think we should revert some of the changes we made recently in this PR and stick to `/home/copr-dist-git`

docker/backend/Dockerfile
FrostyX: Hmm, I don't know what happened. I tried removing it and saw no problems. So, updated.

praiskup (approved):

PR #3811, opened 2025-07-30
frontend: redirect to the current URL after OIDC login

FrostyX:
See #3788
See #3769

This uses cookie to remember the current URL before we get redirected to OIDC and then destroys it after a redirect. This is the only way (confirmed by @abompard) because an OIDC provider can only redirect to a whitelisted URL, which cannot contain a wildcard.
So it cannot be https://copr.fedorainfracloud.org/oidc_auth?next=* Cookie is also what Bodhi does.


frontend/coprs_frontend/coprs/auth.py
```diff
@@ -31,11 +31,16 @@ def next_url(cls):
...
             if cls._is_safe_next_url(url):
-                return flask.request.values.get("next")
+                return url
 
         if url := flask.request.referrer:
             if cls._is_safe_next_url(url):
-                return flask.request.referrer
+                return url
+
+        # When reading the URL from a cookie, pop it so it's usable only once
```
praiskup: s/cookie/session/? (see the response.set_cookie() and get_cookie())

frontend/coprs_frontend/coprs/auth.py
```diff
@@ -31,11 +31,16 @@ def next_url(cls):
...
+        if url := flask.session.pop("next", None):
+            if cls._is_safe_next_url(url):
+                return url
```
praiskup, line 43: Shouldn't we move this logic above the referrer logic?  (meaning that, even if referrer is set, we should redirect per "next"?)

frontend/coprs_frontend/coprs/auth.py
FrostyX: I can change it, whichever makes more sense. Both words are true:
https://flask.palletsprojects.com/en/stable/quickstart/#sessions

> there is also a second object called [session](https://flask.palletsprojects.com/en/stable/api/#flask.session) which allows you to store information specific to a user from one request to the next. This is implemented on top of cookies for you and signs the cookies cryptographically. What this means is that the user could look at the contents of your cookie but not modify it

frontend/coprs_frontend/coprs/auth.py
FrostyX: Good point, updated.

frontend/coprs_frontend/coprs/auth.py
praiskup: Please change the wording :pray: 
The whole session is implemented using just one "session cookie", and if you say "cookies" it is really confusing (we don't use cookies @ Copr)

praiskup: Ad the difference, take a look how we store session variables.

FrostyX: > The whole session is implemented using just one "session cookie"

Ahaaa, I did not understand that from the docs. Updated, PTAL.

praiskup (approved):

Issue #3497, opened 2024-11-01
Pulp - Forking repositories and fedora-branching


FrostyX:

FrostyX:
Marking this `ci-blocker` because builds for our PRs started failing with:

```
>>> Status code: 404 for https://console.redhat.com/api/pulp-content/public-copr
>>> Status code: 404 for https://console.redhat.com/api/pulp-content/public-copr
>>> Status code: 404 for https://console.redhat.com/api/pulp-content/public-copr
>>> Status code: 404 for https://console.redhat.com/api/pulp-content/public-copr
>>> Usable URL not found                                                        
Failed to download metadata (baseurl: "https://download.copr.fedorainfracloud.org/results/@copr/copr-dev/fedora-43-x86_64/") for repository "copr_copr_copr_dev": Usable URL not found

Copr build error: Build failed
```

This is the newly branched F43 chroot, which happened only after the project was migrated to Pulp.

PR #3813, opened 2025-08-03
Implement Pulp forking

### LINKED ISSUE #3500: sign && re-sign (no support for obs-sign in PULP)

FrostyX:
Fix #3500
See #3497


FrostyX:
There are still some leftover TODOs, tests are failing and linter is broken, but it's starting to work. So in case anyone wanted to see the progress on this feature.

FrostyX:
Oops, sorry, I pushed this into `fedora-copr` branch, not my fork branch.

FrostyX:
This PR depends on PR #3817 so marking it as blocked.

docker/backend/files/etc/copr/copr-be.conf
```diff
@@ -64,6 +64,12 @@ prune_days=14
 redis_host=redis
 redis_port=6379
 
+# URL where Pulp results are available
+# For easy setups it can be generated on the fly from `pulp status`. But is not
+# possible for more complicated deployments like hosted Fedora Pulp.
+pulp_content_url=http://pulp:80/pulp/content
+
+
```
FrostyX, line 72: We will need to add this setting also to the ansible playbooks for production and STG. We already have the value there because we set it in the frontend config. We just need to use the same variable for backend config as well.

backend/copr_backend/pulp.py
```diff
@@ -426,19 +426,24 @@ def delete_content(self, repository, artifacts):
...
-            if query:
+        query = "("
+        for i, build_id in enumerate(build_ids):
+            if i:
                 query += " OR "
             query += f"pulp_label_select=\"build_id={build_id}\""
+        query += ")"
+
+        if chroot:
+            query += f" AND pulp_label_select=\"chroot={chroot}\""
```
FrostyX, line 442: This turned out to be an unneeded change, but I kept it because it could be useful later.

PR #3817, opened 2025-08-06
Fix Pulp builds for CoprDirs

FrostyX:
I don't know when this stopped working, it was definitely broken before
PR #3807. The builds in CoprDirs was publishing the results into the main
repository instead of the CoprDir repository. This PR fixes that.

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -688,10 +688,21 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
             self.job.target_dir_name,
             build_id=self.job.build_id,
         )
-        if result:
-            # Only PulpStorage returns package HREFs
-            if not self.storage.create_repository_version(self.job.chroot, result):
-                raise BackendError("Pulp: Failed to create repository version.")
+
+        # Only PulpStorage returns package HREFs
+        if not result:
+            return
+
+        success = self.storage.create_repository_version(
+            self.job.project_dirname,
+            self.job.chroot,
+            result,
+        )
+        if not success:
+            raise BackendError(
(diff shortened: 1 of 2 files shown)


FrostyX:
This PR includes a commit from PR #3816, so I am marking this one blocked and we will rebase afterward.

backend/copr_backend/background_worker_build.py
```diff
@@ -688,10 +688,21 @@ def _upload_results_to_storage(self):
             self.job.target_dir_name,
             build_id=self.job.build_id,
         )
-        if result:
-            # Only PulpStorage returns package HREFs
-            if not self.storage.create_repository_version(self.job.chroot, result):
-                raise BackendError("Pulp: Failed to create repository version.")
+
+        # Only PulpStorage returns package HREFs
```
praiskup, line 692: It's just a bit weird that we can not encapsulate this into the storage classes' logic.

backend/copr_backend/background_worker_build.py
```diff
@@ -688,10 +688,21 @@ def _upload_results_to_storage(self):
...
+        if not result:
+            return
+
+        success = self.storage.create_repository_version(
+            self.job.project_dirname,
+            self.job.chroot,
+            result,
```
praiskup, line 699: Nit, would be nice to use name `hrefs` instead of `result`.

praiskup (approved):

backend/copr_backend/background_worker_build.py
FrostyX: +1 once all the Pulp features are finished, we should try to revisit the abstractions and interfaces.

PR #3841, opened 2025-08-27
update e-mail

praiskup:

### DIFF
--- a/frontend/coprs_frontend/coprs/auth.py
+++ b/frontend/coprs_frontend/coprs/auth.py
@@ -103,6 +103,8 @@ class UserAuth:
         """
         user = UsersLogic.get(username).first()
         if user:
+            if email is not None:
+                user.mail = email
             return user
         app.logger.info("Login for user '%s', "
                         "creating a database record", username)
--- a/frontend/coprs_frontend/coprs/logic/users_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/users_logic.py
@@ -193,7 +193,6 @@ class UserDataDumper(object):
             "api_login": self.user.api_login,
             "api_token": self.user.api_token,
             "api_token_expiration": self.user.api_token_expiration.strftime("%b %d %Y %H:%M:%S"),
-            "gravatar": self.user.gravatar_url,
         }
 
     @property
--- a/frontend/coprs_frontend/coprs/logic/users_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/users_logic.py
@@ -138,7 +138,8 @@ class UsersLogic(object):
                 "api_login": "",
                 "api_token": "",
                 "api_token_expiration": datetime.date(1970, 1, 1),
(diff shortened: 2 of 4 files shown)


frontend/coprs_frontend/coprs/auth.py
```diff
@@ -103,6 +103,8 @@ def get_or_create_user(username, email=None, timezone=None):
         """
         user = UsersLogic.get(username).first()
         if user:
+            if email is not None:
+                user.mail = email
```
FrostyX, line 107: This won't update the email value in the database, though?

frontend/coprs_frontend/coprs/logic/users_logic.py
```diff
@@ -138,7 +138,8 @@ def delete_user_data(cls, user):
                 "api_login": "",
                 "api_token": "",
                 "api_token_expiration": datetime.date(1970, 1, 1),
-                "openid_groups": None}
+                "openid_groups": None,
+                "mail": "deleted@example.com"}
```
FrostyX: We show an email address in https://copr.fedorainfracloud.org/coprs/frostyx/ . We should make sure "None" pointing to `mailto:None` doesn't appear there

(Maybe it doesn't, I didn't test yet)

frontend/coprs_frontend/coprs/auth.py
praiskup: Yes, the object is committed, I believe.

frontend/coprs_frontend/coprs/logic/users_logic.py
praiskup: does this apply to deleted accounts?

praiskup:
Per meeting:  We should log the user-off (drop session) and redirect to copr homepage.

frontend/coprs_frontend/coprs/logic/users_logic.py
FrostyX: Sorry, to cause all the confusion about this @praiskup. I was convinced that this sets `"mail"` to `None`. Only now I see that it is actually "deleted@example.com".

I am not particularly happy about it, but given the fact that nobody uses this feature, let's use this value instead of rewriting the model to allow `None` values.

frontend/coprs_frontend/coprs/auth.py
FrostyX: Hard to test because it doesn't update the value when using Kerberos (+hacks for localhost login) in my docker-compose environment. We'll need to see on STG.

frontend/coprs_frontend/coprs/logic/users_logic.py
praiskup: Actually, I realized that `deleted@exmaple.com` confuses kerberos logic; it thinks it has the needed data, and lets the user in. :-( I think we need to go with "empty string" instead.

PR #3853, opened 2025-09-01
cleanup: remove gently_gpg_sha256 config option

xsuchy:
This reverts db73448bda5e164e726f533a4c62f72675381c0d that is no longer needed


nikromen:
we need this for rhcopr :(

PR #3854, opened 2025-09-01
Stop using deprecated %py3_build/%py3_install macros 

xsuchy:

### DIFF
--- a/dist-git/copr-dist-git.spec
+++ b/dist-git/copr-dist-git.spec
@@ -61,12 +61,16 @@ This package contains Copr services for Dist Git server.
 %setup -q
 
 
+%generate_buildrequires
+%pyproject_buildrequires
+
+
 %build
-%py3_build
+%pyproject_wheel
 
 
 %install
-%py3_install
+%pyproject_install
 
 install -d %{buildroot}%{_datadir}/copr/dist_git
 install -d %{buildroot}%{_sysconfdir}/copr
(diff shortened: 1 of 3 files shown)


dist-git/copr-dist-git.spec
```diff
@@ -106,8 +110,8 @@ install -m0644 -D conf/copr-dist-git.sysusers.conf %{buildroot}%{_sysusersdir}/c
 
 %files
 %license LICENSE
-%python3_sitelib/copr_dist_git
-%python3_sitelib/copr_dist_git*egg-info
+%{python3_sitelib}/copr_dist_git
+%{python3_sitelib}/copr_dist_git-*.dist-info/
```
praiskup, line 114: the double-curly brackets are not needed

keygen/setup.py
```diff
@@ -37,6 +37,3 @@
     include_package_data=True,
     zip_safe=False,
 )
-
-#TODO: fix dirty cleanup
-shutil.rmtree("src/copr_keygen.egg-info")
```
praiskup, line 42: We can stop including shutil (per pylint warning).

dist-git/copr-dist-git.spec
xsuchy: but they are cool :)

keygen/setup.py
xsuchy: Right. Removed.

PR #3870, opened 2025-09-09
frontend, backend: implement branch-fedora support for Pulp projects

FrostyX:
See #3497

Regardless of whether the project uses backend or Pulp as a storage, we always need to branch the data on backend because we have have logs, info files, etc there.


praiskup:
> In production we shouldn't need to run it for the second time. But this is
useful for development.

:) it took years to do this right for the old storage, so heh, we'll very likely re-run the script even in production while hot-fixing for f44 branching

backend/copr_backend/actions.py
```diff
@@ -391,12 +396,18 @@ def run(self):
                 self.log.info("Create directory: %s", chrootdir)
                 os.makedirs(chrootdir)
 
+            if isinstance(self.storage, PulpStorage):
```
praiskup, line 399: Just a note -- these `isinstance()` checks are not nice for OOP.  It would be much nicer if we could live without them and modify the interface of the storage classes (even if one storage keeps the method no-op), you can just do things like `self.storage.the_method(data)`.

FrostyX: I agree, at some places I just didn't want to eyeball the abstractions. Once we are 100% Pulp compatible, we could do some final touches and remove all those `isinstance` conditions.

PR #3896, opened 2025-09-17
backend: change-storage: fix sub-project handling

praiskup:


backend/run/copr-change-storage
```diff
@@ -182,8 +182,8 @@ def main():
                 all_package_hrefs.extend(package_hrefs)
 
             # Add build results to the repository
-            if not storage.create_repository_version(project, chroot, all_package_hrefs):
-                log.error("Failed to create repository version: %s", chroot)
+            if not storage.create_repository_version(subproject, chroot, all_package_hrefs):
+                log.error("Failed to create repository version for chroot: %s", chroot)
```
praiskup, line 186: The traceback (see commit message) is still produced here, in case the repository is really non-existing.
It's probably impossible to happen in this script, but from an API perspective - we might want to return None instead of IndexError?

PR #3900, opened 2025-09-20
Fix duplicate NEVRA error when migrating projects to Pulp

FrostyX:
While trying to migrate the `@copr/copr-pull-requests` project to Pulp, we
encountered the following error:

    Cannot create repository version. More than one rpm.package content with the
    duplicate values for name, epoch, version, release, arch, location_href.

This is a common Copr use-case, building a package with the same NEVRA but
different content again and again, so why the error started happening only now?

Everything works perfectly when submitting consequent builds with same NEVRA and
different content because createrepo happens after every build and therefore
Pulp knows it is supposed to remove the old package from the repository and add
the new one.

However, when migrating data using the `copr-change-storage` scrpit, we upload
all the RPM packages and then do createrepo only once for all of them. But when
we want to add multiple packages at the same time, Pulp doesn't know which of
them is the old one.

In this PR I am implementing a mechanism that prevents migrating duplicate
NEVRAs into a Pulp repository.

### DIFF
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -75,7 +75,7 @@ class Storage:
         """
         raise NotImplementedError
 
-    def upload_build_results(self, chroot, results_dir, target_dir_name, max_workers=1, build_id=None):
+    def upload_build_results(self, rpm_paths, chroot, max_workers=1, build_id=None):
         """
         Add results for a new build to the storage
         """
@@ -117,6 +117,24 @@ class Storage:
         """
         Fork build results from one project to another
         """
+
+    def find_build_results(self, results_dir):
+        """
+        Return a list of absolute paths of all RPMs within this directory.
+        The results are returned in an undetermined order.
+        """
+        results = []
+        for root, _, files in os.walk(results_dir):
+            for name in files:
+                if os.path.basename(root) == "prev_build_backup":
+                    continue
+
+                if not name.endswith(".rpm"):
+                    continue
+
+                path = os.path.join(root, name)
+                results.append(path)
+        return results
 
 
 class BackendStorage(Storage):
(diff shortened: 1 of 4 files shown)


FrostyX:
The PR is a bit scary-looking, so I deployed it on STG and ran beaker tests against it. Everything passes fine.

backend/run/copr-change-storage
```diff
@@ -100,12 +100,24 @@ def add_redirect(fullname):
...
 
 
+def build_id_from_rpm_path(rpm):
+    """
+    Parse a build ID from an absolute path to an RPM package.
+    This is a naive implementation that relies on the exact directory structure
+    that we have for our results on Copr backend.
+    """
+    _chroot_path, builddir, _filename = rpm.rsplit("/", 2)
+    return str(int(builddir.split("-")[0]))
```
praiskup: Can we have a short unit-test for this method?

backend/run/copr-change-storage
```diff
@@ -177,11 +188,39 @@ def main():
...
+                        if os.path.basename(rpm) == os.path.basename(uploaded_rpm):
+                            break
+                        uploaded_rpm = None
+                    if not uploaded_rpm:
+                        continue
+
+                    # At this point, we know that a package with the same NEVRA
+                    # was already uploaded.
+                    previous_build_id = build_id_from_rpm_path(uploaded_rpm)
+                    if build_id > previous_build_id:
```
praiskup: I'd prefer if build_ids were integers, rather than strings

backend/run/copr-change-storage
```diff
@@ -177,11 +188,39 @@ def main():
...
+                        # If the already uploaded package comes from an older
+                        # build we want to exclude it from being added to the
+                        # repository
+                        del uploaded[uploaded_rpm]
```
praiskup: Can we use dictionary instead of list for this?
Seems safer and clearer to do `uploaded[basename] = pathname`

backend/copr_backend/storage.py
```diff
@@ -118,6 +118,23 @@ def fork_project(self, src_fullname, dst_fullname, builds_map):
         Fork build results from one project to another
         """
 
+    def find_build_results(self, results_dir):
+        """
+        Return a list of absolute paths of all RPMs within this directory
```
praiskup: Could we add a note explicitly stating that the os.walk() (and this method) output order isn't guaranteed? I feel like I've searched for this specific detail multiple times, and it would make my future life a lot easier! :-)

praiskup:
In general LGTM

backend/run/copr-change-storage
FrostyX: Updated

backend/run/copr-change-storage
FrostyX: Yeah, me too. Really good catch, I don't know what I was thinking. Updated.

backend/copr_backend/storage.py
FrostyX: Updated.

backend/run/copr-change-storage
FrostyX: > Can we use dictionary instead of list for this?

The `uploaded` variable already is a dictionary, not list

> Seems safer and clearer to do uploaded[basename] = pathname

I tried, but we _need_ the keys to be full paths so that we can parse build IDs from them when needed. The `del uploaded[uploaded_rpm]` should be safe though, because we don't iterate over the thing that we are deleting.

praiskup: Still, could we do something like this?
```
uploaded["tar-1.35-6.fc43.x86_64"] = {
    "build_id": 1,
    "path": "/path/to/000001-tar/tar-1.35-6.fc43.x86_64.rpm",
    "pulp_href": "the href I'm not sure about",
}
```

While on it, `uploaded` is a suboptimal name, we upload all the files, but we keep only a subset of them in the variable.

praiskup: Aha, no ... I'm just confused, it's a set of RPMs "to be uploaded", right?

praiskup: 🤷 I think I'd again prefer a very short meeting, just so we can explain to each other.

backend/tests/run/test_copr_change_storage.py
```diff
@@ -0,0 +1,31 @@
...
+import tempfile
+import importlib.util
+
+
+# This beauty is needed because we hit several import-related issues in
+# conjunction:
+# 1. Importing from a script that is not in $PYTHONPATH
+# 2. The script containing dashes instead of underscores in its name
+# 3. The script not having .py extension
+# Just pretend it's not here.
```
praiskup: Thank you for the unexpectedly hard work on the test case :heart: very appreciated

FrostyX: Funnily enough, after the refactoring we discussed, the function is not needed anymore :D

PR #3905, opened 2025-09-24
backend: log conflicting NEVRAs when uploading

praiskup:


backend/run/copr-change-storage
```diff
@@ -191,6 +191,8 @@ def main():
                         # build we don't need to bother with uploading this
                         # package
                         if build_id < uploaded[basename]["build_id"]:
+                            log.info("Skipping %s from %s, newer already uploaded",
+                                     build_id, basename)
```
FrostyX: I think you swapped the arguments

FrostyX (approved):

PR #3924, opened 2025-10-01
Add option to migrate all projects for an owner

### LINKED ISSUE #3923: Ability to migrate all projects for a given owner to Pulp
Our current `copr-change-script` works on the project-level. But our `BlockedOwnersLimit` works on the owner-level. It would make the most sense to 

1. Update config to block builds and actions for a given owner
2. Run migration for every project of this owner
3. Update the config again to remove the block

FrostyX:
Fix #3923

### DIFF
--- a/backend/run/copr-change-storage
+++ b/backend/run/copr-change-storage
@@ -9,6 +9,7 @@ import os
 import sys
 import argparse
 import logging
+from copr.v3 import Client
 from copr_common.log import setup_script_logger
 from copr_backend.helpers import BackendConfigReader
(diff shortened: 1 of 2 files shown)


FrostyX:
The CI fails in rawhide for frontend (where is only one line change) for what seems to be an unrelated reason.

praiskup:
> The CI fails in rawhide for frontend (where is only one line change) for what seems to be an unrelated reason.

Seems backend fails, too :-(

praiskup (approved):

PR #3930, opened 2025-10-03
testing-farm: skeleton for sanity tests

praiskup:


testing-farm/tests/sanity/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -eux
+
+true
```
FrostyX, line 3: Not sure if you are going to try the "all packages on host" approach or the docker-compose approach first. Just for the record

The docker-compose instructions are here 
https://frostyx.cz/posts/copr-docker-compose-without-supervisord
and the steps needs to be done are

```
$ git clone https://github.com/fedora-copr/copr.git
$ cd copr
$ docker-compose up -d
$ docker exec -it copr-frontend-1 bash
[copr-fe@frontend /]$ init-database.sh
```

And we have instructions for running beaker tests against the docker-compose here
https://docs.pagure.org/copr.copr/sanity_tests.html
I have a suspicion that `docker network connect copr_default test-env` will be `copr-default` now.

praiskup: I think I want to merge as is, because what I promised was just the skeleton for now.

Then, I'd really prefer experimenting with ~Packit~CI-built packages first (edit, we don't have packit for monorepo, yet).

FrostyX (approved):

PR #3933, opened 2025-10-05
Implement chunked uploads to Pulp

### LINKED ISSUE #3920: copr-change-storage should better re-try
```
Pulp: create_content: /api/v3/content/rpm/packages/upload/ /var/lib/copr/public_html/results/@mock/mock-deps/epel-9-aarch64/09397618-distribution-gpg-keys/distribution-gpg-keys-1.114-1.el9.noarch.rpm
  File "/usr/bin/copr-change-storage", line 251, in <module>
    main()
    ~~~~^^
  File "/usr/bin/copr-change-storage", line 201, in main
    results = storage.upload_build_results(
        to_upload,
        chroot,
        build_id=build_id,
    )
  File "/usr/lib/python3.13/site-packages/copr_backend/storage.py", line 424, in upload_build_results
    response = future.result()
  File "/usr/lib64/python3.13/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.13/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/usr/lib64/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.13/site-packages/copr_backend/storage.py", line 402, in upload_rpm
    response = self.client.create_content(path, labels)
  File "/usr/lib/python3.13/site-packages/copr_backend/pulp.py", line 363, in create_content
    package = self.send("POST", uri, data=data, files=files)
  File "/usr/lib/python3.13/site-packages/copr_backend/pulp.py", line 242, in send
    response = request.send(
        url=url,
    ...<2 lines>...
        files=files,
    )
  File "/usr/lib/python3.13/site-packages/copr_common/request.py", line 60, in send
    return self._send_request_repeatedly(url, method=method, data=data,
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                         **kwargs)
                                         ^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/copr_common/request.py", line 129, in _send_request_repeatedly
    return self._send_request(url, method=method, data=data,
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                              **kwargs)
                              ^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/copr_common/request.py", line 107, in _send_request
    raise RequestError(msg, response)
copr_common.request.RequestError: Request client error on https://mtls.internal.console.redhat.com/api/pulp/public-copr/api/v3/content/rpm/packages/upload/: 400 Bad Request

real    4m56.077s
user    0m7.487s
sys     0m9.348s
[copr@copr-be ~][
```

FrostyX:
Fix #3920


FrostyX:
I was bored and curious, so I took a swing at this.

When building, it successfully gets pass the new code and only fails when trying to create a new repository version after that. I didn't have time to continue, so I am not sure what's wrong yet. The batched upload seems to work correctly though (there is even check that sha256 on our side matches the sha256 that Pulp has after putting the chunks together).

The code still needs some polishing, though.

CC @dkliban just so you know, but I don't need any help just yet :-)

backend/copr_backend/pulp.py
```diff
@@ -363,6 +366,59 @@ def create_content(self, path, labels):
...
+        response = self.send("POST", uri, data)
+
+        upload_href = response.json()["pulp_href"]
+        upload_url = self.config["base_url"] + upload_href
+
+        # TODO
+        size = 10000
+        sha256 = hashlib.sha256()
+        chunks = enumerate(self._read_chunks(path, size=size))
+        for i, chunk in chunks:
```
praiskup: I'm not sure if there are any implications, but I'd prefer keeping the chunks in memory for as short period of time as possible, IOW, I would avoid the `chunks` variable and just iterated over a "generator" (feel free to yield `(ID, chunk)` pairs in `_read_chunks()`).

backend/copr_backend/pulp.py
```diff
@@ -363,6 +366,59 @@ def create_content(self, path, labels):
...
+            start = i * size
+            end = start + len(chunk) - 1
+            headers = {
+                "Content-Range": "bytes {0}-{1}/*".format(start, end),
+            }
+            files = {
+                "file": ("chunk{0}".format(i), io.BytesIO(chunk), "application/octet-stream"),
+            }
+            response = self.send("PUT", upload_url, files=files, headers=headers)
```
praiskup: ignored response output

backend/copr_backend/pulp.py
```diff
@@ -363,6 +366,59 @@ def create_content(self, path, labels):
...
+            sha256.update(chunk)
```
praiskup: Please in-line comment what happens.

backend/copr_backend/pulp.py
```diff
@@ -363,6 +366,59 @@ def create_content(self, path, labels):
...
+
+
+        # TODO
+        commit_url = self.config["base_url"] + upload_href + "commit/"
+        data = {
+            "sha256": sha256.hexdigest(),
```
praiskup, line 427: Now I’m a bit confused. Is PULP responsible for verifying the SHA-256? If so, I think we shouldn’t calculate the SHA-256 "per chunk", but only for the original complete file, and let PULP confirm we haven't made any mistake.

FrostyX: > Is PULP responsible for verifying the SHA-256?

Correct

> I think we shouldn’t calculate the SHA-256 "per chunk", but only for the original complete file

I thought too but all examples of how to calculate sha256 for a file did so by iterating over chunks. I guess it's possible to do at once but since we want to do this exclusively for large files, we should IMHO do chunks anyway. But if you would like to separate the logic from the upload, we could have a simple `sha256_for_file` function. We would have to iterate over the file twice (once for sha, second time when uploading) but it wouldn't be so bad, I guess.

backend/copr_backend/storage.py
```diff
@@ -399,7 +399,25 @@ def upload_rpm(self, path, labels):
...
+                if ex.response.status_code != 400:
+                    raise RequestError(str(ex)) from ex
+                self.log.info(
+                    "Standard Pulp upload failed with %s, trying chunked "
+                    "upload. The response was: %s", ex.response.status_code,
+                    ex.response.text,
+                )
+                # In case the normal upload failed because of a timeout,
+                # try again but chunked
+                response = self.client.create_content_chunked(path, labels)
```
FrostyX: This is how I would like to solve the 400 timeout issue. What do you think?

FrostyX:
Just in case we need it in the future, here is the implementation of the `upload_rpm` method that fall-backs to the chunked uploads. 

```python
def upload_rpm(self, path, labels):
    """
    Add an RPM to the storage
    """
    # When a file is large enough, go straight to the chunked upload, don't
    # even try the normal upload
    if os.path.getsize(path) > 1024 * 1024 * 80:
        response = self.client.create_content_chunked(path, labels)
    else:
        try:
            # We want to use the normal upload as much as possible
            response = self.client.create_content(path, labels)
        except RequestError as ex:
            if ex.response.status_code != 400:
                raise RequestError(str(ex)) from ex
            self.log.info(
                "Standard Pulp upload failed with %s, trying chunked "
                "upload. The response was: %s", ex.response.status_code,
                ex.response.text,
            )
            # In case the normal upload failed because of a timeout,
            # try again but chunked
            response = self.client.create_content_chunked(path, labels)

    if not response.ok:
        self.log.error("Failed to create Pulp content for: %s, %s",
                       path, response.text)
        return response

    return response
```

FrostyX:
And I updated the PR so that `create_content_chunked` is never used.

backend/copr_backend/pulp.py
FrostyX: Updated with:

```python
with open(path, "rb") as fp:
    sha256 = hashlib.file_digest(fp, "sha256")
```

praiskup (approved):

Issue #3846, opened 2025-08-28
authlib.integrations.base_client.errors.OAuthError: invalid_scope


nikromen:
at frontend:

```
Request: GET https://copr.fedorainfracloud.org/oidc_auth/?error=invalid_scope&error_description=unknown%2Bscope%2Bopenid%2Bprofile%2Bemail%2Bhttps://id.fedoraproject.org/scope/groups%2Bhttps://id.fedoraproject.org/scope/agreements%2Brequested&state=6WvZ7BMHtlVDREgdzA8I6x8q05ojiN
User: None
Traceback (most recent call last):
  File "/usr/lib/python3.13/site-packages/flask/app.py", line 880, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.13/site-packages/flask/app.py", line 865, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/views/misc.py", line 105, in oidc_auth
    oidc.copr.authorize_access_token()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/authlib/integrations/flask_client/apps.py", line 85, in authorize_access_token                                                                                               raise OAuthError(error=error, description=description)
```

praiskup:
Triage: this is from frontend.log

praiskup:
This is pretty frequent, indeed.
```
[Sun Oct 05 00:59:59.118617 2025] [wsgi:error] [pid 1476392:tid 1476398] [remote 114.119.129.171:26675] ERROR:coprs:Admin-only exception
[Sun Oct 05 00:59:59.118646 2025] [wsgi:error] [pid 1476392:tid 1476398] [remote 114.119.129.171:26675] Request: GET https://copr.fedorainfracloud.org/oidc_auth/?error=invalid_scope&error_description=unknown%2Bscope%2Bopenid%2Bprofile%2[Sun Oct 05 00:59:59.118660 2025] [wsgi:error] [pid 1476392:tid 1476398] [remote 114.119.129.171:26675] User: None
[...]

PR #3934, opened 2025-10-05
frontend: do not traceback for OAuthError

praiskup:
We are getting 'invalid_scope' for some reason, but I'm unable to reproduce it now. Perhaps if we advise our users to contact us, we'll get a usable user report in the future.

Fixes: #3846


nikromen:
So the traceback is no longer in the logs? if not, then this seems reasonable thing to do to me

PR #3942, opened 2025-10-08
doc: document we don't support high-perf s390x

praiskup:


FrostyX (approved):
Heh, the PR title got me confused. But other than that, makes sense.

praiskup:
I synced the $Title here with the commit message.

PR #3946, opened 2025-10-09
frontend: fix swapped canceled and ended builds

### LINKED ISSUE #3844: When canceling a build, the web-ui shows the task status reversed (e.g. "canceled", when in should be "failed" and vice versa)
I cancelled a job manually because there was a compile error and noticed something odd:

The build jobs that failed are now listed as "cancellated" in the webui, while the one that were still running and thus got cancelled are listed "failed". In should be the other way around.

Screenshot from https://copr.fedorainfracloud.org/coprs/g/kernel-vanilla/next/build/9498568/ (notice that the job that failed after a few minutes are marked as "cancelled") : Image

nikromen:
Fix #3844

### DIFF
--- a/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
+++ b/frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
@@ -11,7 +11,7 @@ from coprs.logic.coprs_logic import (
     MockChrootsLogic,
 )
 from coprs.exceptions import CoprHttpException, ObjectNotFound
-from coprs.helpers import streamed_json
+from coprs.helpers import streamed_json, FINISHED_STATUSES
 
 from coprs.views import misc
 from coprs.views.backend_ns import backend_ns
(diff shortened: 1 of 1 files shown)


frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -272,18 +272,18 @@ def pending_cancel_builds():
 def build_task_canceled(task_id):
     """ Report back to frontend that the task was canceled on backend """
     models.CancelRequest.query.filter_by(what=task_id).delete()
-    was_running = flask.request.json
-    if not was_running:
-        if '-' in task_id:
-            try:
-                build_chroot = BuildsLogic.get_build_task(task_id)
+    skip_those = [StatusEnum("succeeded"), StatusEnum("failed"), StatusEnum("skipped"), StatusEnum("forked")]
```
praiskup: Can you use this?
https://github.com/fedora-copr/copr/blob/5fd9fdbec96cae269b87f377a7e1a94b5a0bee0b/frontend/coprs_frontend/coprs/helpers.py#L54

praiskup:
otherwise lgtm, thank you

Issue #3875, opened 2025-09-11
Make sure copr_prune_srpms.py and copr_prune_results.py don't traceback on permission errors


FrostyX:
See also #3874

Sometimes due to our mistake, it can happen that there is some directory owned by root. The prunning scripts should just catch it and continue

```
[copr@copr-be ~][PROD]$ time /usr/bin/copr_prune_srpms.py                                                                                               08:04:22 [462/1196]
Traceback (most recent call last):                                                                                                                                         
  File "/usr/bin/copr_prune_srpms.py", line 127, in <module>                                                                                                               
    main()                                                                                                                                                                 
    ~~~~^^                                                                           
  File "/usr/bin/copr_prune_srpms.py", line 120, in main                                                                                                                   
    prune(opts.destdir, days, args.dry_run, args.stdout)                                                                                                                   
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                   
  File "/usr/bin/copr_prune_srpms.py", line 95, in prune                                                                                                                   
    shutil.rmtree(subdir)                                                            
    ~~~~~~~~~~~~~^^^^^^^^                                                            
  File "/usr/lib64/python3.13/shutil.py", line 763, in rmtree                                                                                                              
    _rmtree_safe_fd(stack, onexc)                                                                                                                                          
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^                                                    
  File "/usr/lib64/python3.13/shutil.py", line 707, in _rmtree_safe_fd                                                                                                     
    onexc(func, path, err)                                                                                                                                                 
    ~~~~~^^^^^^^^^^^^^^^^^                                                           
  File "/usr/lib64/python3.13/shutil.py", line 700, in _rmtree_safe_fd                                                                                                     
    onexc(os.unlink, fullname, err)                                                  
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^                                                  
  File "/usr/lib64/python3.13/shutil.py", line 696, in _rmtree_safe_fd               
    os.unlink(entry.name, dir_fd=topfd)                                              
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                    
PermissionError: [Errno 13] Permission denied: '/var/lib/copr/public_html/results/jplesnik/_backup_perl-test/srpm-builds/09188373'                                         
[INFO/MainProcess] process shutting down  
[...]

Issue #3874, opened 2025-09-11
The prune scripts are not sending exceptions to prune.log


praiskup:
```
[copr@copr-be ~][PROD]$ time /usr/bin/copr_prune_srpms.py                                                                                               08:04:22 [462/1196]
Traceback (most recent call last):                                                                                                                                         
  File "/usr/bin/copr_prune_srpms.py", line 127, in <module>                                                                                                               
    main()                                                                                                                                                                 
    ~~~~^^                                                                           
  File "/usr/bin/copr_prune_srpms.py", line 120, in main                                                                                                                   
    prune(opts.destdir, days, args.dry_run, args.stdout)                                                                                                                   
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                   
  File "/usr/bin/copr_prune_srpms.py", line 95, in prune                                                                                                                   
    shutil.rmtree(subdir)                                                            
    ~~~~~~~~~~~~~^^^^^^^^                                                            
  File "/usr/lib64/python3.13/shutil.py", line 763, in rmtree                                                                                                              
    _rmtree_safe_fd(stack, onexc)                                                                                                                                          
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^                                                    
  File "/usr/lib64/python3.13/shutil.py", line 707, in _rmtree_safe_fd                                                                                                     
    onexc(func, path, err)                                                                                                                                                 
    ~~~~~^^^^^^^^^^^^^^^^^                                                           
  File "/usr/lib64/python3.13/shutil.py", line 700, in _rmtree_safe_fd                                                                                                     
    onexc(os.unlink, fullname, err)                                                  
    ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^                                                  
  File "/usr/lib64/python3.13/shutil.py", line 696, in _rmtree_safe_fd               
    os.unlink(entry.name, dir_fd=topfd)                                              
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                    
PermissionError: [Errno 13] Permission denied: '/var/lib/copr/public_html/results/jplesnik/_backup_perl-test/srpm-builds/09188373'                                         
[INFO/MainProcess] process shutting down  
```

nikromen:
- we did not catch this in /var/log/copr-backend/prune.log - it wasn't logged at all.
[...]

PR #3947, opened 2025-10-09
Handle permission errors

antonpetrov145:
in this PR:
- created a new helper function for error messages
- added handling of permission errors
- continue with script execution if such errors occur 

fix for #3875 and #3874

### DIFF
--- a/backend/run/copr_prune_results.py
+++ b/backend/run/copr_prune_results.py
@@ -321,10 +321,6 @@ def clean_copr(path, days=DEF_DAYS, verbose=True):
             # Handle Permission Denied error explicitly
             LOG.error("No permission to delete path %s", dir_path)
             continue
-        except Exception as exception:  # pylint: disable=broad-exception-caught
-            # Catch and log any other exception
-            LOG.error("Exception when deleting path %s, %s", dir_path, str(exception))
-            continue
 
         # also remove the associated log in the main dir
         build_id = os.path.basename(dir_path).split('-')[0]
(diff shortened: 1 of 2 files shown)


backend/run/copr_prune_results.py
```diff
@@ -315,7 +315,16 @@ def clean_copr(path, days=DEF_DAYS, verbose=True):
...
         if verbose:
             LOG.info('Removing: %s', dir_path)
-        shutil.rmtree(dir_path)
+        try:
+            shutil.rmtree(dir_path)
+        except PermissionError:
+            # Handle Permission Denied error explicitly
+            LOG.error("No permission to delete path %s", dir_path)
+            continue
+        except Exception as exception:  # pylint: disable=broad-exception-caught
```
praiskup: I'm fine with the PermissionError above, but do we need this catch-all exception though?

I mean, if something more serious happens, it might still be wise to not-ignore the problem, and terminate the script with a traceback.

backend/run/copr_prune_srpms.py
```diff
@@ -92,7 +103,18 @@ def prune(path, days, dry_run=False, stdout=False):
...
             if not dry_run:
-                shutil.rmtree(subdir)
+                try:
+                    shutil.rmtree(subdir)
+                except PermissionError:
+                    # Handle Permission Denied error explicitly
+                    error_message = "No permission to delete path {}".format(subdir)
+                    print_remove_error(error_message, stdout)
+                    continue
+                except Exception as exception:  # pylint: disable=broad-exception-caught
```
praiskup: Same here.

praiskup:
Thank you very much for your first PR!  Very appreciated. :heart:

backend/run/copr_prune_results.py
antonpetrov145: Makes sense, I will remove that.

praiskup (approved):

PR #3950, opened 2025-10-12
python: Make copr.v3.helpers.wait official part of API

### LINKED ISSUE #3653: make `copr.v3.helpers.wait` official part of API
This comes from slack conversation - it seems like wait is ncie use-case - it does the `copr watch-build <BUILD_ID>`. But the helpers module is not documented, so we could make the wait method "official"

nikromen:
Fix #3653

### DIFF
--- a/python/copr/v3/__init__.py
+++ b/python/copr/v3/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from .helpers import config_from_file
+from .helpers import config_from_file, wait
 from .client import Client
 from .proxies import BaseProxy
 from .proxies.project import ProjectProxy
@@ -20,6 +20,7 @@ from .exceptions import (CoprException,
 
 __all__ = [
     "config_from_file",
+    "wait",
     "Client",
 
     "BaseProxy",
--- a/python/docs/ClientV3.rst
+++ b/python/docs/ClientV3.rst
@@ -38,6 +38,7 @@ Quick overview
     client_v3/data_structures.rst
     client_v3/error_handling.rst
     client_v3/pagination.rst
+    client_v3/utilities.rst
     client_v3/working_with_proxies_directly.rst
 
 
--- a/python/docs/client_v3/utilities.rst
+++ b/python/docs/client_v3/utilities.rst
@@ -1,0 +1,53 @@
+.. _utilities:
+
+Helper Functions
+================
+
+The copr.v3 module provides several utility functions that make working with builds and other operations more convenient.
+
+wait function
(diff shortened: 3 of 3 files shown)


praiskup:
LGTM, but do we have the method documented?

FrostyX:
> LGTM, but do we have the method documented?

I don't think we have, so +1 for the documentation.
Also, I tried to use the method recently and got some errors. Not sure if I was doing something wrong or if there is a bug.

praiskup:
Discussed on mtg:  Please fix `copr-cli` so it uses `wait` the same way as our plain-API-end-users should.

nikromen:
> Please fix copr-cli so it uses wait

so it seems copr-cli does not use that.

praiskup:
> so it seems copr-cli does not use that.

That explains

>  Also, I tried to use the method recently and got some errors. Not sure if I was doing something wrong or if there is a bug.

nikromen:
>  I tried to use the method recently and got some errors

could you please provide reproducer? The package seems to be working for me:

```
[root@716d247255e5 /]# python
Python 3.13.7 (main, Aug 14 2025, 00:00:00) [GCC 15.2.1 20250808 (Red Hat 15.2.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from copr.v3 import wait
>>> print(wait)
<function wait at 0x7f45b1050360>
>>> print(wait.__module__)
copr.v3.helpers
>>>
```

nikromen:
> That explains

aha, this task was about providing the wait method as part of python sdk, so users can avoid calling from copr.v3.helpers import wait, but do from copr.v3 import wait... so probably nothing related to cli

praiskup:
Check this build: https://app.readthedocs.org/projects/python-copr/builds/27646165/

I think that the documentation that we add here isn't actually added, per my local attempt:

```
python -m sphinx -T -b html -d xxx -D language=en . foo
```

Can you please check?

Note: There are many irrelevant warnings in the build process, I think these deserve a separate ticket (no need to fix them here).

nikromen:
Thanks for providing a way how to get the docs.
> Can you please check?

I can see it there

image

and the page itself:

image

praiskup:
I think I'd avoid the generated "Timeout" example, it's just unnecessary for code review here :-) and we don't provide examples for `param interval` either.

Otherwise lgtm.

FrostyX:
> I think I'd avoid the generated "Timeout" example

+1, it's basically 20 lines of saying "You can do `wait(build, timeout=3600)`"

PR #3951, opened 2025-10-12
doc: Document how PULP pruning works

### LINKED ISSUE #3496: pruning

nikromen:
Fix #3496

### DIFF
--- a/doc/user_documentation.rst
+++ b/doc/user_documentation.rst
@@ -958,6 +958,14 @@ even though that build might not be the newest one.  Also, if there are
 two builds of the same package version, it is undefined which one is going
 to be kept.
 
+.. note::
+   **For projects using the Pulp backend**, the retention policy is different, and the 14-day preservation period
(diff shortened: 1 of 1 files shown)


doc/user_documentation.rst
```diff
@@ -958,6 +958,13 @@ even though that build might not be the newest one.  Also, if there are
 two builds of the same package version, it is undefined which one is going
 to be kept.
 
+.. note::
+   For projects using the **Pulp** backend, the time-based retention policy (14 days)
+   **does not apply**. Instead, build retention is based on the number of builds
+   per package. Copr keeps up to the **5** newest package versions based on their NEVRA.
+   When a new, higher NEVRA build is submitted, the oldest one is removed to
+   maintain this limit.
```
praiskup: From the meeting: Would be nice to have a link describing the PULP feature we are using.
Another small wish:  I would underline the difference, i.e.

"the difference in PULP storage is that we don't guarantee 14 days preservation period, if you provide builds frequently - we end up garbage the older builds pretty quickly"

PR #3954, opened 2025-10-13
Add Pagination for PulpClient.get_content

### LINKED ISSUE #3880: PULP cli get_content() implement pagination
There's this TODO, and I fear some list of artifacts may easily reach this 1000-limit.
https://github.com/fedora-copr/copr/blob/acc07a29da68f126f5704e0d693796d43997237e/backend/copr_backend/pulp.py#L475-L478

nikromen:
note: ready for initial review, however I did not test this on stg yet

Fix #3880

### DIFF
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -14,6 +14,37 @@ from copr_common.lock import lock, LockTimeout
 from copr_common.redis_helpers import get_redis_connection
 
 
+class PaginatedResponse:
+    """
+    A response-like object that maintains backward compatibility with PULP API responses
+    while containing all paginated results.
+    """
+    def __init__(self, all_results, original_response):
+        self._all_results = all_results
+        self._original_response = original_response
+
+        self.status_code = original_response.status_code
+        self.ok = original_response.ok
+        self.text = original_response.text
+
+    def json(self):
+        """
+        Mimic Pulp paginated response, but with all results included
+        """
+        return {
+            "count": len(self._all_results),
+            "next": None,  # All results are included
+            "previous": None,
+            "results": self._all_results
+        }
+
+    def raise_for_status(self):
+        """
+        Mimic requests.Response.raise_for_status()
+        """
+        self._original_response.raise_for_status()
+
+
 class BatchedAddRemoveContent:
     """
     Group a set of `add_and_remove` tasks for a single Pulp repository.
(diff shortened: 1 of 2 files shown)


backend/copr_backend/pulp.py
```diff
@@ -487,15 +512,49 @@ def get_content(self, build_ids, chroot=None, fields=None):
...
             query += f" AND pulp_label_select=\"chroot={chroot}\""
 
-        # Setting the limit to 1000, but in the future we should use pagination
-        params = {"q": query, "offset": 0, "limit": 1000}
-        if fields:
-            params["fields"] = ",".join(fields)
+        all_results = []
+        offset = 0
+        limit = 1000
+        total_count = None
```
nikromen: those can be configured... do we need this to be configurable? at least the `limit`?

backend/copr_backend/pulp.py
```diff
@@ -487,15 +512,49 @@ def get_content(self, build_ids, chroot=None, fields=None):
...
+            next_url = data.get("next")
+            if not next_url or len(results) == 0:
+                break
+
+            offset += limit
+            page_num += 1
+
+        self.log.info("Pulp: get_content: fetched %d items across %d pages", len(all_results), page_num)
+
+        return PaginatedResponse(all_results, total_count or len(all_results), response)
```
praiskup: Can't we keep the interface of the method, i.e., just return the same data format as before, just merging the results?

backend/copr_backend/pulp.py
```diff
@@ -487,15 +512,49 @@ def get_content(self, build_ids, chroot=None, fields=None):
...
+            if page_num == 1:
+                self.log.info("Pulp: get_content: %s, query = %s", uri, query)
+            else:
+                self.log.debug("Pulp: get_content: fetching page %d (offset=%d)", page_num, offset)
+
+            response = self.send("GET", uri)
+
+            if not response.ok:
+                self.log.error("Failed to fetch content page %d: %s", page_num, response.text)
+                return response
```
FrostyX: We are using the `SafeRequest` abstraction, so IMHO all non-ok status codes should raise an exception. I am not sure if this condition is ever true. If yes, should we raise an exception instead of returning the response?

backend/copr_backend/pulp.py
```diff
@@ -487,15 +512,49 @@ def get_content(self, build_ids, chroot=None, fields=None):
...
 
-        # Setting the limit to 1000, but in the future we should use pagination
-        params = {"q": query, "offset": 0, "limit": 1000}
-        if fields:
-            params["fields"] = ",".join(fields)
+        all_results = []
+        offset = 0
+        limit = 1000
+        total_count = None
+        page_num = 1
```
FrostyX: I don't think we need the variable. We basically only use it to distinguish the first run of the loop, so that we can print a different log text. I'd probably simplify the code and drop the distinction.

backend/copr_backend/pulp.py
```diff
@@ -14,6 +14,31 @@
...
+    """
+    A response-like object that maintains backward compatibility with PULP API responses
+    while containing all paginated results.
+    """
+    def __init__(self, all_results, total_count, original_response):
+        self._all_results = all_results
+        self._total_count = total_count
+        self._original_response = original_response
+        self.status_code = original_response.status_code
+        self.ok = original_response.ok
```
FrostyX, line 27: The `backend/run/copr-pulp-query` script uses `response.raise_for_status()` which is not provided by this class. So either we would need to provide it here, or change the script to use `response.ok`

backend/copr_backend/pulp.py
```diff
@@ -487,15 +512,49 @@ def get_content(self, build_ids, chroot=None, fields=None):
...
+                self.log.info("Pulp: get_content: total items to fetch: %d", total_count)
+
+            all_results.extend(results)
+
+            next_url = data.get("next")
+            if not next_url or len(results) == 0:
+                break
+
+            offset += limit
+            page_num += 1
```
FrostyX: I was thinking if we should move the pagination logic to a separate function like `def all_pages(self, uri, offset=0, limit=1000):` that would do the `while` loop, etc, but after some experimentation, I think it's not as good idea as I originally though, and we should keep it as it is now.

backend/copr_backend/pulp.py
FrostyX: IMHO we don't need configurable `limit`.

backend/copr_backend/pulp.py
nikromen: Do you mean like unifying the pagination logic across components (e.g. pagination from frontend,...) to copr-common and use some default interface for pagination?

PR #3965, opened 2025-10-17
backend: don't create "spawner", "terminator" and "vmm" logs

praiskup:


FrostyX:
> don't create "spawner", "terminator" and "vmm" logs

The obvious question is, why not?

praiskup:
Good point, fixed! :)

FrostyX (approved):

PR #3972, opened 2025-10-22
Migrate from pytz to zoneinfo

nullr0ute:
Migrate the single user of pytz to zoneinfo as the later is part of the stdlib as of py3.9 so the former is now obsolete.


nullr0ute:
Thanks @FrostyX

PR #3989, opened 2025-11-04
Ignore the "Regenerate" button for Pulp projects without manual createrepo

### LINKED ISSUE #3988: New build results doesn't appear to reach Pulp repositories
We discussed this issue with @praiskup on Slack
An example of affected build is here: https://copr.fedorainfracloud.org/coprs/g/copr/copr-dev/build/9753127/

TL;DR: The backend build results directory doesn't have RPM packages anymore, but they don't appear in the Pulp repository. The `backend.log` however, says that the data was successful uploaded to Pulp. This is because for some reason the Pulp distribution for this project points to a publication and not a repository. And the reason is the "Regenerate" button.

FrostyX:
Fix #3988

### DIFF
--- a/common/copr_common/request.py
+++ b/common/copr_common/request.py
@@ -78,7 +78,7 @@ class SafeRequest:
             req_args["headers"] = headers
             req_args["timeout"] = self.timeout / 5
             method = method.lower()
-            if method in ["post", "put"]:
+            if method in ["post", "put", "patch"]:
                 req_args["data"] = data if files else json.dumps(data)
             response = requests.request(method, url, **req_args)
         except requests.RequestException as ex:


backend/copr_backend/storage.py
```diff
@@ -377,19 +377,31 @@ def init_project(self, dirname, chroot, reason=None):
...
+
+            # Only update the Pulp distribution if the project uses the manual
+            # createrepo feature. If not, do nothing.
+            # We cannot rely on the `self.devel` option because doesn't say
+            # anything about the project setting of the feature, it only says
+            # that the current action should affect the public repository. If
+            # the distribution points to a publication, than the project uses
+            # manual createrepo, if it points to a repository, it is a normal
+            # project.
+            if public_distribution["publication"]:
```
praiskup, line 390: I think I'm confused, again :-) can we have a meeting about this?
I thought we have two distribution, the devel distribution and the "production" (when we click the button, production gets promoted to devel state).
But builds should be done against "devel" distribution, which is being continuously updated.

FrostyX: I might be confused also because we changed this feature several times. But I think everything you said is correct.

The question is, what should happen when a project doesn't use manual createrepo and the Regenerate button is clicked?

- For backend storage: We want to regenerate the repo because it sometimes fixes issues
- For pulp storage: We either want to force Pulp to regenerate the repo or do nothing. I chose nothing as it was easier to implement

What we definitely don't want to do is to trigger the logic for promoting devel to production, which this PR is trying to fix.

praiskup: I'm mostly concerned about the basic design now :-) because the comment here implies there's just one distribution, instead of two?  Let's have a quick f2f chat to discuss this once we find some time.

PR #4006, opened 2025-11-12
Explain how to migrate projects to Pulp

FrostyX:
SSIA

### DIFF
--- a/doc/maintenance/pulp.rst
+++ b/doc/maintenance/pulp.rst
@@ -5,3 +5,31 @@ Pulp maintenance
 
 In case of any suspected Pulp-related issues, follow the Fedora Infra SOP
 https://docs.fedoraproject.org/en-US/infra/sysadmin_guide/copr/#_pulp_issues
+
+
+Migrate data to Pulp
+--------------------
+
+Before starting a migration, notify the user to not perform any builds or
+actions while the project is being migrated. Or outright block them in
+``/etc/copr/copr-be.conf`` like so::
+
+  # Don't run any builds or actions for these owners
+  blocked_owners =
+      frostyx,
+      praiskup
+
+Don't forget to ``systemctl restart copr-backend.target``. Please be aware that
+this doesn't stop builds and actions that were already running.
+
+To migrate a single project from the backend storage to Pulp, run the following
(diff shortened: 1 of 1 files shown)


doc/maintenance/pulp.rst
```diff
@@ -5,3 +5,30 @@ Pulp maintenance
...
+--------------------
+
+Before starting a migration, notify the user to not perform any builds or
+actions while the project is being migrated. Or outright block them in
+``/etc/copr/copr-be.conf`` like so::
+
+  # Don't run any builds or actions for these owners
+  blocked_owners =
+      frostyx,
+      praiskup
```
praiskup, line 20: We can warn that already running builds/actions are not stopped.

doc/maintenance/pulp.rst
```diff
@@ -5,3 +5,30 @@ Pulp maintenance
...
+command. You might want to prefix it with ``time``::
+
+  sudo -u copr copr-change-storage --src backend --dst pulp --project frostyx/hello
+
+To migrate all project for a specified user, run::
+
+  sudo -u copr copr-change-storage --src backend --dst pulp --owner frostyx
+
+The migration doesn't remove the original data. They are just not being used
+anymore.
```
praiskup, line 35: There's "but ....", non-RPM data...

doc/maintenance/pulp.rst
FrostyX: Good point, updated a little bit below

> Don't forget to ``systemctl restart copr-backend.target``. Please be aware that
this doesn't stop builds and actions that were already running.

doc/maintenance/pulp.rst
FrostyX: Not sure what you mean here.

praiskup: I meant the directory structure, logs, etc. These should be preserved maybe? Not sure how pruning mechanisms react on directories without rpms...

FrostyX: => document how to remove the migrated RPMs
... also make sure prunerepo won't delete logs if RPMs are not there anymore.

PR #3863, opened 2025-09-04
frontend, backend, cli: drop support for Fedora Modularity

FrostyX:
As of October 2024, all Modularity features in Copr, except for module hotfixes, have been deprecated and are planned for removal from the codebase. This decision aligns with the broader retirement of the Fedora Modularity project, which concluded with Fedora 39 and RHEL 9.

For more details on this transition, please see:
https://frostyx.cz/posts/copr-modularity-the-end-of-an-era

PR #4016, opened 2025-11-26
frontend: remove last leftovers of modularity

FrostyX:
See PR #3863


praiskup:
Copr-frontend seems FTBFS

PR #4018, opened 2025-11-26
Preload copr.forked_from.group and copr.forked_from.user

FrostyX:
See PR #3981

For some reason we need to preload also this field since F43

### DIFF
--- a/frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_table.html
+++ b/frontend/coprs_frontend/coprs/templates/coprs/detail/_builds_table.html
@@ -34,7 +34,7 @@
 
         <tr class="build-row" >
           <td data-order="{{ build.id }}">
-            <b><a href="{{ build_href_from_sql(build) }}">
+            <b><a href="{{ copr_url('coprs_ns.copr_build', copr, build_id=build.id) }}">
               {{ build.id }}
             </a></b>
           </td>


frontend/coprs_frontend/coprs/views/coprs_ns/coprs_builds.py
```diff
@@ -64,7 +64,10 @@ def render_copr_build(build_id, copr):
 @req_with_copr
 @req_with_pagination
 def copr_builds(copr, page=1):
-    _preload = copr.group, copr.forked_from, copr.dirs
+    _preload = copr.group, copr.dirs
+    _preload = copr.forked_from and copr.forked_from.group
+    _preload = copr.forked_from and copr.forked_from.user
```
praiskup, line 69: I'm not sure here, but overriding the `_preload` variable might negatively affect the objects that it referred to before

FrostyX: I doubt it. IMHO the important part is accessing those relationships and therefore forcing the query. I didn't try, but I'd expect this to work even if we didn't store the values to `_preload`

praiskup:
Per previous chat, it seems this isn't yet complete?  Some project still fails, yet gives us code 200?

FrostyX:
> Per previous chat, it seems this isn't yet complete?

It's complete. I found a project that was still tracebacking and fixed it by adding this line:

```python
 _preload = copr.forked_from and copr.forked_from.user
```

I am not aware of any project that still tracebacks because of the "preloading" issue. Frontend is/was slow, but that's a separate issue.

PR #4025, opened 2025-11-29
Fix copr-cli download for Pulp projects

### LINKED ISSUE #4014: Make sure `copr-cli download-build` works for Pulp projects
I cannot remember if we already tested this but I expect it to be broken

FrostyX:
Fix #4014

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -10,6 +10,7 @@ import re
 import subprocess
 import sys
 import time
+import json
 import warnings
 from collections import defaultdict
 from configparser import ConfigParser
@@ -675,6 +676,10 @@ class Commands(object):
         build = self.client.build_proxy.get(args.build_id)
         base_len = len(os.path.split(build.repo_url))
         build_chroots = self.client.build_chroot_proxy.get_list(args.build_id)
+        project = self.client.project_proxy.get(
+            build.ownername,
+            build.projectname,
+        )
 
         for chroot in build_chroots:
             if args.chroots and chroot.name not in args.chroots:
@@ -684,12 +689,18 @@ class Commands(object):
                 sys.stderr.write("No data for build id: {} and chroot: {}.\n".format(args.build_id, chroot.name))
                 continue
 
+            dst = os.path.join(args.dest, chroot.name)
             cmd = ['wget', '-r', '-nH', '--no-parent', '--reject', '"index.html*"', '-e', 'robots=off', '--no-verbose']
-            cmd.extend(['-P', os.path.join(args.dest, chroot.name)])
+            cmd.extend(['-P', dst])
             cmd.extend(['--cut-dirs', str(base_len + 4)])
 
             if args.rpms:
                 cmd.extend(["-A", "*.rpm"])
+
+            # Unfortunately, when the project is in Pulp, we need also
+            # `results.json` so that we know what filenames to download from Pulp
+            if args.rpms and project.storage == "pulp":
+                cmd.extend(["-A", "results.json"])
 
             if args.spec:
                 cmd.extend(["-A", "*.spec"])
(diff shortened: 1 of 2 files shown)


cli/copr_cli/main.py
```diff
@@ -713,6 +724,39 @@ def action_download_build(self, args):
...
+                # no filters were specified or `--rpms` was used, then the file
+                # exists.
+                continue
+
+            for nevra in results["packages"]:
+                filename = "{N}-{V}-{R}.{A}.rpm".format(
+                    N=nevra["name"],
+                    V=nevra["version"],
+                    R=nevra["release"],
+                    A=nevra["arch"],
```
FrostyX, line 746: I forgot on `nevra["epoch"]` ... need to add that

praiskup: Epoch is somewhat tricky, it can be number or None.

FrostyX: I found some random package in Fedora that has epoch

```
$ rpm -qip ~/Downloads/rmt-1.6-18.fc43.x86_64.rpm |grep Epoch
Epoch       : 2
```

It seems @xsuchy was right and epoch doesn't appear in filenames. Which would mean this PR is complete from my POV.

Edit: Oops, a test is failing

cli/tests/test_cli.py
```diff
@@ -375,8 +375,16 @@
 @mock.patch('copr_cli.main.subprocess')
 @mock.patch('copr.v3.proxies.build.BuildProxy.get')
 @mock.patch('copr.v3.proxies.build_chroot.BuildChrootProxy.get_list')
+@mock.patch('copr.v3.proxies.project.ProjectProxy.get')
 @mock.patch('copr_cli.main.config_from_file', return_value=mock_config)
-def test_download_build(config_from_file, build_chroot_proxy_get_list, build_proxy_get, mock_sp, capsys):
+def test_download_build(
```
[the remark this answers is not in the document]
FrostyX: I don't understand why the linter complains for 6/5 ... Shouldn't the limit be 10 according to our pylintrc?

https://github.com/fedora-copr/copr/blob/625f0907045e379789dbe5428c0d03b5c18e0dd5/pylintrc_clients#L82-L83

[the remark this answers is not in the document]
praiskup: please check `max-positional-arguments=` and set it to some (unreasonable) value

cli/tests/test_cli.py
```diff
@@ -375,8 +375,16 @@
...
+    config_from_file,
+    project_proxy_get,
```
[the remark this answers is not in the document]
praiskup: is that a fixture?

FrostyX:
I fixed the linter, PTAL.

This turned out to be a blocker for Packit, and we needed to revert back to using our backend storage instead of Pulp. It would be worth having this released soon (this week ideally), so that we can try switching Packit to Pulp again.

cli/copr_cli/main.py
```diff
@@ -713,6 +724,39 @@ def action_download_build(self, args):
...
+                )
+                url = "{baseurl}/{chroot}/Packages/{letter}/{filename}".format(
+                    baseurl=build.repo_url,
+                    chroot=chroot.name,
+                    letter=filename[0],
+                    filename=filename,
+                )
+                response = requests.get(url)
+                if not response.ok:
```
nikromen, line 755: why we do not log/raise or somehow tell that the response failed? won't this backfire to us in the future with being silent with some bugs that would normally be logged/tracebacked?

praiskup: This still deserves a comment, at least :-) I'd expect it's fair to fail here.

nikromen: well yes but we do not fail, rather ignore the error and silently continue. That's what is confusing for me

FrostyX: Sorry for the delay. Valid points here, I added an error message and reasoning for the `continue`.

PR #4031, opened 2025-12-01
doc: mention hot-fixes for migration of persistent VMs

nikromen:

### DIFF
--- a/doc/how_to_upgrade_persistent_instances.rst
+++ b/doc/how_to_upgrade_persistent_instances.rst
@@ -285,7 +285,8 @@ to start on Fedora N+2 with up-2-date set of packages, do the ``dnf update`` now
 Get it working
 --------------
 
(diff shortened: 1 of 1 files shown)


FrostyX:
Looks reasonable, but let's block this until we know what @xsuchy prepared for us regarding the hot fixing workflow.

praiskup:
My bet is that we'll have to start using a proper tagging mechanism (Tito).  I'm not against - lets formalize what we do, but it would be some additional work we have to acknowledge.

I don't think that block the merge of this simple doc-fix, though.

praiskup:
@xsuchy promised to provide a maintenance document for hot-fixing (via Tito) so this might eventually be redundant?

nikromen:
depends on how the workflow will be - we will still need to install the package containing hot-fixes if this won't be delivered via playbooks

PR #4032, opened 2025-12-02
Wait until both copr_base and copr_coprdir exist

FrostyX:
Otherwise we might start CoprDir builds sooner than their repository exists and
fail. This is especially problem for Pulp projects where the initial createrepo
takes a bit longer than our local createrepo_c on backend:

    Failed to download metadata (baseurl:
    "https://download.copr.fedorainfracloud.org/results/@copr/copr-pull-requests:pr:4016/fedora-42-x86_64/")
    for repository "copr_coprdir": Usable URL not found WARNING: DNF5 command
    failed, retrying, attempt #3, sleeping 10s

I think this started to be an issue since PR #3369

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -50,11 +50,11 @@ DATETIME_FORMAT = "%Y-%m-%d %H:%M"
 
 MESSAGES = {
     "give_up_repo":
-        "Giving up waiting for copr_base repository, "
+        "Giving up waiting for {0} repository, "
         "please try to manually regenerate the DNF repository "
         "(e.g. by 'copr-cli regenerate-repos <project_name>')",
     "repo_waiting":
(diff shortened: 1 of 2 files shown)


FrostyX:
I mainly opened this PR to start a discussion. We then talked about the issue with @praiskup and our best guess is that there is something wrong with:

https://github.com/fedora-copr/copr/blob/4239a9f3676f6c624762f87415af40b3600ca944/backend/copr_backend/storage.py#L530-L533

rpmbuild/mock.cfg.j2
```diff
@@ -54,6 +54,7 @@ skip_if_unavailable=0
 metadata_expire=0
 cost=1
 best=1
+retries=10
```
praiskup: Uh, I have never seen this in action.  Good to know this exists, but yeah ... we should take a look at background worker's waiting loop and fix it.

Issue #4034, opened 2025-12-02
ISE in worker.log


xsuchy:
I see this error in copr-be

```
[2025-12-02 15:33:35,536][ ERROR][PID:2186717][/usr/bin/copr-backend-process-build.managed.pid-2186717][background_worker_build.py:handle_task:1110] Unexpected exception
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/copr_backend/background_worker_build.py", line 1104, in handle_task
    self.handle_build()
    ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/copr_backend/background_worker_build.py", line 1096, in handle_build
    self._get_build_job()
    ~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/copr_backend/background_worker_build.py", line 391, in _get_build_job
    self.job = BuildJob(resp.json(), self.opts)
               ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/copr_backend/job.py", line 82, in __init__
    for key, val in task_data.items():
                    ^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
```

praiskup:
Triage: we are afraid this isn't the "created and immediately removed build" -> let's take a look.

praiskup:
On previous copr-backend:
```
gunzip  < worker.log-20251214.gz | grep Attribute | wc -l
49858
```

On F43 copr-backend: 
```
[root@copr-be copr-backend][PROD]# grep Attribute worker.log | wc -l
11890
```

praiskup:
```
[2025-12-14 01:02:40,071][  INFO][PID:1934585][/usr/bin/copr-backend-process-build.managed.pid-1934585][msgbus.py:__init__:157] initializing bus
[2025-12-14 01:02:40,119][ ERROR][PID:1934585][/usr/bin/copr-backend-process-build.managed.pid-1934585][background_worker_build.py:handle_task:1126] Unexpected exception
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/copr_backend/background_worker_build.py", line 1120, in handle_task
    self.handle_build()
    ~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/copr_backend/background_worker_build.py", line 1112, in handle_build
    self._get_build_job()
[...]

Issue #3715, opened 2025-04-29
Sentry integration


FrostyX:
We have Red Hat organization, creating a new project and team there is easy. Other projects use it as well, e.g. Packit, LogDetective.

praiskup:
Triage: We can configure it even if we have too many Tracebacks now (but it is better than a year back).  Let's resolve them later anyway.

praiskup:
Study, create a demo session, maybe SOP => 3 days.

PR #4036, opened 2025-12-02
do not TB when frontend retuns malformed data

xsuchy:
Fixes: #4034

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -383,12 +383,13 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
 
         try:
             resp = self.frontend_client.get(target)
-        except FrontendClientException as ex:
+            resp_json = resp.json()
+            resp_json.items() # if response is malformated, this raise AttributeError
+        except (FrontendClientException, AttributeError) as ex:
             self.log.error("Failed to download build info: %s", str(ex))
             msg = "Failed to get the build task {}".format(target)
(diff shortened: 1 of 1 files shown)


xsuchy:
This does not resolve the malformed response itself. Because I have no idea what it cause. But in such case the backend will not traceback.

backend/copr_backend/background_worker_build.py
```diff
@@ -383,12 +383,13 @@ def _get_build_job(self):
 
         try:
             resp = self.frontend_client.get(target)
-        except FrontendClientException as ex:
+            resp_json = resp.json()
+            resp_json.items() # if response is malformated, this raise AttributeError
+        except (FrontendClientException, AttributeError) as ex:
```
praiskup, line 388: If this isn't causing disasters, I would let the Traceback appear in logs.  This seems like we have a bug in copr-frontend, and probably even in FrontendClient what we would just hide here.

nikromen: well it would still raise the backend error no?

praiskup: yes, but without the useful data for debugging (log.error vs. log.exception)

praiskup: @xsuchy can you please re-consider we need this?

BTW, while discussing whether we need or don't need this - we moved the #3715 to triage.

praiskup:
standup time: we agreed on closing this

PR #4043, opened 2025-12-05
Safety-belts post migration to PULP

praiskup:


FrostyX:
We will later need this per-owner in addition to per-project. But that can be done later and we are in a rush now.

nikromen:
we do not want this anymore

PR #4057, opened 2025-12-08
backend: possibility to localy backup migrated RPMs and then delete

FrostyX:
See PR #4043

If there are NEVRA duplicities, we will upload them to Pulp but users might not be able to reach them. We can make temporary backups just to be safe.

We also need to start removing the data immediately because the disk space won't hold much longer.

### DIFF
--- a/backend/run/copr-change-storage
+++ b/backend/run/copr-change-storage
@@ -9,6 +9,8 @@ import os
 import sys
 import argparse
 import logging
+import shutil
+from pathlib import Path
 from copr.v3 import Client
 from copr_common.log import setup_script_logger
 from copr_backend.helpers import BackendConfigReader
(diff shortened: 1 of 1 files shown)


praiskup:
> We also need to start removing the data immediately because the disk space won't hold much longer.

This doesn't seem to apply anymore.  I'm for keeping the code as is, because I'm a bit lazy to review.

@nikromen If you are OK to review and take the responsibility for the risks that this might bring, I'd prefer having this at least rebased on top of `main` (I believe that would trigger the Gemini review).

FrostyX:
We really have a lot of disk space now, I am fine with closing this PR.

nikromen:
@FrostyX if even you don't see that this script is useful anymore, I'd rather close this

PR #4058, opened 2025-12-09
frontend: set the font-size to 16px

cathaysia:

### DIFF
--- a/frontend/coprs_frontend/coprs/static/copr.css
+++ b/frontend/coprs_frontend/coprs/static/copr.css
@@ -4,6 +4,7 @@
 
 body {
   background-color: #eee;
+  font-size: 0.875rem;
 }
 
 footer dt h3 {
--- a/frontend/coprs_frontend/coprs/static/css/bootstrap.css
+++ b/frontend/coprs_frontend/coprs/static/css/bootstrap.css
@@ -1012,8 +1012,8 @@ th {
   box-sizing: border-box;
 }
 html {
-  font-size: 10px;
   -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+  font-size: 1rem !important;
 }
 body {
   font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;


nikromen:
hi, any reason why?

cathaysia:
The default font size (9px?) is too small for me; it's difficult for me to read. Although I can zoom in, the page becomes cluttered when zoomed in.

Larger fonts also look more aesthetically pleasing.

praiskup:
Would you mind posting screenshots before and after the change, so it is obvious where is the difference; and we would let the community vote in the next month maybe?

> hi, any reason why?

I was asking the same question; answers to "why?" would be preferably directly in git commit message.

cathaysia:
before:
image

after:
image

FrostyX:
Agreed that our font size is probably a bit too small. But I wouldn't go to 16px, that's too much. Good for prose, but IMHO too much for the information we are showing. I'd bump it to 14px max.

But the second screenshot clearly shows that it won't be as easy as just increasing the number. We need to fix whatever page elements it breaks.

cathaysia:
what is about this:

before:
image

after:

image

```diff
- html { font-size: 10px; }
- body { font-size: 12px; }
+ body { font-size: 0.875rem; }
```


12px -> 14px

xsuchy:
I am -1 to use directly px. 
I am +1 to use "rem" sizes. 0.875 is fine with me.

frontend/coprs_frontend/coprs/static/copr.css
```diff
@@ -4,6 +4,7 @@
 
 body {
   background-color: #eee;
+  font-size: 0.875rem;
```
nikromen, line 7: yep sorry for wanting so many contradicting changes for you :D I deployed your change to our staging instance (https://copr.stg.fedoraproject.org/coprs/) and it was even smaller :/ ... I tried to put here 1.4 and deploy it and it seems OK to me (again deployed https://copr.stg.fedoraproject.org/coprs/ so you can have a look) - is that font-size ok with you?
```suggestion
  font-size: 1.4rem;
```

cathaysia: It seems to be covered by patternfly.min.css. If you disable this rule, 0.875rem looks fine.
image

cathaysia: I added a new rule to force the font size of the html to 1rem.
Can you redeploy one to see the effect? I don't know how to preview locally

praiskup:
edit: wrong ticket

frontend/coprs_frontend/coprs/static/copr.css
nikromen: for some reason this is still broken :/ https://copr.stg.fedoraproject.org/coprs/

PR #4062, opened 2025-12-09
cli: make the in-place token update the default

praiskup:

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -301,16 +301,30 @@ class Commands(object):
         """
         Simply print out the current user as defined in copr config.
         """
+
+        if not os.path.exists(args.config):
+            # We don't want to generate the file when user authenticates
+            # via gssapi, or?
+            raise CoprConfigException(f"File {args.config} not found.  Can't update.")
+
+        # We know there's a TOCTOU.  It is better to check that the file is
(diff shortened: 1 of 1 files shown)


nikromen:
the build succeeded except pending s390x which we don't have enough resources for that... let's live with that for this moment, meging

cli/copr_cli/main.py
```diff
@@ -1144,12 +1158,6 @@ def setup_parser():
     parser_new_api_token = subparsers.add_parser(
         "new-api-token",
         help="Generate a new API token. Be careful, this destroys the old one!")
-    parser_new_api_token.add_argument(
-        "--in-place",
-        "-i",
-        action="store_true",
-        help="Replace the login and token values in the Copr config",
-    )
```
praiskup, line 1152: @FrostyX we are probably going to merge this before the release tomorrow, sorry for the late announce (I realized this is needed? when I started writing release notes).

We can re-add the option later, if we really find it useful for anybody.

FrostyX: No problem. 

The option would make more sense if we didn't override the existing key, but rather create a new one while keeping the previous one working. Then you could have a use-case when you want to generate the key only to STDOUT and then maybe use it somewhere else than the system/user you are at.

But since we invalidate the previous key, it only makes sense to update the Copr config automatically.

Issue #4066, opened 2025-12-11
common: correct the SafeRequests re-tries for file uploads


praiskup:
The method works with file descriptors, upon re-try we should either seek_set to 0, or just stop wiring with FDs, and teach the method to open filenames.

praiskup:
Originally reported by @jpodivin

PR #4067, opened 2025-12-11
Safer handling for retries

jpodivin:
When attempting to send a request, we first reset pointer to the start of  the file.
The fix checks that we are actually dealing with an object that is 'file like' before trying this. Although it's unlikely that something else could get in.

This should help with #4066

### DIFF
--- a/common/copr_common/request.py
+++ b/common/copr_common/request.py
@@ -111,6 +111,8 @@ class SafeRequest:
     def _send_request_repeatedly(self, url, method, data=None, **kwargs):
         """
         Repeat the request until it succeeds, or timeout is reached.
+        After each failed upload attempt, the file pointers in
+        kwargs['files'] are reset to `0`.
         """
         sleep = self.SLEEP_INCREMENT_TIME
         start = time.time()
@@ -130,6 +132,11 @@ class SafeRequest:
             except RequestRetryError as ex:
                 self.log.warning("Retry request #%s on %s: %s", i, url,
                                  str(ex))
+                # For each file object in the dictionary we reset the pointer to start
+                # if anything happens, fail fast.
+                for file in kwargs.get("files", {}).values():
+                    file.seek(0)
+
                 time.sleep(sleep)
                 sleep += self.SLEEP_INCREMENT_TIME
 


common/copr_common/request.py
```diff
@@ -65,6 +65,17 @@ def _send_request(self, url, method, data=None, **kwargs):
...
         headers = kwargs.get("headers", None) or {}
         headers["User-Agent"] = self.user_agent
+
+        if files:
+            # For each file object in dictionary we reset the pointer to start
+            # Check that we are actually dealing with file objects before trying anything.
+            for file in files.values():
+                if hasattr(file, 'seek') and hasattr(file, 'tell'):
+                    try:
+                        file.seek(0)
```
praiskup: Could we just seek_set 0 in `except RequestRetryError as ex` below, inside `_send_request_repeatedly`?

We could avoid those hasattr() checks likely, and simply seek -> if it tracebacks, it's OK -- we have to seek, otherwise we are risking just too much.

common/copr_common/request.py
```diff
@@ -130,6 +143,17 @@
             except RequestRetryError as ex:
                 self.log.warning("Retry request #%s on %s: %s", i, url,
                                  str(ex))
+                # For each file object in the dictionary we reset the pointer to start
+                # if anything happens, fail fast.
+                for file in kwargs.get("files", {}).values():
+                    try:
+                        file.seek(0)
+                    except Exception as ex:
```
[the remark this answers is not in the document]
praiskup: I'd simply avoid the try-except block here; let the bomb explode.

common/copr_common/request.py
praiskup: Now, when `_send_request_repeatedly` is modified, we probably don't need this part?

praiskup:
Can you please fix the commit message so it starts with `common: your subject` and describe why we are doing the change in commit message?

common/copr_common/request.py
```diff
@@ -65,6 +65,7 @@ def _send_request(self, url, method, data=None, **kwargs):
 
         headers = kwargs.get("headers", None) or {}
         headers["User-Agent"] = self.user_agent
+
```
praiskup: nit, whitespace

common/copr_common/request.py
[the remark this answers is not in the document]
praiskup: ping on this, do we need to catch the exception?

common/copr_common/request.py
```diff
@@ -130,6 +132,11 @@ def _send_request_repeatedly(self, url, method, data=None, **kwargs):
             except RequestRetryError as ex:
                 self.log.warning("Retry request #%s on %s: %s", i, url,
                                  str(ex))
+                # For each file object in the dictionary we reset the pointer to start
+                # if anything happens, fail fast.
+                for file in kwargs.get("files", {}).values():
+                    file.seek(0)
```
praiskup, line 138: Per our meeting, @FrostyX promised to find an example of RPM that we re-uploaded, and we would check that the RPM was correctly uploaded to pulp (that would imply that this seek is not needed, because it is done by requests itself).

PR #4078, opened 2025-12-13
cli: colorize build statuses

FrostyX:
The `python-rich` package is available for EPEL9+ so for EPEL8 we would need to do some workaround.

### DIFF
--- a/cli/copr_cli/helpers.py
+++ b/cli/copr_cli/helpers.py
@@ -33,3 +33,23 @@ def print_project_info(project):
         additional_repos_str = " ".join(project.additional_repos)
         print("  Additional repo: {0}".format(additional_repos_str))
     print("")
+
+
+def colorize_status(status):
+    """
+    Return colorized status in the Rich Markup
+    https://rich.readthedocs.io/en/latest/markup.html
+    """
+    color = {
+        "importing": "blue",
+        "pending": "blue",
+        "starting": "yellow",
+        "running": "yellow",
+        "forked": "green",
+        "skipped": "green",
+        "failed": "red",
+        "succeeded": "green",
+        "canceled": "default",
+        "waiting": "default",
+    }.get(status, "default")
+    return "[{0}]{1}[/{0}]".format(color, status)
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -14,6 +14,8 @@ import json
 import warnings
 from collections import defaultdict
 from configparser import ConfigParser
+from rich.text import Text
+from rich.console import Console
 
 import requests
 
@@ -38,7 +40,11 @@ from copr.v3 import (
     CoprConfigException, CoprNoResultException, CoprAuthException,
 )
 from copr.v3.pagination import next_page
-from copr_cli.helpers import cli_use_output_format, print_project_info
+from copr_cli.helpers import (
+    cli_use_output_format,
+    print_project_info,
+    colorize_status,
+)
 from copr_cli.monitor import cli_monitor_parser
 from copr_cli.printers import cli_get_output_printer as get_printer
 from copr_cli.util import get_progress_callback, serializable, package_version
@@ -139,7 +145,7 @@ def requires_api_auth(func):
 
 
 class Commands(object):
-    def __init__(self, config_path):
+    def __init__(self, config_path, no_color=False):
         self.config = config_from_file(config_path)
 
         if self.config.get("gssapi") is None:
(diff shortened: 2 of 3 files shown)


FrostyX:
I was bored on a weekend and wanted to try the `rich` package. If you don't like this, we can drop it, no problem :-)

It looks really cool though.

Screenshot_2025-12-13_17-36-06

cli/copr_cli/helpers.py
```diff
@@ -33,3 +33,23 @@ def print_project_info(project):
...
+        "pending": "blue",
+        "starting": "yellow",
+        "running": "yellow",
+        "forked": "green",
+        "skipped": "green",
+        "failed": "red",
+        "succeeded": "green",
+        "canceled": "default",
+        "waiting": "default",
+    }.get(status, "default")
```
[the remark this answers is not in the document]
FrostyX: I'd rather take the performance hit here instead of moving this top-level. Or if we want to play nice, we could do pattern matching

```python
match status:
    case "importing": color = "blue"
    case "pending": color = "blue"
    case "starting": color = "yellow"
    # ...
    case _: color = "default"
```

praiskup:
> for EPEL8 we would need to do some workaround.

Or we can package it.

cli/copr_cli/main.py
```diff
@@ -1153,6 +1166,12 @@ def setup_parser():
     parser.add_argument("--version", action="version",
                         version="%(prog)s version " + package_version("copr-cli"))
 
+    parser.add_argument(
+        "--color",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
praiskup: Terminal colors are normally on by default, the frameworks detect incompatible terminals and automatically turn colors off...  if something screws up, there should be `--no-color`.

cli/copr_cli/helpers.py
```diff
@@ -33,3 +33,23 @@ def print_project_info(project):
...
+        "importing": "blue",
+        "pending": "blue",
+        "starting": "yellow",
+        "running": "yellow",
+        "forked": "green",
+        "skipped": "green",
+        "failed": "red",
+        "succeeded": "green",
+        "canceled": "default",
+        "waiting": "default",
```
praiskup, line 53: It's pity to keep canceled as "default".
Eventually, explicit "default" value doesn't make sense (it's returned if key is not found in dict).

cli/copr_cli/helpers.py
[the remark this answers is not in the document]
praiskup: Agreed, lets keep it as is.

cli/copr_cli/helpers.py
FrostyX: I wanted to use the same colors that we have on the website

Screenshot_2025-12-13_18-04-24

cli/copr_cli/main.py
FrostyX: This `argparse.BooleanOptionalAction` implements both `--color` and `--no-color` :-)

FrostyX: But as Gemini pointed out, this is available only since EPEL9+ so we will probably need to do a mutually exclusive group or something.

cli/copr_cli/helpers.py
praiskup: colors are up to you, but those fields setting "default" to "default" are redundant, right?

cli/copr_cli/main.py
praiskup: Hm, do we need the --color option?

FrostyX:
>> for EPEL8 we would need to do some workaround.
>
> Or we can package it.

Good call, it wasn't that hard
https://src.fedoraproject.org/rpms/python-rich/pull-request/16

cli/copr_cli/helpers.py
FrostyX: > but those fields setting "default" to "default" are redundant, right?

Correct, they are. I just wanted to enumerate the all of them to for easier searching and future changes. And I also wanted to be more explicit what their color is. But yea, they would be picked by the `.get(..., "default")` otherwise.

FrostyX:
The CI is still failing because the `python3-rich` for EPEL8 is still in testing but it should start working once it gets to stable. I tested locally in container and the feature works well on EPEL8.

praiskup (approved):

praiskup:
Can you please make sure that python-rich auto-detects terminal, and disables colors everywhere else? (pipe, > file redirection, etc).

FrostyX:
Piping seems to behave fine

```
[jkadlcik@hive copr]$ PYTHONPATH=$HOME/git/copr/cli/:$HOME/git/copr/python/ $HOME/git/copr/cli/copr status 9890311 > /tmp/out.txt
[jkadlcik@hive copr]$ cat /tmp/out.txt
succeeded
```

(no colors and no color sequences)

PR #4087, opened 2025-12-15
backend: skip projects with comps enabled when migrating to pulp

### LINKED ISSUE #4027: Comps support for PULP repositories
It seems people rely on this feature.  See #4026

We should probably keep such projects on Backend for a while, see  #3993.

nikromen:
~~this should in theory work, however I haven't tested it yet, so no mergy please (but ready for review)~~ works:

```
[root@copr-be-dev ~][STG]# copr-change-storage --src backend --dst pulp --project praiskup/ping
Skipping praiskup/ping - comps.xml not supported in PULP
[root@copr-be-dev ~][STG]#
```

Fix #4027

### DIFF
--- a/backend/run/copr-change-storage
+++ b/backend/run/copr-change-storage
@@ -9,7 +9,8 @@ import os
 import sys
 import argparse
 import logging
-from copr.v3 import Client
+import time
+from copr.v3 import Client, CoprRequestException
 from copr_common.log import setup_script_logger
 from copr_backend.helpers import BackendConfigReader
 from copr_backend.constants import PULP_REDIRECT_FILE
@@ -252,6 +253,29 @@ def query_project(owner, name, config):
     return project
 
 
+def has_comps(project, config):
+    """
+    Check if any chroot in the project has comps configured
+    """
+    client = Client({"copr_url": config.frontend_base_url})
+    for chroot in project.chroot_repos:
+        while True:
+            try:
+                chroot_detail = client.project_chroot_proxy.get(
+                    project.ownername, project.name, chroot
+                )
+                break
+            except CoprRequestException as ex:
+                log.warning("Retrying to get chroot %s for %s/%s: %s",
+                            chroot, project.ownername, project.name, str(ex))
+                time.sleep(5)
+
+        if chroot_detail.get("comps_name"):
+            return True
+
+    return False
+
+
 def all_projects_for_owner(owner, config):
     """
     Return full names of all projects for a given owner
@@ -299,6 +323,11 @@ def main():
                 .format(args.project, args.dst)
             )
             sys.exit(0)
+
+        if has_comps(project, config):
+            print("Skipping {0} - comps.xml not supported in PULP".format(args.project))
+            sys.exit(0)
+
         change_storage_for_project(args.project, args.dst, config)
     elif args.owner:
         projects = all_projects_for_owner(args.owner, config)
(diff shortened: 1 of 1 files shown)


backend/run/copr-change-storage
```diff
@@ -252,6 +252,24 @@ def query_project(owner, name, config):
...
+    """
+    client = Client({"copr_url": config.frontend_base_url})
+    for chroot in project.chroot_repos:
+        try:
+            chroot_detail = client.project_chroot_proxy.get(
+                project.ownername, project.name, chroot
+            )
+            if chroot_detail.get("comps_name"):
+                return True
+        except Exception as ex:
```
nikromen: will fix later

backend/run/copr-change-storage
```diff
@@ -252,6 +252,24 @@ def query_project(owner, name, config):
     return project
 
 
+def has_comps(project, config):
+    """
+    Check if any chroot in the project has comps configured
+    """
+    client = Client({"copr_url": config.frontend_base_url})
```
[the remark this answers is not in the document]
nikromen: meh.... the code already has it this way, I don't want to refactor it because of this. Either way most work is done by uploading the data to pulp itself... this improvement would be minimal in that context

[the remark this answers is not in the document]
xsuchy: This will be called alway only once. Ignoring.

backend/run/copr-change-storage
```diff
@@ -252,6 +252,24 @@ def query_project(owner, name, config):
...
+    for chroot in project.chroot_repos:
+        try:
+            chroot_detail = client.project_chroot_proxy.get(
+                project.ownername, project.name, chroot
+            )
+            if chroot_detail.get("comps_name"):
+                return True
+        except CoprRequestException as ex:
+            log.warning("Failed to query chroot %s: %s", chroot, str(ex))
+            continue
```
xsuchy: I think there should be either raise() or sys.exit(1) here.

nikromen: I rather removed the try-except... if it raises an exception, the error message will be outputted anyway

backend/run/copr-change-storage
```diff
@@ -252,6 +252,24 @@ def query_project(owner, name, config):
...
 
 
+def has_comps(project, config):
+    """
+    Check if any chroot in the project has comps configured
+    """
+    client = Client({"copr_url": config.frontend_base_url})
+    for chroot in project.chroot_repos:
+        try:
+            chroot_detail = client.project_chroot_proxy.get(
```
praiskup: This may fail for many reasons, e.g. when the frontend becomes temporarily unavailable.  Can we use the try-indefinitely approach?

nikromen: oh I thought I removed this try-except since I wanted to remove it hehe :D

isn't there high risk of the script then to become stuck with the indefinitely try? e.g. with some broken chroots or something

Also similar errors could happen during the other parts of the script (e.g. when querying projects or project details) so I thought this would be a job of the caller of this script to handle the errors - this script migrates only 1 project and we will migrate projects massively later, so can't we handle it there when migrating a lot of projects (on per project basis)?

nikromen: I mean I am not strongly against trying indefinitely, but my issue is that this should be then fixed across the script rather than only here

praiskup: When you run this script, you typically monitor what is happening... can we inform the user (us) that we re-trying indefinitely (and then, if frontend dies, we should jump on it and fix quickly).

nikromen: okay we can do that here... but we should be consistent and do it for the rest of the script as well before initiating mass migration

praiskup (approved):

PR #4100, opened 2026-01-05
Run Copr infrastructure with podman kube play

nikromen:
I know this is a really long PR :/, this is summary what is capable of:

Run the full Copr infrastructure locally with a single command. All services (frontend, backend, distgit, keygen, database, redis, resalloc, pulp) deploy as pods via podman kube play with Kustomize overlays.

Four deployment modes: `just up` - latest development packages from @copr/copr-dev; `just up-release` - stable Fedora RPMs `just up-pr` 3127 - test a specific pull request `just up-local` - live editing with host source code mounted

See other operational helpful commands by running `just help`

Per-service ConfigMaps for easy configuration without image rebuilds. Multiple Kustomize overlays (base, dev, dev-local) for different use cases.

Beaker integration test environment included - run the full test suite against the local stack via setup-local-copr.sh.

Designed as the foundation for the possibility of future OpenShift deployment.


nikromen:
whoops, missclick

beaker-tests/DockerTestEnv/setup-local-copr.sh
```diff
@@ -0,0 +1,94 @@
...
+#   setup-local-copr prod            # Fedora production
+#   setup-local-copr --auto-token    # Local + auto-generate API token
+
+set -e
+
+AUTO_TOKEN=false
+TARGET="local"
+
+for arg in "$@"; do
+    case "$arg" in
```
praiskup: nit: double quotes around case operand are weird (redundant)

beaker-tests/DockerTestEnv/setup-local-copr.sh
```diff
@@ -0,0 +1,94 @@
...
+    [ "$PROTOCOL" = "http" ] && ENCRYPTED="false"
+
+    mkdir -p ~/.config
+    cat > ~/.config/copr << COPR_EOF
+[copr-cli]
+login = $API_LOGIN
+username = $USERNAME
+token = $API_TOKEN
+copr_url = $COPR_URL
+encrypted = $ENCRYPTED
```
praiskup, line 83: I scratched my head, and opened https://github.com/fedora-copr/copr/issues/4328 :-)

beaker-tests/Sanity/copr-cli-basic-operations/helpers
```diff
@@ -14,7 +14,7 @@ EVIL_HELLO=https://github.com/fedora-copr/copr-test-sources/raw/main/evilhello-2
 SRPM_BUILDTAG=https://github.com/fedora-copr/copr-test-sources/raw/main/buildtag-0-0.src.rpm
 COPR_HELLO_GIT=https://github.com/fedora-copr/copr-test-hello.git
 COPR_HELLO_2_GIT=https://github.com/fedora-copr/copr-test-hello-2.git
-DNF_COPR_ID=tested-copr
+: "${DNF_COPR_ID:=${FRONTEND_URL#*://}}"
```
praiskup: This is no longer testing the ID substitution, though.

testing-farm/all-on-single-host.sh
```diff
@@ -8,12 +8,9 @@ cd "$tests_dir" || exit 1
 OWNER=$(copr-cli whoami)
 export OWNER
 
-FRONTEND_PUBLIC_HOST=$(curl -4 ifconfig.me)
-# copr-frontend thinks it's hosting publicly, this e.g., affects how .repo files
-# are generated.
-export FRONTEND_PUBLIC_HOST
```
praiskup, line 14: this would complicate manual running (I often do when I debug problems)

praiskup:
Can we have a better name of the directory rather than `deployment`?  I'm sure people would confuse it with some real deployment of Copr.  edit: `kube-deploy`?

nikromen:
could be, eventually I would want to merge this with openshift deployment and see whether and when we could possibly migrate to it, but that is big if... it could be renamed later then

FrostyX:
I am not -1 on the PR because it is a self-contained change that doesn't affect anything else. But the fact that we cannot call it `kubernetes` because we have one already, is IMHO a clear sign that we are not doing things correctly.

I hope that one day we will be able to refactor all of our deployment methods and share as much between them as possible. Now that I think about it, I think I had the same rant about the single machine testing-farm deployment.

nikromen:
I want this one day to swallow kubernetes and openshift. But since it is not yet ready I can't.

But I think we probably can delete the kubernetes - it was once created by someone who maintained his own copr stack and used it for their purposes - not by us. I think it is deprecated though and it can be deleted? Perhaps I should ask them

beaker-tests/DockerTestEnv/setup-local-copr.sh
nikromen: ack, let's handle this separately in #4328.

beaker-tests/Sanity/copr-cli-basic-operations/helpers
nikromen: Good catch. Reverted to `"${DNF_COPR_ID:=tested-copr}"` so it keeps the default matching the plugin config file name while allowing env var override for kube startup.

testing-farm/all-on-single-host.sh
nikromen: Right, sorry. Changed to `"${FRONTEND_PUBLIC_HOST:=$(curl -4 ifconfig.me)}"` so it preserves automatic detection for manual runs but allows override via env var when running via kube.

nikromen:
I am going to run tests (also locally against kube) to investigate whether anything is broken by my latest change

/packit test

nikromen:
both tests seems to work, @praiskup PTAL

kube-deploy/README.md
```diff
@@ -0,0 +1,76 @@
+# Run Copr infra with podman kube play
+
+Run the full COPR infrastructure locally using `podman kube play` with Kubernetes
+manifests. Designed for development, testing, and as the foundation for
+OpenShift deployment.
+
+## Quick Start
+
```
xsuchy, line 8: Bug: directory is `kube-deploy/`, not `deployment/`.
```suggestion
cd kube-deploy/
```

kube-deploy/containers/frontend/rootfs/frontend-init
```diff
@@ -0,0 +1,95 @@
...
+            "priority": 80,
+        },
+    ]
+    for dg in distgits:
+        if not DistGitInstance.query.filter(DistGitInstance.name == dg["name"]).first():
+            db.session.add(DistGitInstance(**dg))
+            print(f"Created DistGit instance '{dg['name']}'")
+
+    # Add user to groups
+    user = User.query.filter(User.username == 'jdoe').first()
```
xsuchy, line 82: Bug: This hardcodes `'jdoe'` but the bash section above uses `DEV_USER="${TEST_REMOTE_USER:-jdoe}"`. If someone sets `TEST_REMOTE_USER=alice`, this Python block still looks up `jdoe`.

Either pass `DEV_USER` as an env var and read it with `os.environ.get('DEV_USER', 'jdoe')`, or interpolate it into the heredoc (drop the `'` quotes around `PYEOF` so bash expands `$DEV_USER`).

beaker-tests/DockerTestEnv/setup-local-copr.sh
```diff
@@ -0,0 +1,94 @@
...
+        COPR_URL="https://copr.stg.fedoraproject.org"
+        ;;
+    prod|production)
+        echo "Configuring for Fedora COPR production..."
+        HOSTNAME="copr.fedorainfracloud.org"
+        PROTOCOL="https"
+        PORT="443"
+        COPR_URL="https://copr.fedorainfracloud.org"
+        ;;
+    *)
```
xsuchy, line 44: Bug: Production hostname should be `copr.fedoraproject.org`, not `copr.fedorainfracloud.org`. Same on line 47 for `COPR_URL`.

beaker-tests/Sanity/copr-cli-basic-operations/helpers
```diff
@@ -124,7 +124,7 @@ setup_checks()
         *dev-copr.devel.redhat.com) ;;
         *[0-9]*.[0-9]*.[0-9]*.[0-9]*) ;;  # single-host testing, best effort ipv4 match
         # https://docs.copr.fedorainfracloud.org/sanity_tests.html#sanity-tests
-        *frontend) ;;
+        *frontend*) ;;
```
xsuchy, line 127: This glob is too loose. `*frontend*` matches `my-frontend-test.evil.com`. Intent is to allow `frontend:5000` (with port), but this weakens the safety check.

Consider `*frontend|*frontend:*` or similar to only match the hostname optionally followed by a port.

kube-deploy/containers/backend-httpd/Containerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY rootfs/ /
+
+RUN mkdir -p /var/lib/copr/public_html/results && \
+    chown -R 987:root /var/lib/copr /var/log/nginx /var/run /etc/nginx && \
+    chmod -R g+rwX /var/lib/copr /var/log/nginx /var/run /etc/nginx
+
+USER 987
+EXPOSE 5002
+
+ENTRYPOINT []
```
xsuchy, line 18: The `__RESOLVER__` substitution depends on `/etc/resolv.conf` having a `nameserver` line. If missing (e.g., misconfigured container), `awk` returns empty, `sed` produces invalid nginx config, and nginx won't start.

Add a fallback:
```bash
CMD ["sh", "-c", "RESOLVER=$(awk '/^nameserver/{print $2; exit}' /etc/resolv.conf); RESOLVER=${RESOLVER:-127.0.0.11}; sed -i \"s/__RESOLVER__/${RESOLVER}/g\" /etc/nginx/nginx.conf && exec nginx -g 'daemon off;'"]
```

kube-deploy/containers/database/rootfs/init-db.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -f "$PGDATA/PG_VERSION" ]; then
+    echo "Existing PostgreSQL data found."
+    EXISTING_VERSION=$(cat "$PGDATA/PG_VERSION")
+    CURRENT_VERSION=$(postgres --version | grep -oP '\d+' | head -1)
+
+    if [ "$EXISTING_VERSION" != "$CURRENT_VERSION" ]; then
+        echo "WARNING: Data version ($EXISTING_VERSION) doesn't match PostgreSQL version ($CURRENT_VERSION)"
+        echo "Cleaning old data and reinitializing..."
+        rm -rf "$PGDATA"/*
```
xsuchy, line 17: This silently destroys all database data on PostgreSQL major version mismatch. A developer upgrading their Fedora (which pulls a new PostgreSQL) loses all their local Copr data without warning.

Consider requiring an explicit env var (e.g., `FORCE_REINIT=1`) or at least exiting with clear instructions to back up first.

kube-deploy/containers/database/rootfs/init-db.sh
```diff
@@ -0,0 +1,60 @@
...
+maintenance_work_mem = 64MB
+checkpoint_completion_target = 0.9
+log_min_duration_statement = 500
+max_connections = 100
+EOF
+
+pg_ctl -D "$PGDATA" -w start
+
+psql -c "CREATE USER \"$POSTGRESQL_USER\" WITH PASSWORD '$POSTGRESQL_PASSWORD';"
+psql -c "CREATE DATABASE \"$POSTGRESQL_DATABASE\" OWNER \"$POSTGRESQL_USER\" ENCODING 'UTF-8';"
```
xsuchy, line 52: Granting `SUPERUSER` is more privilege than Alembic needs, even for dev. This also means any SQL injection in the frontend code gets full DB access.

Alembic needs `CREATE`/`ALTER`/`DROP` on tables. Consider narrower grants, or at minimum document why SUPERUSER is required here if there's a specific migration that needs it.

kube-deploy/manifests/base/pulp.yaml
```diff
@@ -0,0 +1,83 @@
...
+      component: pulp
+  template:
+    metadata:
+      labels:
+        app: copr
+        component: pulp
+    spec:
+      containers:
+        - name: pulp
+          image: ghcr.io/pulp/pulp:latest
```
xsuchy, line 24: Unpinned `:latest` tag. Pulp breaking changes will randomly break the deployment. Pin to a specific version (e.g., `ghcr.io/pulp/pulp:3.xx`).

kube-deploy/containers/frontend/rootfs/frontend-init
```diff
@@ -0,0 +1,95 @@
...
+    fedora-43-x86_64 fedora-43-i386 fedora-43-aarch64 \
+    fedora-43-ppc64le fedora-43-s390x \
+    fedora-42-x86_64 fedora-42-i386 fedora-42-aarch64 \
+    fedora-41-x86_64 fedora-41-i386 fedora-41-aarch64 \
+    fedora-rawhide-x86_64 fedora-rawhide-i386 fedora-rawhide-aarch64 \
+    epel-9-x86_64 epel-9-aarch64 epel-9-ppc64le epel-9-s390x \
+    epel-8-x86_64 epel-8-aarch64 epel-8-ppc64le \
+    epel-7-x86_64 \
+    epel-6-x86_64; do
+    ./manage.py create-chroot "$chroot" 2>/dev/null || true
```
xsuchy, line 31: The `2>/dev/null || true` pattern on `create-chroot` is fine, but the same pattern on `create-db` (line 19, via the overall script structure) and `update_indexes_required` (line 34) masks real failures. If `create-db` partially fails, the frontend starts with a broken database.

Consider letting `create-db` fail loudly (remove `|| true` / `2>/dev/null` from critical setup steps).

xsuchy (commented):
## Review

Solid architecture — Kustomize base+overlays, dynamic builder provisioning via resalloc mirrors production patterns, four deployment modes well thought out. Several issues worth addressing before merge.

### What looks good

- Clean Kustomize structure with base/dev/dev-local overlays
- Dynamic builder provisioning via resalloc+podman-remote — close to production patterns
- SSH key management (generated at build, mounted at runtime via initContainer)
- ConfigMap-based configuration — no image rebuilds for config changes
- init containers with proper wait loops for service dependencies
- `overlay-pr-packages` mechanism for testing specific PRs — clever
- Beaker test integration with `run-local` / `setup-local-copr` workflow

### Minor items (no inline comment)

- Several scripts use `set -e` but not `set -u` — risk of silent empty variable expansion (`run-backend`, `signd-entrypoint`, `setup-local-copr.sh`).
- Distgit Containerfile line 47 has a CMD that gets overridden by the deployment manifest — dead code.

kube-deploy/containers/keygen/Containerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN set -ex && \
+    sed -i 's/Listen 80/#Listen 80/g' /etc/httpd/conf/httpd.conf && \
+    chmod g+rwx /var/log/httpd && \
+    chmod 0755 /usr/bin/sign && \
+    ln -sf /tmp/sign.conf /etc/sign.conf && \
+    sed -i "s|getpass.getuser() != 'copr-signer'|False|" /usr/bin/gpg-copr && \
+    mkdir -p /var/run/signd /var/run/httpd && \
+    chown root:root /var/run/signd /var/run/httpd && \
+    chmod 0770 /var/run/signd /var/run/httpd && \
+    # Pre-create keygen directories with secure permissions
```
xsuchy, line 33: This string-match `sed` patch is fragile — breaks if upstream changes the expression in `gpg-copr`. Consider wrapping the binary or using an env var override if supported.

beaker-tests/DockerTestEnv/Dockerfile
```diff
@@ -9,8 +9,8 @@ RUN dnf -y update \
     && if test "$COPR_PACKAGES" = devel; then dnf -y copr enable @copr/copr-dev; fi \
     && dnf -y install hostname htop net-tools iputils vim git sudo \
               openssh-server psmisc python-jedi procps-ng findutils tmux \
```
xsuchy, line 11: `lftp` was removed — could this break existing beaker tests that depend on it? Also `texinfo` was added (line 12) without explanation — what needs it?

kube-deploy/manifests/overlays/dev/patch-resalloc.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: Deployment
+metadata:
+  name: copr-resalloc
+spec:
+  template:
+    spec:
+      containers:
+        - name: resalloc
+          securityContext:
+            seLinuxOptions:
```
xsuchy, line 15: Running resalloc as root with `spc_t` SELinux type is necessary for host podman socket access, but this security trade-off should be documented in the README (the resalloc container has full host access).

Issue #4030, opened 2025-12-01
ISE on batches page


xsuchy:
When accessing https://copr.fedorainfracloud.org/batches/detail/82634/

I see in log:
```
ERROR:coprs:Admin-only exception
Request: GET https://copr.fedorainfracloud.org/batches/detail/82634/
User: None
Traceback (most recent call last):
  File "/usr/lib/python3.14/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/lib/python3.14/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/views/batches_ns/coprs_batches.py", line 16, in coprs_batch_detail
    return render_template("batches/detail.html", batch=batch, deps=deps)
  File "/usr/lib/python3.14/site-packages/flask/templating.py", line 150, in render_template
    return _render(app, template, context)
  File "/usr/lib/python3.14/site-packages/flask/templating.py", line 131, in _render
    rv = template.render(context)
  File "/usr/lib/python3.14/site-packages/jinja2/environment.py", line 1295, in render
    self.environment.handle_exception()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/usr/share/copr/coprs_frontend/coprs/templates/batches/detail.html", line 3, in top-level template code
    {% extends "layout.html" %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/templates/layout.html", line 81, in top-level template code
    {% block body %}{% endblock %}
    ^^^^^^^^^^^^^^^^^^^
  File "/usr/share/copr/coprs_frontend/coprs/templates/batches/detail.html", line 41, in block 'body'   
    {{ builds_table(batch.builds) }}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/jinja2/runtime.py", line 784, in _invoke
[...]

PR #4117, opened 2026-01-13
frontend: use build.copr instead of copr context for batches builds

nikromen:
Commit 0f656de83 changed build URL generation from build_href_from_sql() to copr_url(), but with the whole context variable `copr`. This fixes undefined error when viewing batch details page where no copr context available exists.

Fixes: #4030


nikromen:
works on stg

FrostyX (approved):

PR #4137, opened 2026-01-26
docs: document pulp behavior with same NVRA but different epoch packages

### LINKED ISSUE #4041: The same NVRA but bumped epoch, what is PULP saying on this?
Take a look at the behavior, and find a cure if needed.

nikromen:
Fix #4041

### DIFF
--- a/doc/user_documentation.rst
+++ b/doc/user_documentation.rst
@@ -969,6 +969,34 @@ to be kept.
 Projects that opted-in for :ref:`creating_repositories_manually`, are
 exempt from the old package removal because of technical limitations.
 
+.. _`faq-pulp-same-nvra-different-epoch`:
+
+.. rubric:: Packages with the same NVRA but different Epoch on Pulp backend :ref:`¶ <faq-pulp-same-nvra-different-epoch>`
+
+**For projects using the Pulp backend**, there is a known limitation when building
+packages with the same NVRA (Name-Version-Release-Architecture) but different Epoch values.
+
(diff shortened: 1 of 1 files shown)


doc/user_documentation.rst
```diff
@@ -969,6 +969,33 @@ to be kept.
...
+
+RPM package filenames are generated from NVRA only (Epoch is not included in
+the filename). For example, both ``foo-1.0-1.fc42.x86_64.rpm`` with Epoch 0 and
+the same package with Epoch 1 will have the identical filename.
+
+However, Pulp constrains repository uniqueness by NEVRA (including Epoch). This
+means two packages can co-exist in a repository with the same NVRA but different
+epochs, but they will have conflicting filenames. When this happens:
+
+- Both packages exist in the repository metadata
```
praiskup, line 987: this is an interesting point

doc/user_documentation.rst
```diff
@@ -969,6 +969,33 @@ to be kept.
...
+- Only one package file is actually accessible via the repository URL
+- The package with the more recent build time "wins" and becomes accessible
```
praiskup: I would add "... regardless the epoch value" just to underline the fact that this may lead to "package downgrades"

praiskup:
Otherwise LGTM

doc/user_documentation.rst
nikromen: that's what i get from talk with pulp folks... anyway user still can't get the package

PR #4143, opened 2026-01-28
Improve readability of chroots build graphs

### LINKED ISSUE #3325: Nicer statistics
The current plots for "builds divided by chroots" are pretty hard to read right now: too many (meaningless) colors, and labels are just one after the other. One quick and easy improvement would be to:

- put bars horizontally, with well-placed labels;
- assign meaningful colors per distribution.

Something like this:

image

In this vertical format, you can put together the two of them: last 24 hours and last 90 days. And then you can remove the tooltip (the table that shows up when the pointer is on top).

sundaram123krishnan:
Fixes #3325

### DIFF
--- a/frontend/coprs_frontend/coprs/static/js/graphs.js
+++ b/frontend/coprs_frontend/coprs/static/js/graphs.js
@@ -63,26 +63,112 @@ function lineGraph(data, ticks, bind, format) {
     var chartDay = c3.generate(chart);
 };
 
-function chrootGraph(data, bind) {
-    chart = graphConfig();
-    chart.axis.x = {show: false};
-    chart.bindto =  bind;
-    chart.data = {
-        columns: data,
-        type: 'bar'
+function chrootGraph(data, bind, options) {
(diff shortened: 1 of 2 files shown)


praiskup:
Can you please provide a screenshot? (preferably again with the imported database from Fedora Copr)

sundaram123krishnan:
There were some misalignment between the last 24h and 90 days graphs, fixed.

sundaram123krishnan:
image

praiskup:
Seems like the 24h chart isn't sorted?  And I think I'd prefer if we kept the "full chart" opt-in;  i.e., keep only the first ~15 in each chart displayed, and JS-expand the rest of the chart upon user's click?

sundaram123krishnan:
image

sundaram123krishnan:
made the changes as suggested i.e added a chart limit (defaulting to 15) and sorted both graphs in descending order

praiskup:
Nice one.  Thank you!

PR #4189, opened 2026-02-26
Adding robots.txt endpoint

jpodivin:
The API will return robots.txt from static sources, unless it doesn't exist, for whatever reason. In that case it returns robots.txt blocking all crawlers and agents.

### DIFF
--- a/frontend/coprs_frontend/coprs/static/robots.txt
+++ b/frontend/coprs_frontend/coprs/static/robots.txt
@@ -1,0 +1,97 @@
+# Well known AI agents and bots are allowed to access the website
+# with exception of the certain paths
+
+User-Agent: GPTBot
+User-Agent: ClaudeBot
+User-Agent: Claude-User
+User-Agent: Claude-SearchBot
+User-Agent: CCBot
+User-Agent: Google-Extended
+User-Agent: Applebot-Extended
+User-Agent: Facebookbot
+User-Agent: Meta-ExternalAgent
+User-Agent: Meta-ExternalFetcher
+User-Agent: diffbot
+User-Agent: PerplexityBot
+User-Agent: Perplexity‑User
+User-Agent: Omgili
+User-Agent: Omgilibot
+User-Agent: webzio-extended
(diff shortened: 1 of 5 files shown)


praiskup:
> In that case it returns robots.txt blocking all crawlers and agents.

I appreciate the work here, but we really want the (good) bots to index our pages?  Can we have some discussion about why Anubis (already implemented) isn't enough, and what problems are we solving?

frontend/coprs_frontend/coprs/static/robots.txt
```diff
@@ -0,0 +1,97 @@
+# Well known AI agents and bots are allowed to access the website
+# with exception of the certain paths
```
praiskup, line 2: It's not ideal to copy-paste this upstream, it should go to fedora infrastructure config, where already are a few of them:

```
$ find -name robots.txt*
./roles/mailman3/files/robots.txt
./roles/koji_hub/files/robots.txt
./roles/copr/dist_git/files/robots.txt
./roles/copr/frontend/files/robots.txt
./roles/people/files/robots.txt
./roles/pagure/templates/robots.txt.j2
./roles/nagios_server/files/httpd/robots.txt
./roles/httpd/website/templates/robots/robots.txt.lockbox01
./roles/httpd/website/templates/robots/robots.txt
```

Huh, and we already have something for Copr there:
```
$ find -name robots.txt* | grep copr
./roles/copr/dist_git/files/robots.txt
./roles/copr/frontend/files/robots.txt
```

praiskup: Quite an old one, for devel only:
https://forge.fedoraproject.org/infra/ansible/commit/b3b9c1fd482e4c6626453064cf7026b2e8d51d7b

@xsuchy any reasons why we should / shouldn't do this for prod?

jpodivin: If robots.txt is handled by ansible role, it makes no sense to introduce API endpoint here. I say we close this, and introduce it in the ansible role instead.

Issue #4192, opened 2026-02-28
The build of the RPM package in Fedora Copr failed, and opening builder-live.log.gz resulted in a 404 Not Found error.


EvernightFedora:
I tried built the RPM package,Build Failed and click the builder-live.log.gz get 404 not found error.

<img width="1440" height="812" alt="Image" src="https://github.com/user-attachments/assets/c993f806-1939-41db-86d3-7ac878eb8c68" />

FrostyX:
Thank you for the report @EvernightFedora,
I think this may be #4184 and/or #4182 and/or #3116. @praiskup is working on this.

FrostyX:
@praiskup, this must be related to Pulp. The initial Pulp repository for the project was not created - https://console.redhat.com/api/pulp-content/public-copr/

Does this actually have the same root cause as #4186?

mineiro:
Hi, I'm hitting what looks like the same root cause on my project (mineiro/hyprland).

I recently enabled aarch64 on it, and all builds fail immediately — the backend spends ~60s polling for the copr_base repo and then gives up:

```
 Backend process error: Giving up waiting for copr_base repository, please try to
  manually regenerate the DNF repository (e.g. by 'copr-cli regenerate-repos <project_name>')
```

The repomd.xml for the aarch64 chroots was never created — they all 404 via Pulp:

```
$ curl -s -o /dev/null -w '%{http_code}' -L \
      'https://download.copr.fedorainfracloud.org/results/mineiro/hyprland/fedora-43-aarch64/repodata/repomd.xml'
  404
```

The x86_64 chroots work fine.

I ran `copr-cli regenerate-repos mineiro/hyprland` and also tried the website "Regenerate Repositories" button — neither created the missing aarch64 repo metadata. The project is confirmed "storage": "pulp".

Example failed build: https://copr.fedorainfracloud.org/coprs/mineiro/hyprland/build/10179486/

The builder-live.log.gz is also 404 since mock never actually starts — same symptom as the OP.

Just wanted to add another data point.

Best regards!

FrostyX:
Thank you @mineiro.

Also, let me link a relevant discussion here - https://discussion.fedoraproject.org/t/failed-shortly-after-creating-a-new-copr-project-and-uploading-the-spec-file-to-start-the-build/182110

FrostyX:
Sooo, my theory is that this is in fact the Pulp timeout problem. See, we now have a lot of waiting actions:

```
[...]

Issue #4186, opened 2026-02-22
Timeouts from /api/pulp/public-copr/api/v3/content/rpm/packages


FrostyX:
Our `/var/log/copr-backend/actions.log` is full of these.

Things to notice:

- "Retry request `#736`" so this is not a temporary issue
- "read timeout=12.0" ... is our timeout 12s somewhere? I doubt it

```
[2026-02-22 17:09:02,131][WARNING][PID:1416688][/usr/bin/copr-backend-process-action.managed.pid-1416688][request.py:_send_request_repeatedly:131] Retry request #736 on https://mtls.internal.console.redhat.com/api/pulp/public-copr/api/v3/content/rpm/packages/?q=%28pulp_label_select%3D%22build_id%3D9635659%22%29+AND+pulp_label_select%3D%22chroot%3Dfedora-rawhide-x86_64%22&offset=0&limit=1000&fields=prn: Requests error on https://mtls.internal.console.redhat.com/api/pulp/public-copr/api/v3/content/rpm/packages/?q=%28pulp_label_select%3D%22build_id%3D9635659%22%29+AND+pulp_label_select%3D%22chroot%3Dfedora-rawhide-x86_64%22&offset=0&limit=1000&fields=prn: HTTPSConnectionPool(host='mtls.internal.console.redhat.com', port=443): Read timed out. (read timeout=12.0)
[2026-02-22 17:10:28,001][WARNING][PID:1205326][/usr/bin/copr-backend-process-action.managed.pid-1205326][request.py:_send_request_repeatedly:131] Retry request #514 on https://mtls.internal.console.redhat.com/api/pulp/public-copr/api/v3/content/rpm/packages/?q=%28pulp_label_select%3D%22build_id%3D10128773%22%29+AND+pulp_label_select%3D%22chroot%3Dfedora-42-aarch64%22&offset=0&limit=1000&fields=location_href: Requests error on https://mtls.internal.console.redhat.com/api/pulp/public-copr/api/v3/content/rpm/packages/?q=%28pulp_label_select%3D%22build_id%3D10128773%22%29+AND+pulp_label_select%3D%22chroot%3Dfedora-42-aarch64%22&offset=0&limit=1000&fields=location_href: HTTPSConnectionPool(host='mtls.internal.console.redhat.com', port=443): Read timed out. (read timeout=12.0)
```

FrostyX:
Interestingly, this returns the results without any problems:

```
[...]

PR #4197, opened 2026-03-03
Use SafeRequest.timeout only as a timeout per request

FrostyX:
See #4186
See #4192

We are seeing unexpected timeouts from Pulp

    Read timed out. (read timeout=12.0)

and from copr-keygen

    Read timed out. (read timeout=24.0)

It took me a long time to notice that we abuse `SafeRequest.timeout` for both timeout per request and total timeout of how long should `SafeRequest` keep trying again and again. To make it make sense, we had divide the timeout by some constant to calculate the timeout per request.

The `PulpClient` basically does this:

    request = SafeRequest(
        timeout=60,
        try_indefinitely=True,
    )

you would read it as "timeout for every request is 60s and we are going to try indefinitely". Which is incorrect. The actual timeout per request was 60 / 5 = 12.

This PR untangles this logic, and uses `SafeRequest.timeout` only as a timeout per request. How long should we try is now controlled by `SafeRequest.attempts` and for backward compatibility also `SafeRequest.try_idefinitely`.

### DIFF
--- a/common/copr_common/request.py
+++ b/common/copr_common/request.py
@@ -14,20 +14,28 @@ class SafeRequest:
     Build a HTTP request and _safely_ send it.
 
     If the server cannot be reached, the request is repeated either indefinitely
-    or until a timeout is reached.
+    or until a number of failed attempts is reached.
     """
 
     # Prolong the sleep time before asking frontend again
     SLEEP_INCREMENT_TIME = 5
 
     def __init__(self, auth=None, cert=None, log=None,
-                 try_indefinitely=False, timeout=2 * 60, user_agent=None):
+                 try_indefinitely=False, timeout=30, attempts=None,
+                 user_agent=None):
         # pylint: disable=too-many-positional-arguments
         self.auth = auth
         self.cert = cert
         self.log = log
-        self.try_indefinitely = try_indefinitely
         self.timeout = timeout
+
+        if attempts and try_indefinitely:
+            raise ValueError("Can't set both attempts and try_indefinitely")
+        if not attempts and not try_indefinitely:
+            attempts = 5
+
+        self.attempts = attempts
+        self.try_indefinitely = try_indefinitely
 
         # Use package name and version for the user agent
         self.package_name = 'copr-common'
@@ -76,7 +84,7 @@ class SafeRequest:
             elif self.auth:
                 req_args["auth"] = self.auth
             req_args["headers"] = headers
-            req_args["timeout"] = self.timeout / 5
+            req_args["timeout"] = self.timeout
             method = method.lower()
             if method in ["post", "put", "patch"]:
                 req_args["data"] = data if files else json.dumps(data)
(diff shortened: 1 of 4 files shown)


FrostyX:
This refactor is not needed. We could just do something similar to what I did in my hotfix
https://github.com/fedora-copr/copr/commit/1d3740d58c932ca6bb6b4076dd1e8900dcbc4ee3

But IMHO this version is cleaner. What do you think?

dist-git/copr-dist-git.spec
```diff
@@ -1,4 +1,4 @@
-%global copr_common_version 0.25.1~~dev0
+%global copr_common_version 1.5~~dev0
```
FrostyX: I think this is incorrect, we have new instructions now
https://docs.pagure.org/copr.copr/maintenance/development_versions.html

common/copr_common/request.py
```diff
@@ -110,19 +118,16 @@ def _send_request(self, url, method, data=None, **kwargs):
...
-
         i = 0
         while True:
             i += 1
-            if not self.try_indefinitely and time.time() > stop:
+            if not self.try_indefinitely and i > self.attempts:
                 raise RequestError(
                     "Attempt to talk to server timeouted "
-                    "(we gave it {} attempts)".format(i))
+                    "(we gave it {} attempts)".format(self.attempts))
```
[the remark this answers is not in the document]
FrostyX: Actually, the current message makes sense to me. We attempted to talk to a server. Every request timeouted. We have it 5 attempts.

common/copr_common/request.py
```diff
@@ -21,14 +21,22 @@ class SafeRequest:
...
         self.auth = auth
         self.cert = cert
         self.log = log
-        self.try_indefinitely = try_indefinitely
         self.timeout = timeout
 
+        if attempts and try_indefinitely:
+            raise ValueError("Can't set both attempts and try_indefinitely")
+        if not attempts and not try_indefinitely:
+            attempts = 5
```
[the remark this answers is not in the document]
FrostyX: I did this on purpose because zero attempts doesn't make sense.

common/copr_common/request.py
```diff
@@ -110,19 +118,16 @@ def _send_request(self, url, method, data=None, **kwargs):
...
         """
         sleep = self.SLEEP_INCREMENT_TIME
-        start = time.time()
-        stop = start + self.timeout
-
         i = 0
         while True:
             i += 1
-            if not self.try_indefinitely and time.time() > stop:
+            if not self.try_indefinitely and i > self.attempts:
```
praiskup, line 127: This is a (good, I bet) change that we should admit in docs.  We no longer implement the overall timeout feature.

dist-git/copr-dist-git.spec
praiskup: yeah, let's use the recommended pattern ... (if buggy, we can at least fix the docs).

praiskup: Even though:
> when making a release, we remove the .PATCH suffix and bump to either MAJOR+1 or MAJOR.MINOR+1).

I'm not sure who protects us from **not** doing so.

common/copr_common/request.py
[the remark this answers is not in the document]
praiskup: What if you get 500?

praiskup:
Fe need to rebase on top of #4203

melhaddadi1978-lgtm (approved):

FrostyX:
I fixed the docstring, the spec version, and rebased from `main`.

Issue #3628, opened 2025-02-16
Convert pagure.io projects to GitHub?


praiskup:
- [x] https://pagure.io/prunerepo
- [x] https://pagure.io/rpkg-util
- [x] https://pagure.io/copr/license-validate
- [ ] https://pagure.io/rhcopr/design -- can be named copr-frontend-red-template?

Documentation repos:

- [ ] https://docs.pagure.org/copr.copr/
- [ ] https://docs.pagure.org/fedora-sponsors/

FrostyX:
> Convert pagure.io projects to GitHub?

+1

Somewhere I commented that this is my preferred option, but I cannot remember where, so confirming once again.

FrostyX:
We should probably use the same migration tool as before https://fedora-copr.github.io/posts/coprs-migration-to-github

praiskup:
Meeting time: No known shut-down date for pagure.io yet, working on testing forgejo instance (we can experiment now, plans related to the hard switch are about to happen around FLOCK).
@lachmanfrantisek seems some people will likely want to keep pagure.io running forever :-(

FrostyX:
For the record here is the offiicial Fedora documentation for migrating to Forgejo
https://docs.fedoraproject.org/en-US/forge-documentation/migration/pagure/

However, I'd still prefer migrating to GitHub. No reason to have some of our projects elsewhere.

FrostyX:
I updated the issue description so that we have checkboxes. Please add anything missing.

We all agreed, we want those projects on GitHub, so I created a new migration script
https://github.com/fedora-copr/pagure-to-github
and migrated the first project (prunerepo). I'll do the rest tomorrow.

praiskup:
Great job with the migration script!

I totally forgot about the documentation.  Are we able to preserve hostnames for generated documentation?  What is your plan there?

FrostyX:
The two websites were migrated here:

- https://docs.pagure.org/copr.copr/ -> https://fedora-copr.github.io/copr/
- https://docs.pagure.org/fedora-sponsors/ -> https://frostyx.cz/fedora-sponsors/

Those URLs are not very good. I guess https://fedora-copr.github.io/copr/ is okay if we decide it is not worth any additional trouble, but something like https://docs.copr.fedorainfracloud.org/ would be better. Or any other idea?

[...]

PR #4198, opened 2026-03-03
Deploy documentation to GitHub pages

FrostyX:
See #3628

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,0 +1,57 @@
+name: Deploy documentation to GitHub Pages
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'doc/**'
+  pull_request:
+    paths:
+      - 'doc/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install make
+        run: dnf install -y make
+
+      - name: Install dependencies
+        run: yes | make deps
+        working-directory: doc
+
+      - name: Build HTML
+        run: make dirhtml
+        working-directory: doc
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: doc/_build/dirhtml
+
+  deploy:
+    if: github.ref == 'refs/heads/main'
+    needs: build
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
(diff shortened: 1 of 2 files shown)


FrostyX:
This fails because of

```
Ensure GITHUB_TOKEN has permission "id-token: write".
```

which is interesting because that didn't happen when I tried to deploy from my fork. I successfully deployed into https://frostyx.cz/copr/ ... so if we merged the PR, we would deploy into https://fedora-copr.github.io/copr . Then we could set up some redirects from the old hostname. Nirik told me that it would be possible.

Do you like this approach? Should I finish this PR?

doc/index.rst
```diff
@@ -1,6 +1,8 @@
 Copr Buildsystem
 ================
 
+Testing GitHub pages!
+
```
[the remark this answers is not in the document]
FrostyX: Yep we need to remove this before merging.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,58 @@
...
+        working-directory: doc
+
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: doc/_build/dirhtml
+
+  deploy:
+    # TODO Uncomment before merging
+    # if: github.ref == 'refs/heads/main'
```
FrostyX: We also need to uncomment this before merging.

.github/workflows/gh-pages.yml
```diff
@@ -0,0 +1,58 @@
...
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'doc/**'
+  pull_request:
+    paths:
+      - 'doc/**'
```
praiskup, line 11: does it mean that when we modify doc/ in PR, we get the build done (but not release)?  that would be great!

FrostyX: I think so, yes.

praiskup:
> which is interesting because that didn't happen when I tried to deploy from my fork. I successfully deployed into https://frostyx.cz/copr/ 

Have you deployed by a PR run?  Or after merge?  Did you have more privileges in your repo?
I think we have to make deploying a post-merge-only action (wouldn't the deploy action "just work" if we hit the merge button?).

OTOH, we should try to build the docs with every single push --force.

FrostyX:
See this commented line in the PR:

```yaml
  deploy:
    # TODO Uncomment before merging
    # if: github.ref == 'refs/heads/main'
```

I did this, and temporarily changed my permission settings to allow all PRs deploy.

Once I uncomment the line, it will behave exactly how you want. Build in every PR but then throw away. And build+deploy on push (after merging to main).

FrostyX:
I updated the PR. You can see it was built successfully but the deployment was skipped.

FrostyX:
And we are deployed here - https://fedora-copr.github.io/copr/

FrostyX:
Now available at https://docs.copr.fedorainfracloud.org/

PR #4215, opened 2026-03-10
Remove old code and remove dependency on DNF4 supporting packages

xsuchy:


rpmbuild/copr-rpmbuild.spec
```diff
@@ -151,10 +145,9 @@ Requires: yum-utils
 %latest_requires dnf5-plugins
 %endif
 
-%latest_requires python3-dnf
+%latest_requires python3-dnf5
```
praiskup, line 148: We probably want both `python3-dnf` and `python3-dnf5` here.

The point of those deps is to allow non-bootstrapped Mock biulds for older distributions.

praiskup:
> DNF4 packages will be removed soon.

Can you provide a link?

rpmbuild/copr-rpmbuild.spec
xsuchy: Updated.

FrostyX: This causes some troubles ... the package is not there?

```
[jkadlcik@hive ~]$ docker run -it fedora:43 bash
[root@3665af2ea794 /]# dnf install python3-dnf5
...
No match for argument: python3-dnf5
You can try to add to command line:
  --skip-unavailable to skip unavailable packages
```

Issue #4149, opened 2026-01-31
new-api-token: need to use --config; expiration comment in the config not being updated


brianjmurrell:
I'm getting an error trying to use the new `new-api-token` command:

```
$ copr new-api-token
Traceback (most recent call last):
  File "/usr/bin/copr", line 33, in <module>
    sys.exit(load_entry_point('copr-cli==2.4', 'console_scripts', 'copr-cli')())
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib/python3.14/site-packages/copr_cli/main.py", line 2061, in main
    getattr(commands, arg.func)(arg)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/usr/lib/python3.14/site-packages/copr_cli/main.py", line 305, in action_new_api_token
    if not os.path.exists(args.config):
           ~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "<frozen genericpath>", line 19, in exists
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

It seems that `args.config` is `None`.

Am I supposed to explicitly set the config using `--config` for this command, when I don't have to do it for any other command?

```
$ copr --config ~/.config/copr new-api-token
New API token was updated in /home/brian/.config/copr.
```

So the token is being updated but the `# expiration date: ` in the config file is not being updated so now I have no idea when my token will expire and will only know when it happens, when I can no longer use the token to refresh itself.

brianjmurrell:
Any assessment of this?

FrostyX:
I had a little bit of free time, so I submitted a PR for the first part of the issue, which is the traceback when using default config. Please see PR #4223.

I still need to do something about the `# expiration date: `

FrostyX:
I updated the PR with the `# expiration date:` fix as well.

brianjmurrell:
Nice!

nikromen:
@FrostyX can this be closed?

praiskup:
Seems so, thanks.  Please reopen if needed.

brianjmurrell:
FWIW, here's a script that can be run daily to renew your token when needed:

```bash
#!/bin/bash

set -euo pipefail

CONFIG_FILE=~/.config/copr
CURRENT_DATE=$(date +%Y-%m-%d) # Default to real system date

[...]

PR #4223, opened 2026-03-18
Don't traceback new-api-token for default config

FrostyX:
See #4149

### DIFF
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -2,6 +2,7 @@
 
 # pylint: disable=too-many-lines
 
+import io
 import argparse
 import datetime
 import logging
@@ -336,17 +337,29 @@ class Commands(object):
         self._update_config(config_path, {
             "login": result.api_login,
             "token": result.api_token,
-        })
+        }, result.expiration)
         print(f"New API token was updated in {config_path}.")
 
 
-    def _update_config(self, path, data):
+    def _update_config(self, path, data, expiration):
         if ConfigUpdater:
             config = ConfigUpdater()
             config.read(path)
             for key, value in data.items():
                 config["copr-cli"][key].value = value
             config.update_file()
+
+            with open(path, "r+") as fp:
+                content = fp.read()
+                content = re.sub(
+                    r"# expiration date: .*",
+                    f"# expiration date: {expiration}",
+                    content,
+                )
+                fp.seek(0)
+                fp.write(content)
+                fp.truncate()
+
         else:
             # This breaks the previous formatting, removes comments inside of
             # the config file, etc. That's why we only use it as a fallback.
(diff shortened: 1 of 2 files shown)


python/copr/v3/helpers.py
```diff
@@ -18,12 +21,9 @@ def __init__(self, items, meta=None, response=None, proxy=None):
...
 def config_from_file(path=None):
     raw_config = configparser.ConfigParser()
-
     config = {}
-    default_path = os.path.join("~", ".config", "copr")
-
     try:
-        exists = raw_config.read(os.path.expanduser(path or default_path))
+        exists = raw_config.read(os.path.expanduser(path or DEFAULT_CONFIG_PATH))
     except configparser.Error as ex:
```
NankoPrinzhorn, line 27: While watching you're video on Youtube, I noticed that this might be incorrect.

With the change to using DEFAULT_CONFIG_PATH. the highlighted line no longer behaves as before.

What I'm reading is a expanduser function within a expanduser function
```python
DEFAULT_CONFIG_PATH = os.path.expanduser(os.path.join("~", ".config", "copr"))
exists = raw_config.read(os.path.expanduser(path or DEFAULT_CONFIG_PATH))

// Becomes
exists = raw_config.read(os.path.expanduser(path or os.path.expanduser(os.path.join("~", ".config", "copr"))))
```

 I'm no Python developer so I could be wrong in assuming this.

FrostyX: > While watching you're video on Youtube, I noticed that this might be incorrect.

I am still amazed when people watch one of my videos and then actually comment on the PR. It's so freaking cool! Thank you :-)

> What I'm reading is a expanduser function within a expanduser function

You are right, but in this case, I think it is fine because `os.path.expanduser` is idempotent.

```python
>>> import os
>>> os.path.expanduser(os.path.expanduser("~/.config/copr")) \
...     == os.path.expanduser("~/.config/copr")
True
```

NankoPrinzhorn: In that case, i've said nothing 😃

cli/copr_cli/main.py
```diff
@@ -314,46 +316,63 @@ def action_new_api_token(self, args):
...
             config.read(path)
             for key, value in data.items():
                 config["copr-cli"][key].value = value
             config.update_file()
+
+            with open(path, "r+") as fp:
+                content = fp.read()
+                content = re.sub(
+                    r"# expiration date: .*",
+                    f"# expiration date: {expiration}",
```
nikromen, line 356: this could fail in case if user didn't copy to their conf whole string from frontend including the expiration comment (I did)

nikromen: should this be considered user error then?

FrostyX: I tried removing the expiration comment from my config, and it didn't break.

```python
>>> re.sub(r"foo .*", r"bar", "text that doesn't contain f o o")
"text that doesn't contain f o o"
```

so it should really be fine. Or am I reproducing it incorrectly?

nikromen: hmm I can't reproduce it anymore, perhaps I did some user error on my personal computer

PR #4235, opened 2026-03-24
rpmeta integration with copr backend

### LINKED ISSUE #4212: Integrate rpmeta to Copr backend for choosing the hardware resources for builds
Integrate https://github.com/fedora-copr/rpmeta as option to Copr backend to use its recommendations to assign hardware resources.

- It should be opt-in
  - We can easily turn it off in prod
- In the beginning first use case should be simple: if build would take more than ~2 hours, use powerful VM if available.

- There should be some wrapper between backend and rpmeta that translates resalloc pools to its HW info.
  - resalloc pools may be subject to a change so hardcoding it to rpmeta package (or copr package) is not a good idea
  - perhaps some wrapper that once a time gets HW info from each pool?
  - or hardcoded values config file

part of epic: https://github.com/fedora-copr/rpmeta/issues/26

nikromen:
Fix #4212

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -31,6 +31,7 @@ from copr_backend.exceptions import (
     CoprBackendError,
     FrontendClientException,
 )
+from copr_backend.rpmeta import rpmeta_predict_build_time
 from copr_backend.helpers import (
     run_cmd, register_build_result, format_evr,
 )
@@ -1032,6 +1033,8 @@ class BuildBackgroundWorker(BackendBackgroundWorker):
 
         self._wait_for_repos()
         self._cancel_if_requested()
+        self.job.rpmeta_prediction = rpmeta_predict_build_time(
+            self.job, self.opts, self.log)
         self._alloc_host()
         self._alloc_ssh_connection()
         self._check_vm()
--- a/backend/copr_backend/rpmeta.py
+++ b/backend/copr_backend/rpmeta.py
@@ -1,0 +1,158 @@
+"""
+Wrapper between copr-backend and rpmeta API.
+"""
+
+import json
+import logging
+import os
+
+import requests
+
+from copr_backend.helpers import create_file_logger
+
+
+def _get_predictions_logger(log_dir):
+    """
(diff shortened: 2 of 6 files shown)


backend/copr_backend/rpmeta.py
```diff
@@ -0,0 +1,177 @@
...
+    logger.propagate = False
+
+    if not logger.handlers:
+        path = os.path.join(log_dir, "rpmeta-predictions.log")
+        handler = logging.handlers.WatchedFileHandler(path)
+        handler.setFormatter(logging.Formatter("%(message)s"))
+        logger.addHandler(handler)
+
+    _predictions_logger = logger
+    return logger
```
FrostyX, line 24: Can we use 
https://github.com/fedora-copr/copr/blob/490137d451d7df07ba0a1f17095afca69e451830/backend/copr_backend/helpers.py#L548-L557

or `create_file_logger`?

nikromen:
ready for reviews. I still need to wait for https://forge.fedoraproject.org/infra/tickets/issues/13242 to resolve, so I can make the running instances prod(/stg)-ready. Once that ticket is resolved, I'll deploy this change to stg to test it in real action (so please don't merge yet)

nikromen:
oh I didn't realise I also requested the gemini for review, so please don't take it as it still work in progress and not ready for review, it is :D (@praiskup @FrostyX )

backend/copr_backend/background_worker_build.py
```diff
@@ -1032,6 +1033,7 @@ def build(self, attempt):
 
         self._wait_for_repos()
         self._cancel_if_requested()
+        log_rpmeta_prediction(self.job, self.opts, self.log)
```
praiskup, line 1036: This is a pretty weird call TBH :-) 

It says "log_<something>" so it seems like a no-op, and it quite likely modifies some of its arguments.

can we have something like `self.job.<something> = rpmeta_predict_<something>(package_name, log)`?

nikromen: it is now just logging, I planned to rename it in the future. But I renamed it anyway now, PTAL.

praiskup (approved):

PR #4239, opened 2026-03-26
doc: build html not dirhtml pages

FrostyX:
It is unfortunate because `dirhtml` URLs are much nicer, for comparison:

    https://docs.copr.fedorainfracloud.org/maintenance/hypervisors/
    https://docs.copr.fedorainfracloud.org/maintenance/hypervisors.html

but on https://docs.pagure.org/copr.copr we used `html` so it will probably be safer to continue using it, so that redirects from the old URL has less of a chance to break.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -6,9 +6,11 @@ on:
       - main
     paths:
       - 'doc/**'
+      - '.github/workflows/gh-pages.yml'
   pull_request:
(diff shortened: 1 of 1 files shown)


FrostyX:
> so that redirects from the old URL has less of a chance to break.

For the record, we don't have redirects yet. I will ask for them in 
https://forge.fedoraproject.org/infra/tickets/issues/13206

PR #4250, opened 2026-04-01
backend: cleanup backend-side metadata for PULP chroots

### LINKED ISSUE #4209: Do we need to remove repodata on backend for Pulp projects?
We accidentally created repodata for Pulp projects also on backend and it caused issues like #4178.
Do we need to remove them? Are they still causing any problems? 

For the record, what does `change-storage` script do? Does it leave repodata on backend too? Does this cause any problems?

praiskup:
Fixes: #4209

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -10,6 +10,7 @@ import types
 import glob
 import shlex
 import shutil
+from contextlib import suppress
 
 import configparser
 from configparser import ConfigParser
@@ -621,6 +622,15 @@ def ensure_dir_exists(path, log):
     except:
         log.exception("Can't create %s directory", path)
         raise
+
+
+def ensure_dir_removed(directory):
+    """
+    This is to avoid doing try-except for FileNotFoundError in every rmtree
+    call.
+    """
+    with suppress(FileNotFoundError):
+        shutil.rmtree(directory)
 
 
 def get_chroot_arch(chroot):
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -17,7 +17,12 @@ import shutil
 import requests
 from copr_common.enums import StorageEnum, CreaterepoReason
 from copr_common.request import RequestError, SafeRequest
-from copr_backend.helpers import call_copr_repo, build_chroot_log_name, ensure_dir_exists
+from copr_backend.helpers import (
+    build_chroot_log_name,
+    call_copr_repo,
+    ensure_dir_exists,
+    ensure_dir_removed,
+)
 from copr_backend.pulp import PulpClient
 from copr_backend.exceptions import CoprBackendError
 from .sign import resign_rpms_in_dir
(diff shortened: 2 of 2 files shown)


backend/copr_backend/helpers.py
```diff
@@ -623,6 +623,17 @@ def ensure_dir_exists(path, log):
...
 
+def ensure_dir_removed(directory):
+    """
+    This is to avoid doing try-except for FileNotFoundError in every rmtree
+    call.
+    """
+    try:
+        shutil.rmtree(directory)
+    except FileNotFoundError:
+        pass
```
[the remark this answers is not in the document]
praiskup: I didn't want to ignore ENOTDIR.

backend/copr_backend/storage.py
```diff
@@ -493,7 +509,13 @@ def create_repository_version(self, dirname, chroot, package_hrefs_or_prns):
         return self.client.add_content(repository, package_hrefs_or_prns)
 
     def publish_repository(self, chroot, **kwargs):
-        # Publishing occurs after each repository version is created.
+        """
+        PULP case: Publishing occurs after each repository version is created,
+        we only want to do some backend storage cleanup, then quit.
+        No other action needed.
+        """
+        chroot_dir = kwargs["chroot_dir"]
```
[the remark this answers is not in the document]
praiskup: I didn't want to mimic the old pattern because we recently discussed that using asserts (in non-testsuite use-cases) is not the best idea.

backend/copr_backend/helpers.py
FrostyX: Btw I recently discovered `contextlib.suppress`
https://docs.python.org/3/library/contextlib.html#contextlib.suppress
Kinda cool for situations like this.

backend/copr_backend/storage.py
```diff
@@ -485,6 +490,17 @@ def upload_build_results(self, rpm_paths, chroot, max_workers=1, build_id=None):
...
+        """
+        We know we upload stuff to PULP for given chroot, hence there's no need
+        to keep (some) data on our backend.
+
+        Chroot dir format needs to look like:
+        /var/lib/copr/public_html/results/owner/copr-dir-name/fedora-rawhide-x86_64
+        """
+        ensure_dir_removed(os.path.join(chroot_dir, "repodata"))
+        ensure_dir_removed(os.path.join(chroot_dir, "devel", "repodata"))
+
```
FrostyX, line 503: Wait ... but why do we create them in the first place?

praiskup: Do we?  Createrepo_c does, but we do not run it anymore, or?

backend/copr_backend/helpers.py
praiskup: Nice tip!  Thanks.  

Gemini, what's up with you?  You should be suggesting this.

backend/copr_backend/storage.py
FrostyX: Sorry, I must have looked incorrectly, my bad. Checking it now, and we don't.

Soo, we are doing this in the actual backend code instead of a one-off script because there, we wouldn't know what projects are in Pulp, and therefore we wouldn't know which ones to clean up?

praiskup: Yeah, the idea was, well, why not? :)

backend/copr_backend/helpers.py
praiskup: updated

backend/copr_backend/storage.py
FrostyX: The only problem I see is that this is only triggered by `publish_repository` which means any of these actions - https://docs.copr.fedorainfracloud.org/createrepo.html

so people would have to build something in the "broken" project, or do some other action, to get the repodata removed.

That's why I thought dumping the output of this

```sql
select copr.name, public.user.username, public.group.name 
from copr 
join public.user on copr.user_id=public.user.id 
left join public.group on copr.group_id = public.group.id 
where storage=1;
```

into some file on backend and having a one-off script to clean all of them would be easier.

Anyway, what if we use the code you wrote and then generate some action for all of those project to trigger the cleanup? Not sure which one though.

praiskup: > into some file on backend and having a one-off script

Are you sure we need to spend time on the script?  Since starting this task I was rather thinking this is more-or-less “speculative,” we’re essentially dealing with the question: what if the `repodata/` directory remains on the backend side?  Rather than fixing real/existing problem, or?

praiskup: > That's why I thought ... having a one-off script to clean all of them would be easier.

Also, for the record, I originally considered that approach as well, but eventually rejected it. If we decide this is a problem that needs fixing (whether via a one-off script or otherwise), there is a risk of regression. Then, that would/could result in long-term maintenance implications, such as needing to re-run the script periodically.

FrostyX:
Btw I just created a fresh project and migrated it to Pulp to reproduce some unrelated issue
https://copr.fedorainfracloud.org/coprs/frostyx/project-before-migration/

and the repodata weren't deleted during migration.
https://download.copr.fedorainfracloud.org/results/frostyx/project-before-migration/fedora-43-x86_64/repodata/

the reasoning back then was - in case we had a bug in the migration script, and it didn't feed the data to Pulp correctly, we wanted it to be very easy to just go back to using the old backend data and repository. Should we revisit that, and start removing the repodata?

praiskup:
> Should we revisit that, and start removing the repodata?

The value of metadata is zero IMO.  If we have RPMs, we can re-generate.  If we don't, it's a game over.

FrostyX (approved):

PR #4255, opened 2026-04-06
python: add type hints for project proxy

FrostyX:
I know I have been the main antagonist to introducing type hints into our code, as debated for example here
https://github.com/fedora-copr/debate/pull/3

In the meantime, I tried them on personal projects and I quite like them but I am still on the fence whether they are worth it or not. However, our python client API is the one place where it is IMHO clear-cut - as evidenced by us putting the types into docstrings. Notice that I am only adding types to function parameters and returns, not their internal code.

This commit changes only `ProjectProxy` so that we have an example. If we like it, I'll add type hints for the rest of the proxies as well.

I don't use some of the modern type hinting features to remain compatible with EPEL8 and its Python 3.6, hence `List`, `Dict`, and `Optional`.

The reason behind this effort is the Copr MCP server https://github.com/fedora-copr/copr-mcp
It is supposed to produce the best results on typed functions.

### DIFF
--- a/python/copr/test/client_v3/test_projects.py
+++ b/python/copr/test/client_v3/test_projects.py
@@ -1,0 +1,35 @@
+from copr.test import config_location, mock
+from copr.v3 import Client
+from copr.v3.proxies.project import UserPermissions
+
+
+class TestProjectProxy(object):
+    @mock.patch("copr.v3.proxies.Request.send")
+    def test_add(self, send):
+        client = Client.create_from_config_file(config_location)
+        client.project_proxy.add("user1", "foo", ["fedora-rawhide-x86_64"])
+        assert len(send.call_args_list) == 1
+        call = send.call_args_list[0]
+        args = call[1]
+        assert args["method"] == "POST"
+        assert args["params"]["ownername"] == "user1"
+        assert args["data"]["name"] == "foo"
+
+    @mock.patch("copr.v3.proxies.Request.send")
+    def test_set_permissions(self, send):
+        client = Client.create_from_config_file(config_location)
+        permissions: UserPermissions = {
+            "user1": {
+                "builder": "approved",
+                "admin": "nothing",
+            }
+        }
+        client.project_proxy.set_permissions("user1", "foo", permissions)
+        assert len(send.call_args_list) == 1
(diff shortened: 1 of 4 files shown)


praiskup:
> The reason behind this effort is the Copr MCP 

Oh.  I wouldn't expect these modern clever AI tools require us to be that "deterministic" and "explicit".

LecrisUT:
> I don't use some of the modern type hinting features to remain compatible with EPEL8 and its Python 3.6, hence `List`, `Dict`, and `Optional`.

If you don't require type annotation for older epel, you can get away with `__future__.annotations`. Caveat being if you want to run `mypy` on the specific distro packages it would not work.

praiskup:
(I'm not fan of this, at all! but...) how is this PR going?

FrostyX:
I promised I would test this doesn't break `copr-cli` for older EPELs and merge after that, but I didn't get to it yet.

FrostyX:
I tested this on AlmaLinux 8 with Python 3.6, and it works fine.

```
[jkadlcik@hive copr]$ docker run -it almalinux:8 bash
[root@a0d59f6b6be4 /]# dnf install epel-release
[root@a0d59f6b6be4 /]# dnf install https://packages.redhat.com/api/pulp-content/public-copr/@copr/copr-pull-requests:pr:4255/epel-8-x86_64/Packages/p/python3-copr-2.6.post1-1.git.112.24eae4b.el8.noarch.rpm

[root@a0d59f6b6be4 /]# dnf install python36
>>> python36
>>> from copr.v3 import Client
>>> client = Client.create_from_config_file()
>>> chroots = ["fedora-rawhide-x86_64", "fedora-rawhide-i386"]
>>> client.project_proxy.add("frostyx", "test-static-types", chroots, description="Some desc")
Munch({'name': 'test-static-types', ...})

[root@a0d59f6b6be4 /]# dnf install copr-cli
[root@a0d59f6b6be4 /]# copr create frostyx/test-static-types-2  --chroot fedora-rawhide-x86_64
New project was successfully created: https://copr.stg.fedoraproject.org/coprs/frostyx/test-static-types-2/
```

PR #4258, opened 2026-04-10
backend: modify pulp label filter for content queries

git-hyagi:
Instead of filtering build_id and chroot with separate pulp_label_select clauses joined by AND, combine them into a single comma-separated label selector.


FrostyX:
Thank you very much for the PR @git-hyagi,
I tried locally and it seems to work fine.

FrostyX (approved):

PR #4295, opened 2026-04-30
backend: fix rpmeta HW pools warning leaking to stderr

nikromen:
_load_rpmeta_hw_info() used a module-level logger that propagated to the root StreamHandler, which is the only handler available at BackendConfigReader.read() time. Remove the warning -- an empty dict already signals "predictions disabled" and the build path handles it silently.

### DIFF
--- a/backend/copr_backend/helpers.py
+++ b/backend/copr_backend/helpers.py
@@ -25,7 +25,6 @@ import subprocess
 
 import munch
 from munch import Munch
-import yaml
 
 from copr_common.redis_helpers import get_redis_connection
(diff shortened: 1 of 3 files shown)


praiskup:
Do we need to read the yaml everytime in this class?
Seems like we need this metadata on one place only in the whole copr backend, so perhaps we could delay the loading post bakcend config reader (and you could keep logging the standard way).

praiskup:
What's your thought about my previous comment?

nikromen:
it should be deplayed as you suggested @praiskup

PR #4308, opened 2026-05-13
docs: clarify that pruning is based on RPM names, not Copr packages

### LINKED ISSUE #4296: Prune does not distinguish between different packages
RPMs disappeared from the build for no apparent reason:
https://copr.fedorainfracloud.org/coprs/vgaetera/extras/build/10318933/

nikromen:
Fix #4296

### DIFF
--- a/doc/user_documentation.rst
+++ b/doc/user_documentation.rst
@@ -958,6 +958,37 @@ even though that build might not be the newest one.  Also, if there are
 two builds of the same package version, it is undefined which one is going
 to be kept.
 
+.. warning::
+   The pruner operates on **RPM package names** (the ``Name:`` field in
+   the ``.spec`` file) within each repository, **not** on Copr's "package"
+   concept visible in the web UI.
+
+   A Copr "package" (shown on the project Packages tab) is merely a source
+   definition — a recipe for how to produce builds.  Multiple Copr packages
+   can produce RPMs with the same ``Name:``, and a single Copr package can
+   produce RPMs with different names (subpackages).
+
+   The pruner that deletes old builds groups all RPMs in the repository by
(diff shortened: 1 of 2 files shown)


doc/user_documentation.rst
```diff
@@ -966,6 +991,9 @@ to be kept.
    the older builds will be removed much sooner. This behavior is controlled by Pulp's built-in
    ``retain_package_versions`` `repository setting <https://pulpproject.org/pulp_rpm/restapi/#tag/Repositories:-Rpm/operation/repositories_rpm_rpm_create>`_.
 
+   The same distinction between RPM names and Copr package names applies here
```
nikromen, line 1000: we don't have pruner for pulp yet but I assume we would want to have the same behavior? so I think it's worth to document it right now (or else we will forget), or create ticket for it

doc/user_documentation.rst
```diff
@@ -958,6 +958,31 @@ even though that build might not be the newest one.  Also, if there are
...
+   RPM named ``golang``, the pruner treats them as versions of the **same**
+   package and only keeps the latest version.
+
+.. tip::
+   If you need to maintain multiple versions of the same software
+   simultaneously, use one of these strategies:
+
+   - Create a separate Copr project for each version
+   - Or include the version in the RPM package name so the pruner sees them
+     as distinct packages
```
praiskup: Nit: I would explicitly mention the `Name:` tag here.

FrostyX: IMHO it would be nice to show

```
Name:
Version:
Release:
```

For the previous golang example.

nikromen:
updated, PTAL

Issue #4304, opened 2026-05-11
Create skill for releasing Copr RPMs


nikromen:
Once https://github.com/fedora-copr/copr/issues/4054 polishes releasing documentation, we could create skill for claude to help us with this process

FrostyX:
I've been experimenting here
https://github.com/FrostyX/skills/blob/main/copr-release-packages.md
which covers the part that is also used by the hotfix SOP

Demo is here https://www.youtube.com/watch?v=XSivixG4ruo

I don't know if it is better to instruct Claude to keep the skill in sync with the documentation or to backport some of the changes to the documentation itself and maybe put them in some AI-specific info boxes.

PR #4309, opened 2026-05-13
docs: rewrite release procedure to test before tagging

### LINKED ISSUE #4054: Create usable manual for releasing copr's rpms
currently we have https://docs.pagure.org/copr.copr/how_to_release_copr.html but IMO this is not exactly what we want. I have feeling like everyone of us have their slightly different own workflow how to do this and in current state the manual is not usable... e.g. the pre-release part states you should rebase main and do tito tag and push it to upstream, then deploy to devel and run tests. Well, in that case if tests crashes and patches are needed (90% likelihood scenario), then the damage is done and new tags on top of previous is needed. I did exactly this long time ago, then developed workflow to bypass this.

let's have in near future a brainstorm how to unify and correct the workflow so it covers what we actually want.

my topics for the brainstorm meeting:
- what is your workflow/what is my workflow
- could we have release-candidate branch?
  - we often push changes in the process of fixing tests, this is not ideal
  - by this we can decide by the time of branch from main, no new features should be accepted to that, only fixing patches
  - we could have separate project copr-release where via tito we will biuld the RPMs and deploy on devel
  - fix whatever is needed
- schedule outage could be mentioned earlier
(- the .post in tito --test is not helping either)
- another point: make it 2 week process or something like that - to avoid taking release at monday and doing it at wednesday

notes for me: koji race condition mention before pushing to stg infra tags (and mention fix); test ideally with latest copr-cli

nikromen:
I also polished a bit the document to be clear for humans but also for AI agents, so we can link this document and reuse it in https://github.com/fedora-copr/copr/issues/4304

Fix #4054

### DIFF
--- a/doc/how_to_release_copr.rst
+++ b/doc/how_to_release_copr.rst
@@ -10,6 +10,33 @@ good to know what's ahead.
 Keep amending this page if you find something not matching reality or expectations.
 
 
+.. _release_copr_projects:
+
+Copr projects used during the release
+--------------------------------------
+
+`@copr/copr-dev <https://copr.fedorainfracloud.org/coprs/g/copr/copr-dev/>`_
+    Automatic test builds from every push to ``main``.  The CI workflow
+    (``.github/workflows/copr-builds.yml``) rebuilds only the packages that
+    changed in each push (test builds, not tagged releases).  This is the repo
+    you upgrade dev machines from **before tagging**.
+
+`@copr/copr <https://copr.fedorainfracloud.org/coprs/g/copr/copr/>`_
+    Tagged release builds.  After you run ``tito tag`` and push, you manually
+    trigger builds here.  These are the packages that go into Koji / infra
+    repos / production.
+
+.. warning::
+
+   **Version skew in** ``@copr/copr-dev``:  Because CI only rebuilds changed
+   packages, the repo can accumulate builds from different commits.  For
+   example, ``python-copr-common`` might be from commit A while
+   ``copr-backend`` (which depends on it) is still from an older commit B.
+   Before relying on ``@copr/copr-dev`` for pre-release testing, always run
+   ``build_aux/rebuild-copr-stack`` to rebuild all packages in correct
+   dependency order from the current ``main`` HEAD.
+
+
 Pre-release
 -----------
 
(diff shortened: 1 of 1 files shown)


doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
...
+    Tagged release builds.  After you run ``tito tag`` and push, you manually
+    trigger builds here.  These are the packages that go into Koji / infra
+    repos / production.
 
-  git checkout main
-  git pull --rebase
+.. warning::
 
-Run::
+   **Version skew in** ``@copr/copr-dev``:  Because CI only rebuilds changed
```
nikromen, line 31: so it took me a while to dig up the possibility what screwed me that I started to use my custom release repository for this. I think the problem here described should be the issue.

The resolution here to run the rebuild copr stack command should guarantee fresh packages on top of the copr-dev repo, with this it should be always safe to use the copr-dev

doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
 
 Keep amending this page if you find something not matching reality or expectations.
 
+The procedure is organized into sequential phases (Pre-release, Release window,
+Post-release).  Within each phase, sections are meant to be followed top to
+bottom.  Each section states what to do, how to verify success, and what to do
+on failure.
```
FrostyX: This has IMHO zero value for humans, so if intended for us, I'd drop it. If needed for AI ... I was thinking ... it would IMHO be really cool if we could have some box like `.. warning::` or `.. note::` but for AI ... something like `..ai::` ... do you know if it is possible? Also, do you guys like the idea?

doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
...
-only on important things and not creating a work overload with tasks,
-that can be done post-release.
+.. _release_copr_projects:
 
+Copr projects used during the release
+--------------------------------------
 
-Tag untagged packages that have changes in them
-...............................................
+``@copr/copr-dev``
```
FrostyX: Instead of code, let's link the projects so that we can click through

doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
...
 
-    tito tag --use-version X.Y
 
-For others, new version can be bumped automatically::
+Rebuild @copr/copr-dev from current main
+.........................................
 
-    tito tag
+Make sure you are on the ``main`` branch and that it is up-to-date with
+the upstream repository (``fedora-copr/copr``)::
```
FrostyX: I'd not mention the `fedora-copr/copr` because it is obvious. And if we must, let's at least make it a link to the actual GitHub project.

doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
...
 
-Make sure that the %changelog is nice and meaningful, i.e. remove the
-`frontend:`, `rpmbuild:`, etc. prefixes and filter-out entries which are not
-interesting for the package end-users (git-log != %changelog).  Later on, if
-properly polished, the %changelogs' contents may be used for filling the Bodhi
-update text.
+  git checkout main
+  git pull --rebase origin main  # origin = fedora-copr/copr
```
FrostyX: Since we have the note right bellow this, we don't need to specify the  `# origin = fedora-copr/copr` here

doc/how_to_release_copr.rst
```diff
@@ -9,67 +9,88 @@ good to know what's ahead.
...
 
+        git remote -v  # check which remote points to fedora-copr/copr
 
-Build packages
-..............
+Ensure all PRs intended for this release are merged.  Then rebuild the entire
+stack into ``@copr/copr-dev`` in proper dependency order::
 
-Build all the updated packages into ``@copr/copr`` copr project::
+    ./build_aux/rebuild-copr-stack
```
FrostyX, line 73: If this solves the issues we were both having, then this is a major improvement.

doc/how_to_release_copr.rst
```diff
@@ -97,6 +118,31 @@ instances::
...
+   hash appears in the RPM Release field::
+
+       ./releng/install-copr-packages @copr/copr-dev <commit-sha> copr-frontend copr-backend
+
+   If you are worried about stale packages from ``@copr/copr-dev`` (e.g.
+   leftover builds from a previous release cycle), you can explicitly install
+   specific RPMs by URL::
+
+       dnf install https://download.copr.fedorainfracloud.org/results/@copr/copr-dev/fedora-NN-x86_64/NNNNN-copr-frontend/copr-frontend-X.Y-Z.fcNN.noarch.rpm
+
```
FrostyX: There are IMHO several problems with this section.

Not a regression, it's already incorrect, but there is:

```
sudo rbac-playbook -l copr-be-dev.aws.fedoraproject.org \
                   manual/copr/copr-backend-upgrade.yml
sudo rbac-playbook -l copr-be-dev.aws.fedoraproject.org groups/copr-backend.yml

...
```

That `manual/copr/*-upgrade.yml` playbooks doesn't upgrade anything because AFAIK we don't have the `@copr/copr-dev` repository enabled on our STG servers. And IIRC the consensus is that we don't want to.

So I'd vote for removing the "manual" playbooks from the instructions and keeping only the full playbooks. To be explicit, I don't mean removing the playbooks from the ansible repo, I just mean updating the instructions here.

We would use one of the methods to upgrade the packages that @nikromen is suggesting, and moving the playbook runs after that.

>  If you are worried about stale packages from ``@copr/copr-dev`` (e.g.
>   leftover builds from a previous release cycle), you can explicitly install
>   specific RPMs by URL::

Since we force the `./build_aux/rebuild-copr-stack` now, we shouldn't be worried about stale packages anymore, no? I'd drop the whole paragraph.

FrostyX (commented):
Thank you for the updates. This is definitely closer to what I am doing now and seems to be a real improvement on it (mainly the `rebuild-copr-stack` and `install-copr-packages` which I had no idea that exists)

doc/how_to_release_copr.rst
nikromen: yes this is purely for agents. maybe we could instead put this note into the future skill.md instead?

FrostyX: Ok ok, as per #4304:

> I don't know if it is better to instruct Claude to keep the skill in sync with the documentation or to backport some of the changes to the documentation itself and maybe put them in some AI-specific info boxes.

Still don't know what is the correct solution. But yeah, even if the skill was basically just "hey the documentation is at https://docs.copr.fedorainfracloud.org/how_to_release_copr.html, do it", instructing the agent to follow it from the top to bottom would IMHO make more sense in the skill  than the documentation itself.

doc/how_to_release_copr.rst
nikromen: I'll delete then the note mentioning fedora-copr/copr and keep this as is - this is useful at least for those who have fork-first workflow (e.g. my origin is nikromen/copr). I remember doing my first release, I pushed tags to my fork and did release to copr/copr from my fork :D

so I have to bear this fact in mind when doing releases. I see myself in the future, doing the release with AI not knowing this context, burning myself accidentally again. So that;s why I think it's important to have it here

doc/how_to_release_copr.rst
nikromen: yes it should, basically it mimics what I was doing but not in separate repo, rather in copr-dev (which is the ideal situation we want)

doc/how_to_release_copr.rst
nikromen: hmmm thinking of it again, I'll rather remove the comment here and keep the note

doc/how_to_release_copr.rst
```diff
@@ -107,11 +132,52 @@ bugzillas to ON_QA.
...
+Run::
+
+    tito report --untagged-commits
+
+and walk the directories of packages listed, and tag them. During development,
+we sometimes put a `.PATCH` number into our packages versions. See what packages has
+such version::
+
+    find . -maxdepth 2 -name "*.spec" -exec grep -H "Version:" {} + | grep -E '[0-9]+\.[0-9]+\.[0-9]+'
```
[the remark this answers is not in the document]
nikromen: this is literally applied ur suggestion from last review -_-, no

doc/how_to_release_copr.rst
```diff
@@ -97,6 +102,26 @@ instances::
...
+   hash appears in the RPM Release field::
+
+       # On the target machine:
+       ./releng/install-copr-packages @copr/copr-dev <commit-sha> copr-frontend copr-backend
+
+
+Test on dev machines
+....................
+
+Run the :ref:`sanity tests<sanity_tests>` from a Podman container (alternatively
```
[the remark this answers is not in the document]
nikromen: the space is valid typo

nikromen:
thank you @FrostyX for the review, all valid points. PTAL

praiskup (approved):

praiskup:
If something needs updates, let's follow up later.

PR #4310, opened 2026-05-13
packit: add stable status name for testing-farm job

### LINKED ISSUE #4287: Can we have a reminder that /packit test is needed?
It would be very inefficient to run Testing Farm tests for every single push or force-push, as a single run takes 1–2 hours. To avoid this waste, maintainers are required to explicitly trigger them via a `/packit test` comment.

It would be helpful to have a bot monitoring pull requests and remind us if a Testing Farm run is missing after the latest update (e.g., if we git push --force but forget to run `/packit test`, the bot would flag the PR with a `missing-ci-run` label).

nikromen:
This allows using GitHub required status checks to block merging PRs that haven't run /packit test, without needing to update the check name every time the target changes.

See https://packit.dev/docs/configuration#status_name_template

Fix #4287

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -200,6 +200,7 @@ jobs:
       - copr-frontend
     skip_build: true
     manual_trigger: true
+    status_name_template: "packit/testing-farm/gating/{package}"
 
   # Main commit builds:
   #


praiskup (approved):

.packit.yaml
```diff
@@ -200,6 +200,7 @@ jobs:
       - copr-frontend
     skip_build: true
     manual_trigger: true
+    status_name_template: "packit/testing-farm/{package}"
```
mfocko: I'd be more explicit with the intent here and probably include _gating_ (or something like that) in the check name…

Also, given that you want to use it as a merging / gating requirement, it might be more comfortable to trigger it via [label](https://packit.dev/docs/configuration#label) rather than the manual comment; though I'm not really sure now whether it gets triggered by the label itself 👀

praiskup: If triggered by label, how would we re-trigger?

praiskup:
Not merging until we sort out @mfocko's point.

.packit.yaml
mfocko: The label requirement (either for label being set or *not set*) works as a “precondition”, i.e., Packit reacts to events as usual (push, manual trigger via comment, reopening), but the job is skipped (also no status set IIRC) if you miss the required label, or you have the “forbidden” label.

praiskup: Could you please clarify your proposal? The goal of requiring an explicit `/packit test` is to avoid wasting ~2 hours of computing power unnecessarily. This overhead will soon double with the addition of Pulp testing; running it on every PR update or force push simply isn't sustainable. However, we still want to run these tests eventually for most PRs (ideally once).

Are you suggesting a `no-testing-farm-run-needed` label? We already have `no-ci-build` (which skips package builds), so this would be a similar addition.

mfocko: I suggested it as a more convenient way since you want to use it as a merge requirement, i.e., you **have to** run it at least once before the merge.

Truth be told, in any way I think about this, there's always some nuisance:
- if you require the label, you'd still need to comment the first time after adding the label (at least from the brief look at what events we process from GitHub), or rebase at least, and then subsequent updates of the PR will be tested accordingly
- if you use just the manual trigger, you'd have to comment at least once to satisfy the branch protection (it's also possible to require checks to be assigned to the latest changes, so comment per each rebase / small change just before the merge)

Now that you mention the `no-ci-build` label, I don't see it being used for disabling Packit 👀 is that intentional?

nikromen: I admire how fast you appeared here lol :D

I thought about the label approach, but I think it adds complexity without a clear benefit in our case.

With `manual_trigger: true` + required status check, the workflow is already straightforward:
  - PR is opened, the check stays pending (merge blocked)
  - When ready, someone comments /packit test
  - Tests pass, check goes green, PR can be merged

A label would add an extra step (add label + still need an event to trigger) without making this simpler.

nikromen:
Note for setting up the required status check in GitHub branch protection: the check name (packit/testing-farm/gating/copr-frontend) won't appear in the autocomplete until it has been reported at least once.

So after merging this PR, I need to run /packit test on any open PR first. Only then will the check name show up in Settings -> Branches -> "Require status checks to pass before merging" and can be added as a required check.

.packit.yaml
praiskup: I agree, this approach seems good enough to me.  @mfocko take this as an experiment, and feel free to open a PR if you have a better approach in mind.

praiskup: > Now that you mention the no-ci-build label, I don't see it being used for disabling Packit 👀 is that intentional?

Forgot to reply here;  we don't use Packit for RPM builds yet (waiting for a proper monorepo support).

PR #4316, opened 2026-05-13
docker: use Pulp as default storage

FrostyX:
At this point all new Copr projects use Pulp, so it's best to develop with that in mind.


praiskup (approved):

praiskup:
We just need to think about this when configuring Red Hat Copr.

PR #4320, opened 2026-05-18
Change the URL to work with the new API

jpodivin:
New API released with https://github.com/fedora-copr/logdetective-website/releases/tag/v0.32.0 allows submission of all logs from the build. Copr should use it instead of the legacy `url` endpoint, which allowed only a single log file.

Until this change is merged and released, the original endpoint will remain operational.


FrostyX:
There is something wrong with the CI that we need to address but the PR looks good to me

praiskup (approved):
Thank you!

PR #4336, opened 2026-06-01
Revert "backend: don't try to block @copr/PyPI and @copr/PyPI3 actions"

nikromen:
This reverts commit 8addefab8d5b7008773879a18c55d4ec9c28cbf7.


nikromen:
this should be merged and hotfixed after migration of copr pypi is done

FrostyX:
The original was two commits because I messed up the first one and already created a hotfix from it, so I needed a followup. But I'd prefer to squash both of the revert commits into one.

The migration is finished now, so once you squash, I'll merge and hotfix it.

PR #4347, opened 2026-06-11
openshift: convert Jinja2 templates to plain OpenShift templates

FrostyX:
Convert all OpenShift deployment templates from Jinja2/Ansible format to plain YAML templates with parameter substitution. This is needed for our patner repos deployment.

### DIFF
--- a/openshift/.gitignore
+++ b/openshift/.gitignore
@@ -1,1 +1,3 @@
 secret*
+params.env
+*.b64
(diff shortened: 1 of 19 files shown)


FrostyX:
Generated with Claude, we definitely need to clean this up before merging

- It prepared a migration process for moving production instances from the jinja2 templates to the OpenShift templates. We don't need that, we AFAIK don't have production instances yet
- The PR is `+1,721 -0` so it clearly kept all the previous templates and everything. We need to drop them

praiskup:
Obsoleted by #4404

PR #4348, opened 2026-06-11
Run Copr in OpenShift (locally for now)

nikromen:
Kustomize overlay + targets for deploying Copr on a local CRC MicroShift cluster. Builders are provisioned as Kubernetes pods via resalloc to simplify it for now (k8s-new/k8s-delete).

Beaker tests: most tests passes, 9 failures. All failures are environmental (missing chroots, single arch, hardcoded testing environment for podman-kube not for openshift). I'll fix the testing environment later to work with both (mainly bad dnf copr instance issues). After that I'll also provide a short TODO how to run the tests since I still did some manual tweaking on my machine to make the some tests run.

### DIFF
--- a/kube-deploy/README.md
+++ b/kube-deploy/README.md
@@ -63,14 +63,45 @@ and pipes the result to `podman kube play`.
 
 Run `just` to see all available commands.
 
-## Future: OpenShift Deployment
+## Local OpenShift (CRC MicroShift)
 
-This setup is designed as the foundation for OpenShift deployment:
+Run Copr on a local OpenShift cluster using CRC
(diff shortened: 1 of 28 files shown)


FrostyX (approved):
This is only `kube-deploy` change, so I am fine with merging once @nikromen is happy with the change.

I did not review though.

praiskup:
not a draft, merging

nikromen:
whoops I forgot to make it a draft since I wanted to post update to this with all of the beaker tests fixed, my bad :D nvm I'll do follow-up PR later

PR #4376, opened 2026-07-06
Fix create_file_logger guard for pytest >= 9.1 compatibility

hrnciar:
pytest 9.1 injects its own capture handler into non-propagating loggers, causing the `if not logger.handlers:` guard to skip creating the WatchedFileHandler on subsequent calls with different file paths. Check specifically for WatchedFileHandler instead.

Assisted-by: Claude Opus 4.6


hrnciar:
Here is the successful build with the latest pytest - https://copr.fedorainfracloud.org/coprs/thrnciar/pytest/build/10685415/

praiskup (approved):

FrostyX:
Thank you very much @hrnciar

PR #4397, opened 2026-07-20
Port markdown rendering to markdown-it-py

mikelolasagasti:
commonmark is deprecated upstream, and markdown-it-py is the recommended replacement for CommonMark-compatible rendering in Python.

Use MarkdownIt with raw HTML disabled so user-provided HTML is escaped instead of rendered, while preserving Pygments highlighting for markdown code blocks.

More info:

- https://github.com/readthedocs/commonmark.py/commit/df7dcb7eae8b092a83654e791244b9fcaf11a871
- https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/MRHYYJIJ3AW2KP52CZNBVIT2LYLPZDPP/


mikelolasagasti:
@coderabbitai resume

frontend/coprs_frontend/tests/test_views/test_coprs_ns/test_coprs_general.py
```diff
@@ -100,7 +100,9 @@ def test_customize_user_profile_description(self):
 
         show = self.test_client.get("/coprs/{0}/".format(self.u1.name))
         page = show.data.decode("utf-8")
+        assert '<div class="copr-markdown">' in page
         assert "bio" in page
+        assert "&lt;blink&gt;bio&lt;/blink&gt;" in page
```
praiskup, line 105: are the escape sequences really expected here?  Seems this is not going to render well

frontend/coprs_frontend/tests/test_views/test_coprs_ns/test_coprs_general.py
```diff
@@ -366,11 +368,11 @@ def test_copr_detail_allows_submitter_to_cancel_build(
 
     def test_codeblock_html_in_project_description(self, f_users, f_coprs):
         r = self.tc.get("/coprs/{0}/{1}/".format(self.u1.name, self.c1.name))
-        lines = ['<pre><code class="language-python"><div class="highlight"><span></span><span class="c1"># code snippet</span>',
+        lines = ['<pre><code class="language-python"><span class="c1"># code snippet</span>',
                  '    <span class="n">bar</span><span class="p">()</span>',
                  '    <span class="k">return</span> <span class="mi">1</span>',
-                 '</div>',
                  '</code></pre>']
+        escaped_code = ['&lt;blink&gt;Some HTML code &lt;/blink&gt;']
```
praiskup, line 375: Interesting :-) previously it did:

Image

frontend/coprs_frontend/tests/test_views/test_coprs_ns/test_coprs_general.py
mikelolasagasti: Yes, this is expected. With markdown-it-py and `html=False`, raw HTML is escaped rather than rendered or stripped. So `<blink>bio</blink>` is displayed literally as text. That is intentional here because these fields are markdown-only and should not accept arbitrary HTML.

praiskup: Ok, the previous behavior was to drop html tags, now it is going to be visible.  Can we keep the old behavior somehow?

praiskup:
Thank you for working on this, @mikelolasagasti

frontend/coprs_frontend/tests/test_views/test_coprs_ns/test_coprs_general.py
mikelolasagasti: We can preserve the old behavior if preferred, but I’d lean toward keeping the escaped output. These fields are markdown-only and the UI says inline HTML is forbidden. Showing `<blink>bio</blink>` literally makes that clear to the user and avoids silently changing their input by removing tags. Since it is escaped, it is still safe and will not render as HTML. WDYT?

praiskup: Ok, I don't disagree ... just document (in commit message?) how do we opt-in for the old behavior (if we wanted to switch in the future).

mikelolasagasti: Extended the commit message and added a small comment in the the code https://github.com/fedora-copr/copr/pull/4397/changes#diff-5cb72d57c472e2262c7dcec57b3ff14672354c815b259096d334f043935e26c8R59

praiskup (approved):

FrostyX:
I am looking at the failed Testing Farm CI and all of our tests passed. It then fails on `internal/timeout (after-test)`. Seems like a problem outside our codebase, so I think we don't have to block this PR because of it.

PR #4404, opened 2026-07-26
openshift: convert jinja2 templates to a Helm chart

FrostyX:
Obsoletes PR #4347


praiskup:
+1 for moving to Helm.

praiskup:
This is pretty hard to review without trying.  I'm giving +1, but would love to have a ~10 call with "how do you run it" so I can test myself.

praiskup (approved):

FrostyX:
> how do you run it

I preserved the original approach

```
cp helm/values.example.yaml helm/secrets.yaml  # and fill in the gaps
make
```

Edit: And this is the Communishift project I am testing it with https://forge.fedoraproject.org/infra/tickets/issues/13409

praiskup:
This PR is waiting for @nikromen as he has an ongoing review in hands.

praiskup:
@nikromen ping

nikromen:
sorry for the delay, I'll post the review tommorrow morning

openshift/Makefile
```diff
@@ -1,4 +1,17 @@
-AP := ansible-playbook -vv -c local -i localhost,
+AP := ansible-playbook -e ansible_python_interpreter=/usr/bin/python3.14 -vv -c local -i localhost,
```
nikromen: why we need to pin python interpreter like this? Also I think that in the future we could drop (most of) the ansible script since helm command could handle those

openshift/helm/templates/backend.yaml
```diff
@@ -51,7 +51,7 @@ metadata:
   name: backend-home
 type: Opaque
 data:
-  id_rsa: "{{ aws_config.ssh_key.private | b64encode }}"
+  id_rsa: {{ .Values.aws_config.ssh_key.private | toString | b64enc | quote }}
```
nikromen: The `data` expects base64 yes; we could instead use `stringData` to drop all of those `| toString | b64enc | quote` and instead use `| quote` which would simplify this:

```suggestion
stringData:
  id_rsa: {{ .Values.aws_config.ssh_key.private | quote }}
```

this apply to all of `| toString | b64enc | quote`

nikromen (commented):
I really like that we can do this in helm +1 :)

I think that the comments from coderabbit are worth of fixing (those I 👍 ) + the `| toString | base64 | quote` seems like quick fix.

---

Below I am putting comments that we should do in the future, but to not block this PR it could be done as follow ups (just putting it here in comment so it can be tracable). Also with mentioning ID of aceptance criteria it solves.

- secrets baked into config text vs injected as env vars. The config files are rendered as `Secret` object, this means the entire configs now inherits restrictions of real secrets. Also if we will plan to run this chart for different instances (also an internal AC-compliant one) with different config values, splitting config into ConfigMap + secrets only as env vars makes cross-instance debugging way easier (read configs without needing secret access) [SECURITY-0001 aligment].

- image registry is ahrdcoded per service in `values.yaml`... Parametrising it would make bumping versions and App-Interface saas_file image promotion way less painful [RELIABILITY-0001, RELEASING-0001, SECURITY-0002]

- replicas: 1 hardcoded everywhere [RELIABILITY-0001, RELEASING-0003]

openshift/helm/templates/backend.yaml
```diff
@@ -106,19 +106,22 @@ spec:
 
 ---
 kind: DeploymentConfig
```
nikromen, line 115: also one more architectural thingy worth raising: these charts still use openshift's `DeploymentConfig` instead of plain `Deployment`. The first one is legacy at this point, and AppSRE docs heavily recommends avoiding it/migrating. This will cost us portability and buys nothing us here (do not know what exactly they have in mind yet, just posting their words... (also implied in RELEASING-0001) something with bypassing helm/gitops workflows which AppSRE repo is built upon... only thing I know certainly that this is not core kubernetes, thus I would need to migrate this anyway to be able to run these inside podman via `podman kube play` (and reason why you need `--disable-openapi-validation`)).

the change itself shouldn't be complicated but not blocking this PR, just mentioning

openshift/Makefile
FrostyX: Aaah, that was a local fix that I needed because I have some mess in my system, and I accidentally committed it. Thanks for catching that.

openshift/helm/templates/backend.yaml
FrostyX: I agree, we should have some follow-up PR to convert them into `Deployment`. Hopefully this should be a straightforward task with AI.

FrostyX:
Updated

FrostyX:
@nikromen Updated again. There was one instance of `data` vs `stringData` that I left as is because it wasn't as clear-cut as the rest.

nikromen (approved):
thanks!

PR #4408, opened 2026-07-28
frontend,backend: add direct RPM upload build type

nikromen:
Allow publishing one or more already-built .rpm files directly into a single chroot. Frontend exposes a new endpoint for this.

~600 LoC are tests coverage

Obsoletes #4383

AI generated sequence diagram of this feature:

```mermaid
sequenceDiagram
    actor User
    participant FE as Frontend
    participant B as Builder

    User->>FE: upload RPM files plus chroot
    FE->>FE: save RPMs to tmp storage, create Build as succeeded and BuildChroot as pending
    Note over FE: skips SRPM build and dist-git import
    FE-->>User: 200 OK

    Note over FE,B: normal builder dispatch, unchanged

    B->>FE: GET get-build-task, returns prebuilt_rpm_urls
    B->>B: build_rpm sees prebuilt_rpm_urls, calls build_rpm_upload
    loop each uploaded RPM
        B->>FE: download_file
        B->>B: get_rpm_header checks arch vs chroot
    end
    B->>B: write success marker, run_automation_tools generates results.json

    Note over FE,B: normal result collection, sign, createrepo, build_chroot_result, messaging, unchanged

```


nikromen:
image

backend/copr_backend/actions.py
```diff
@@ -19,11 +19,10 @@
 from copr_backend.worker_manager import BackendQueueTask
 from copr_backend.storage import storage_for_enum, BackendStorage, PulpStorage
 
-from .sign import create_user_keys, CoprKeygenRequestError
+from .sign import create_user_keys, get_pubkey, CoprKeygenRequestError
```
nikromen: this change was needed to make CI happy

praiskup: ?? weird, considering this file isn't otherwise changed?

common/copr_common/enums.py
```diff
@@ -118,6 +118,7 @@ class BuildSourceEnum(metaclass=EnumType):
             "scm": 8, # type, clone_url, committish, subdirectory, spec, srpm_build_method
             "custom": 9, # user-provided script to build sources
             "distgit": 10, # distgit_instance, package_name, committish
+            "rpm_upload": 11, # tmp, files -- pre-built RPMs uploaded directly, no SRPM build
```
praiskup, line 121: bump common version, and bump corresponding foo->common requirements

nikromen:
seems like some hiccup in testing farm

/packit test

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -396,7 +396,15 @@ def get_pending_build_tasks(cls, background=None, data_type=None):
...
-            query = query.join(models.Package, models.Package.id == models.Build.package_id)
+            #
+            # This must be an OUTER join: builds with no associated Package
+            # (e.g. ad-hoc "upload"/"url" builds, and every "direct RPM
+            # upload" build, which is package-less by design) would
+            # otherwise be silently excluded from /status/pending/ and the
+            # sidebar queue-size counters by an inner join on a NULL
+            # Build.package_id, even though they are very much real,
+            # in-flight tasks that backend is actively processing.
+            query = query.outerjoin(models.Package, models.Package.id == models.Build.package_id)
```
praiskup: OTOH, having a package name (and allocating Package) should be fairly trivial and safe from the package file name?  I mean, this query has been optimized for years...  if there's no other option, ACK (but would be very great if we could avoid touching this beast).

praiskup (commented):
submitting review

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +739,110 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
+            copr, "Can't build while there is an operation in progress: {action}")
+        users_logic.UsersLogic.raise_if_cant_build_in_copr(
+            user, copr, "You don't have permissions to build in this copr.")
+
+        chroot = cls._find_active_chroot(copr, chroot_name)
+        tmp_name, filenames = cls._save_uploaded_rpms(form_files)
+
+        try:
+            source_json = json.dumps({"tmp": tmp_name, "files": filenames})
+            batch = cls.setup_batch(after_build_id, with_build_id, user)
```
praiskup, line 832: +1, nice to see we support batches

frontend/coprs_frontend/coprs/views/apiv3_ns/schema/schemas.py
```diff
@@ -730,6 +723,17 @@ class _GenericBuildOptions:
         description="Build in the same batch with the Build ID build",
         example=123,
     )
+
+
+@dataclass
+class _GenericBuildOptions(_BatchBuildOptions):
```
praiskup: FWIW, it's pretty weird to see this inheritance (I'd expect vice versa)

frontend/coprs_frontend/coprs/forms.py
```diff
@@ -1440,6 +1459,20 @@ def __new__(cls, active_chroots):
...
+    Build form for the direct RPM upload build source.
+    """
+    def __new__(cls, active_chroots):
+        form = _get_build_form(active_chroots, BaseForm)
+        form.chroot = wtforms.StringField('chroot', validators=[
+            wtforms.validators.DataRequired()])
+        form.pkgs = MultipleFileField('rpms', validators=[
+            FileRequired(),
+            RpmValidator()])
+        return form
```
praiskup, line 1505: Can't we reuse the SRPM upload factory?

frontend/coprs_frontend/coprs/request.py
```diff
@@ -71,17 +71,25 @@ def _get_file_stream(
...
-    with _CoprRequestClass above is in action, and when you are sure that the
-    uploaded file is on the same partition with the target path (so we can
-    hardlink).
+    Store an uploaded werkzeug.datastructures.FileStorage to `path`.  Use
+    this if the hack with _CoprRequestClass above is in action, and when you
+    are sure that the uploaded file is on the same partition with the target
+    path (so we can hardlink).
     """
-    stream = field.data.stream
+    stream = file_storage.stream
```
praiskup, line 83: Can you name the reasons for this change in commit message?

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: I discovered this bug accidentally when uploading the RPM, then in the UI (the /status/pending) was empty and suddenly the build was pushed out of it elsewhere, but I was never able to see it in pendings to begin with.

This same bug occurs when package parameter in build is None, which results in e.g.: if user does something like `copr build --chroot fedora-40-x86_64 something.src.rpm`, BuildChroot is created immediatelly -> package id is null and is null during SRPM import until backend sends pkg_name. In some part of this, the BuildChroot should be in pending state, yet since package id is still null, with the inner join it is never visible in /status/pending

---

also on different note -  due to direct RPM build being package-less, the forking feature is borked... I need to think of this as well (either we accept forking silently skips these for now and document it or somehow implement it)

Re trivial and safe: i don;t think it's actually trivial (I may be wrong again as well as with the skipping dist-git import being complicated); allocating a package means find-or-create by name (unclear what happens with multiple differently-named RPMs in one call to me), reworking a few spots in models.py that rely on package being None to know theres no dist-git repo behind the build, and the worst the `PackagesLogic.build_package()` bypasses `create_new_from_other_build` entirely... and would resubmit using an already-deleted tmp dir

frontend/coprs_frontend/coprs/views/apiv3_ns/schema/schemas.py
nikromen: it's how the schemas were created so I went with it, but I admit this looks like unnecesary inheritance

frontend/coprs_frontend/coprs/forms.py
nikromen: it is already reused - both uses get_build_from; just pkgs is different - user can upload multiple RPMs to one chroot. Also SRPM validator allows src rpm file which would be wrong here so those are different

the BuildFormRpmUploadFactory also currently forbids to upload to more than 1 chroot right now

frontend/coprs_frontend/coprs/request.py
nikromen: there is no behavior change, just refactor for multiple use-case usages. It was needed so `save_uploaded_rpms()` can reuse the same hardlink optimisation for raw file storage, so that's why for this I created separate method reusing it

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +739,110 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
+                submitted_on=int(time.time()),
+                is_background=bool(background),
+                batch=batch,
+                timeout=timeout or app.config["DEFAULT_BUILD_TIMEOUT"],
+            )
+            db.session.add(build)
+
+            cls.assign_buildchroots(build, [chroot], status=StatusEnum("pending"))
+        except Exception:
+            # TODO: some cleanup procedure instead of broad exception catch
```
nikromen, line 861: any ideas on this? do you think this is enough or cleanup procedure is needed?

praiskup: I usually prefer `try: sth except foo: handle finally: cleanup`

praiskup: The exception is more likely to come after commit() which is in different method?

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: we might actually have something already in our code, but this is basically what I meant...
```
VALID_NAME_REGEX = re.compile(r'^[a-zA-Z0-9_\+\-\.]+$')

def get_package_name(url_or_path: str) -> str | None:
    # 1. Odstraní leading path / URL (získá jen název souboru)
    filename = Path(urlparse(url_or_path).path).name
    
    # 2. Odřízne 2 nejpravější pomlčky a vše za nimi
    # Pro "python3-devel-3.12.2-1.fc40.x86_64.rpm" vrstvy:
    # rsplit('-', 2) -> ['python3-devel', '3.12.2', '1.fc40.x86_64.rpm']
    parts = filename.rsplit('-', 2)
    
    if len(parts) < 3:
        return None  # Vřetězcích chybí pomlčky (neodpovídá struktuře name-ver-rel)
        
    pkg_name = parts[0]
    
    # 3. Prověří, že zbytek je validní package name
    if VALID_NAME_REGEX.match(pkg_name):
        return pkg_name
        
    return None
```

The worst-case scenario is either a) a build failure or b) an incorrect package name in the DB (which isn't a major issue, as we used to deal with that all the time when building SCLs).

Having no Package assigned to a build is a bad idea anyway, as users expect features like listing all builds for package X...

praiskup: We have to disable "resubmit" feature, sure.

frontend/coprs_frontend/coprs/forms.py
praiskup: > the BuildFormRpmUploadFactory also currently forbids to upload to more than 1 chroot right now

What are pros/cons of doing this restriction?

frontend/coprs_frontend/coprs/request.py
praiskup: I was cofused, sorry... if you can make smaller changes, please do.  Unnecessary changes like this particular method rename & parameter rename just distracts reviewer.

backend/copr_backend/actions.py
praiskup: Could you please avoid changing this file, once you are on this PR and rebasing or so?

frontend/coprs_frontend/coprs/views/apiv3_ns/schema/schemas.py
praiskup: Well if it was _Batch => _Generic, then :+1: but this is _Generic => _Batch, that seems weird :shrug:

praiskup: (a new thing in this PR)

frontend/coprs_frontend/coprs/forms.py
nikromen: This was agreed upon while defining the task in Jira - as far as I remember this was due to not allowing user to upload same RPM e.g. for fedora to opensuse chroots etc..

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: the try-finally does not quite fit here since we only want to cleanup on failure

you are right on the commit comment but this approach is copied from the `create_new_from_upload`:
```
        try:
            build = cls.create_new(user, copr, source_type, source_json,
                                   chroot_names, pkgs=pkg_url, srpm_url=srpm_url,
                                   copr_dirname=copr_dirname, **build_options)
        except Exception:
            shutil.rmtree(tmp)  # todo: maybe we should delete in some cleanup procedure?
            raise
```

so it should probably work as is. Maybe we could solve this as follow-up? I suspect this would mean moving the commit somewhere the cleanuyp can actually see it fail and the PR is already big change enough...

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: with that what if the user uploads foo.rpm foo-devel.rpm foo-bar.rpm and baz.rpm - we don't know the actual package name... to simplify this I decided to make this feature to upload only 1 RPM per call, so we have 1:1 package and RPM... we can allow multiple RPMs as follow ups in the future to not complicate things right now

nikromen: re the build_package/batch_build bypass I raised earlier...  we still need to skip those explicitly (regardless the package naming issue)

nikromen: this diff should solve issues with packages and project forking - https://github.com/fedora-copr/copr/compare/1e8c33caac6001481988256df810c459538ffcdd..0c3e29434a9b4afd8de29ef00e23ad41bcd66d35

nikromen: confirmed forking works in local copr instance

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: > this approach is copied from

Uh, not sure I like that code :-) because we must be doing rmtree() on multiple places

frontend/coprs_frontend/coprs/forms.py
praiskup: Not sure we eventually made a decision?  I was thinking of it, but ... do we want to protect the user against shooting themselves into the foot here?

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: > what if the user uploads foo.rpm

(Build) failure?  User uploads nonsense (package file name, coming from rpmbuild, always has `<name>-<version>-<release>.suffix` :shrug:).

> simplify this I decided to make this feature to upload only 1 RPM per call

I'm personally against switching from join to outerjoin() for not super great reasons, but I'm prepared to be overruled.

Task management is a can of worms, and you never know what weird combinations of "builds without packages" might suddenly show up—or accidentally end up—in the task queue.

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: (rmtree _should_ live in finally)

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: oh this was my bad example sorry. What I mean is that if user will upload multiple RPMs with parsing the package name would result in different names, e.g.: foo, foo-devel, foo-bar, baz. Let's assume that the rest of NEVRA is correct.

> I'm personally against switching from join to outerjoin()

no problem, I initially put this into the PR because this feature was package-less. Now since it has package included, it could be deleted. However the bug I am mentioning still persist, so I'll at least put it into comment/create follow-up issue. But it was pre-existing problem that is not exclusive to this PR (I also confirmed this problem only in local instance, not in staging instance as well...)

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: what approach then do you recommend instead?

frontend/coprs_frontend/coprs/forms.py
nikromen: ah I thought we made a decision :D but I'd rather keep it minimal as it is since I already received request on putting metadata (build logs, virus scans etc) to include it in the future. We could figure out later how to map these metadata to the RPMs once we got more clear use cases

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: added comment and reverted change

praiskup:
Uh, conflict -- can you please rebase?

common/python-copr-common.spec
```diff
@@ -1,7 +1,7 @@
 %global srcname copr-common
 
 Name:       python-copr-common
-Version:    1.7
+Version:    1.7.1
```
praiskup: We'll need to bump to 1.7.3 now

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -396,6 +396,11 @@ def get_pending_build_tasks(cls, background=None, data_type=None):
             # that this is racy -- Package reference provides some build
             # configuration which can be changed in the middle of the
             # BuildChroot processing.
+            #
+            # Builds with NULL package_id (e.g. "upload"/"url" source
+            # types) are silently missing from /status/pending/ and the
+            # sidebar queue-size counters. Fix could be outerjoin, but
+            # could cause other issues.
```
praiskup, line 403: This is (almost for sure) expected, hence the discussion here in the PR.

We are generally searching for "tasks to be done";  BuildChroots without assigned package entry are not (yet) ready to be processed by the backend (considering this method is used to feed the backend's priority queue).

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +736,143 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
         return build
 
+    @classmethod
+    def _find_active_chroot(cls, copr, chroot_name):
+        for active_chroot in copr.active_chroots:
+            if active_chroot.name == chroot_name:
+                return active_chroot
+        raise BadRequest(
+            f"Chroot {chroot_name} does not exist or is not active "
+            f"in the project {copr.full_name}")
```
praiskup: weird, normally we should get
`Error: Bad request parameters: {'chroots': ["'rhel-10-x86_64' is not a valid choice for this field."]}`

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +736,143 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
+                is_background=bool(background),
+                batch=batch,
+                timeout=timeout or app.config["DEFAULT_BUILD_TIMEOUT"],
+            )
+            db.session.add(build)
+
+            cls.assign_buildchroots(build, [chroot], status=StatusEnum("pending"))
+        except Exception:
+            # TODO: some cleanup procedure instead of broad exception catch
+            shutil.rmtree(os.path.join(app.config["STORAGE_DIR"], tmp_name))
```
praiskup, line 862: I'd prefer using `finally` block.  What is going to happen with the directory when there's no exception?

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
```diff
@@ -166,7 +172,8 @@ def get_build_record(task, for_backend=False):
         "enable_net": task.build.enable_net,
         "git_repo": task.distgit_clone_url,
         "git_hash": task.git_hash,
-        "package_name": task.build.package.name,
+        # "direct RPM upload" builds (see below) have no associated Package
+        "package_name": task.build.package.name if task.build.package else None,
```
praiskup: It's not convenient to approve this.  I'd wish we had the build.package set earlier...

rpmbuild/copr_rpmbuild/helpers.py
```diff
@@ -118,6 +120,44 @@ def get_rpm_header(path):
...
+    except RequestError as ex:
+        raise RuntimeError(f"Failed to download {url}: {ex}") from ex
+
+    # %-decode the URL path *before* taking its basename -- otherwise an
+    # encoded separator (e.g. "..%2f..%2fetc%2fpwn") would only be decoded
+    # into "../../etc/pwn" after basename() already treated it as a single
+    # (safe-looking) path component, allowing it to escape `destination`
+    # once joined.  The query string (e.g. "?dl=1") is dropped by urlparse().
+    filename = os.path.basename(unquote(urlparse(url).path))
+    if not filename or filename in (".", ".."):
```
praiskup: this is very weird check, can we document why we are doing it?

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: on success the tmp dir is left alone on purpose since the builder still needs to fetch the RPM later, so we can't delete it here. It gets deleted once the chroot reaches its end later. The except here covers the part where something fails before build chroot is even committed

frontend/coprs_frontend/coprs/views/backend_ns/backend_general.py
nikromen: good catch, this is leftover from the feature being package-less

edit: seems there were more leftovers

rpmbuild/copr_rpmbuild/helpers.py
nikromen: protection against sus filenames like "" or ".." which then could mean tinkering with parent directory

nikromen: in fact it would be on one directory level, on throwaway builder anyway (so it does not matter that much if user would decide to shoot himself..) but better safe than sorry

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: fixed

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +736,143 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
+                "Only one RPM can be uploaded per call for now "
+                "(multi-RPM upload may be added later).")
+
+        sanitized_names = []
+        for form_file in form_files:
+            sanitized = secure_filename(form_file.filename)
+            if not sanitized or not sanitized.endswith(".rpm"):
+                raise BadRequest(
+                    f"Uploaded filename '{form_file.filename}' is invalid "
+                    "or could not be safely sanitized to a valid .rpm filename.")
```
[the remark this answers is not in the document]
nikromen: this is protected by form but OK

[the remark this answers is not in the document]
nikromen: in fact it is not protected against nosrcrpm (whatever that is) so good catch

frontend/coprs_frontend/coprs/logic/builds_logic.py
nikromen: the bug here is "purely cosmetic": e.g. in the window between builchroot creation and the backend reporting back pkg_name from the dist git import are invisible in pending tasks. The build finishes just fine, it is just never showed in pending

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: aha, thank you for the explanation.

rpmbuild/copr_rpmbuild/helpers.py
praiskup: If you want to validate file basenames, please also check for `\0`, `\\`.  And ideally, do these checks before actually downloading the file.

The check just for `.` and `..` is just very weird.  You should probably check for an empty string, too.

Then, please think about a trailing slash, like `example.com/foo.tar.gz/`.

frontend/coprs_frontend/coprs/logic/builds_logic.py
praiskup: I meant that we shouldn't even need this check?

frontend/coprs_frontend/coprs/logic/builds_logic.py
```diff
@@ -727,6 +736,139 @@ def create_new_from_upload(cls, user, copr, form_field, orig_filename,
...
+        :param str chroot_name: name of the chroot to publish into
+        :param form_files: list of uploaded-file objects
+        :return: models.Build
+        """
+        coprs_logic.CoprsLogic.raise_if_unfinished_blocking_action(
+            copr, "Can't build while there is an operation in progress: {action}")
+        users_logic.UsersLogic.raise_if_cant_build_in_copr(
+            user, copr, "You don't have permissions to build in this copr.")
+
+        chroot = cls._find_active_chroot(copr, chroot_name)
```
praiskup: I don't like this special handling of chroot options.  Can we please use the default mechanism (that allows multiple choices), and if needed (I don't think so) add additional restriction like
```
if max_build_chroot and len(chroot) > max_build_chroot:
    raise ValidationError("You can only select one chroot with this build method")

File: frontend/coprs_frontend/coprs/logic/builds_logic.py (hunk 10 above)
Comment (praiskup) [reply frontend/coprs_frontend/coprs/logic/builds_logic.py:403]: What I'm trying to say is that we don't want to pollute the queue with tasks that are not really buildable (they are not really "pending").

File: frontend/coprs_frontend/coprs/logic/builds_logic.py (hunk 8 above)
Comment (praiskup) [reply frontend/coprs_frontend/coprs/logic/builds_logic.py:861]: From the other thread I understand now, thank you for the patience with me 👼

File: frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py [hunk 17]
```diff
@@ -314,40 +314,48 @@ def create_new_build(options):
...
+    @apiv3_builds_ns.expect(create_build_rpm_upload_input_model)
+    @apiv3_builds_ns.marshal_with(build_model)
+    def post(self):
+        """
+        Create a build from RPM upload
+        Publish an already-built RPM directly for one or more chroots.
+        """
+        if not app.config["DIRECT_RPM_UPLOAD"]:
+            raise BadRequest(
+                "Direct RPM upload is not enabled on this Copr instance")
```
praiskup, line 335: 👍

praiskup (approved):
LGTM

PR #4410, opened 2026-07-29
cli: add --with-deps option to buildpypi

### LINKED ISSUE #4373: RFE: add `--with-deps` args to copr-cli to build a package with its full dependency tree as ordered batches
Relates to: https://github.com/fedora-copr/copr/issues/2639#issuecomment-2614015831

I have been working on a tool which computes the dependency graph for a package (all metadata is being fetched from https://ecosyste.ms/), prunes the nodes which are already available in the official fedora repositories and copr project, topologically sort them in order to build batches in parallel.

This tool is still at very early stages (it's functional but not fully perfect). The  `coprtree` library exposes only one method `resolve_dependencies`,  for a given package with specified chroot and copr project, it returns the deps sorted into levels that can be batch build in parallel, which should be easy to integrate with  `copr-cli`

More info can be found at:  https://github.com/sundaram123krishnan/coprtree

I would like to integrate this with `copr-cli` myself, if the team is open to it.

sundaram123krishnan:
Resolves: #4373
This PR introduces the integration of [coprtree](https://github.com/sundaram123krishnan/coprtree) library with `copr-cli`
As per @FrostyX's guidelines, the library is packaged in official fedora repositories and is loosely coupled with `copr-cli`.
Anyone who wants to use `--with-deps` should install it with:
```sh
sudo dnf install python3-coprtree
```
It supports `pypi` for now.
It's still in early stages, and I would like to add support for other package managers soon.
Feel free to open issues in the `coprtree` repo, if you hit any bugs in the dependency resolution. 
Quick demo:

https://github.com/user-attachments/assets/b0c712b4-513f-417c-9d49-12984b5d3105

Would like to hear the team's thought on the approach.

### DIFF
--- a/cli/copr_cli/helpers.py
+++ b/cli/copr_cli/helpers.py
@@ -9,6 +9,7 @@ information per line. The text-row format prints all information separated
 by a space on a single line.
 """
 
+
 def cli_use_output_format(parser, default='json'):
     """
     Add '--output-format' option to given parser.
@@ -18,6 +19,19 @@ def cli_use_output_format(parser, default='json'):
         choices=["text", "json", "text-row"],
         help=output_format_help,
         default=default,
+    )
+
+
+def cli_with_deps_option(parser):
+    """
+    Add '--with-deps' option to given parser.
+    """
+    parser.add_argument(
+        "--with-deps",
+        action="store_true",
+        help="Build the package dependencies for a given package using the "
+             "coprtree library. More information can be found at: "
+             "https://github.com/sundaram123krishnan/coprtree",
     )
 
 
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -6,13 +6,15 @@ import io
 import argparse
 import datetime
 import logging
+from copy import copy
 import os
 import re
 import subprocess
 import sys
 import time
+from types import SimpleNamespace
 import warnings
-from collections import defaultdict
+from collections import defaultdict, namedtuple
 from configparser import ConfigParser
 from rich.text import Text
 from rich.console import Console
@@ -44,6 +46,7 @@ from copr.v3.pagination import next_page
 import copr.v3.requests as copr_requests
 from copr_cli.helpers import (
     cli_use_output_format,
+    cli_with_deps_option,
     print_project_info,
     colorize_status,
 )
@@ -75,6 +78,8 @@ BOOTSTRAP_MAP = {
     "image": "image",
     None: "default",
 }
+
+SourceMethod = namedtuple("SourceMethod", ["provider", "build_function"])
 
 try:
     input = raw_input
@@ -439,6 +444,16 @@ class Commands(object):
 
         self.print_build_info_and_wait(builds, args)
 
+    def pypi_cb(self, package, **kwargs):
+        """
+        Callback function for building a PyPI package.
+        """
+        return self.client.build_proxy.create_from_pypi(
+            pypi_package_name=package.name,
+            pypi_package_version=package.version,
+            **kwargs)
+
+
     @requires_api_auth
     def action_build_pypi(self, args):
         """
@@ -453,6 +468,12 @@ class Commands(object):
             "spec_template": args.spec_template,
             "python_versions": args.pythonversions,
         }
+        if args.with_deps:
+            source_method = SourceMethod(
+                provider="pypi.org",
+                build_function=self.pypi_cb,
+            )
+            return self.process_build_with_deps(args, source_method)
         return self.process_build(args, self.client.build_proxy.create_from_pypi, data)
 
     @requires_api_auth
@@ -536,6 +557,91 @@ class Commands(object):
                                 project_dirname=project_dirname,
                                 buildopts=buildopts, **data)
         builds = result if type(result) == list else [result]
+        self.print_build_info_and_wait(builds, args)
+
+
+    def process_build_with_deps(self, args, source_method):
+        """
+        Resolve the dependency tree of the requested package, and submit each
+        level of the tree as a separate batch.
+        """
+        # pylint: disable=too-many-locals
+        try:
+            # We want this to be an optional feature and don't require every
+            # user to install the dependency. It may not even be available on
+            # all systems where we need the copr-cli to run.
+            # pylint: disable=import-outside-toplevel
+            from coprtree.coprtree import resolve_dependencies
+            from coprtree.exceptions import CoprtreeError
+            from coprtree.models import BuildEnv, BuildTarget
+        except ImportError as err:
+            raise CoprException(
+                "The --with-deps option requires the coprtree library to be "
+                "installed, install it using sudo dnf install python3-coprtree"
+            ) from err
+
+        if not args.chroots or len(args.chroots) > 1:
+            raise argparse.ArgumentTypeError(
+                "The --with-deps option requires exactly one --chroot for now"
+            )
+
+        username, projectname, project_dirname = self.parse_dirname(args.copr_repo)
+
+        target = BuildTarget(
+            provider=source_method.provider, # only pypi is supported for now
+            name=args.packagename,
+            version=args.packageversion,
+        )
+        env = BuildEnv(chroot=args.chroots[0],
+                       copr_project="{0}/{1}".format(username, projectname))
+
+        if not target.version:
+            print("No package version specified, building the latest version "
+                  "of {0}".format(target.name))
+
+        try:
+            levels = list(resolve_dependencies(target, env))
+        except CoprtreeError as err:
+            raise CoprException(
+                "Can not resolve the dependencies: {0}\n"
+                "Please report this at "
+                "https://github.com/sundaram123krishnan/coprtree/issues"
+                .format(err)) from err
+
+        builds = []
+        build_opts = buildopts_from_args(args)
+
+        batch = build_opts.pop("with_build_id", None)
+        previous_batch = build_opts.pop("after_build_id", None)
+        for number, level in enumerate(levels):
+            for index, package in enumerate(level):
+                actual_options = copy(build_opts)
+                if batch is not None:
+                    actual_options.update({"with_build_id": batch})
+                elif previous_batch is not None:
+                    actual_options.update({"after_build_id": previous_batch})
+
+                package_details = SimpleNamespace(
+                    name=package.name,
+                    version=package.version,
+                )
+                build = source_method.build_function(
+                    package=package_details,
+                    ownername=username,
+                    projectname=projectname,
+                    project_dirname=project_dirname,
+                    buildopts=actual_options,
+                )
+                if index == 0:
+                    print("level {0}:".format(number))
+                print("{0} {1} -> build {2}".format(
+                    package.name, package.version, build.id))
+                builds.append(build)
+                batch = batch or build.id
+            previous_batch = batch
+            batch = None
+
+
         self.print_build_info_and_wait(builds, args)
 
 
@@ -1601,6 +1707,7 @@ def setup_parser():
     parser_build_pypi = subparsers.add_parser("buildpypi", parents=[parser_pypi_args_parent, parser_build_parent],
                                               help="Build PyPI package to a specified copr")
     parser_build_pypi.set_defaults(func="action_build_pypi")
+    cli_with_deps_option(parser_build_pypi)
 
     # create the parser for the "buildgem" command
     parser_build_rubygems = subparsers.add_parser("buildgem", parents=[parser_rubygems_args_parent, parser_build_parent],
--- a/cli/copr-cli.spec
+++ b/cli/copr-cli.spec
@@ -31,6 +31,7 @@ Requires:      python3-rich
 
 Recommends:    python3-progress
 Recommends:    python3-ConfigUpdater
+Recommends:    python3-coprtree
 Suggests:      python3-beautifulsoup4
 
 BuildRequires: python3-copr >= %min_python_copr_version
--- a/cli/man/copr-cli.1.asciidoc
+++ b/cli/man/copr-cli.1.asciidoc
@@ -268,7 +268,7 @@ This limitation comes from the COPR API.
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 usage: copr buildpypi [-h] [-r, --chroot CHROOTS] [--memory MEMORY] [--timeout TIMEOUT] [--nowait]
-                      [--background]
+                      [--background] [--with-deps]
                       [--pythonversions [VERSION [VERSION ...]]] [--packageversion PYPIVERSION]
                       --packagename PYPINAME
                       project
@@ -281,6 +281,10 @@ Version of the PyPI package to be built (by default latest)
 
 --packagename PYPINAME::
 Name of the PyPI package to be built, required.
+
+--with-deps::
+Build also the dependencies of the package. The dependency tree is resolved by the coprtree library. Requires the python3-coprtree package to be installed.  Please report bugs in the dependency resolution to
+https://github.com/sundaram123krishnan/coprtree
 
 
 For the rest of the arguments, see `copr-cli build` command above.
--- a/cli/man/copr-cli.cheat
+++ b/cli/man/copr-cli.cheat
@@ -28,5 +28,8 @@ copr-cli cancel <build_id>
 # to build rpm(s) from GIT
 copr-cli buildscm test-project --clone-url https://github.com/fedora-copr/copr.git --subdir cli
 
+# to build a PyPI package together with its dependencies
+copr-cli buildpypi test-project --packagename pydantic-ai --with-deps --chroot fedora-rawhide-x86_64
+
 # to regenerate repository metadata for a project
 copr-cli regenerate-repos test-project


cli/copr_cli/helpers.py
```diff
@@ -21,6 +27,46 @@ def cli_use_output_format(parser, default='json'):
...
+            build = submit(package, options)
+            if not is_level_printed:
+                print("level {0}:".format(number))
+                is_level_printed = True
+            print("{0} {1} -> build {2}".format(
+                package.name, package.version, build.id))
+            builds.append(build)
+            batch = batch or build.id
+        previous_batch = batch
+    return builds
```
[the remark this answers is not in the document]
sundaram123krishnan: build levels cannot be empty

cli/copr_cli/main.py
```diff
@@ -539,6 +544,70 @@ def process_build(self, args, build_function, data):
...
+        )
+        env = BuildEnv(chroot=args.chroots[0],
+                       copr_project="{0}/{1}".format(username, projectname))
+
+        if not target.version:
+            print("No package version specified, building the latest version "
+                  "of {0}".format(target.name))
+
+        try:
+            levels = list(resolve_dependencies(target, env))
```
praiskup, line 603: Very interesting and cool.  Thank you.

Some deps might already be in Fedora, do you skip those?  I'm not suggesting you should change the code or something, but the behavior of this feature should be well documented. 🙏

cli/copr_cli/main.py
```diff
@@ -539,6 +544,70 @@ def process_build(self, args, build_function, data):
...
+            raise argparse.ArgumentTypeError(
+                "The --with-deps option requires exactly one --chroot for now"
+            )
+
+        username, projectname, project_dirname = self.parse_dirname(args.copr_repo)
+
+        target = BuildTarget(
+            provider="pypi.org", # only pypi is supported for now
+            name=data["pypi_package_name"],
+            version=data["pypi_package_version"],
```
praiskup: Can we abstract the BuildTarget out from this method?  Or is this going to be PyPI only forever?

cli/copr_cli/helpers.py
```diff
@@ -21,6 +27,46 @@ def cli_use_output_format(parser, default='json'):
...
 
 
+def cli_with_deps_option(parser):
+    """
+    Add '--with-deps' option to given parser.
+    """
+    parser.add_argument(
+        "--with-deps",
+        action="store_true",
+        help=with_deps_help,
```
FrostyX: A minor nitpick: I wouldn't define a separate variable and put the text right here.
I can see you did it the same way as we did `output_format_help` in the past, but I think that was a separate variable because we previously used it on multiple places.

cli/copr_cli/main.py
```diff
@@ -539,6 +544,70 @@ def process_build(self, args, build_function, data):
...
+        try:
+            # pylint: disable=import-outside-toplevel
+            from coprtree.coprtree import resolve_dependencies
+            from coprtree.exceptions import CoprtreeError
+            from coprtree.models import BuildEnv, BuildTarget
+        except ImportError as err:
+            raise CoprException(
+                "The --with-deps option requires the coprtree library to be installed, "
+                "install it using sudo dnf install python3-coprtree"
+            ) from err
```
FrostyX, line 581: This is exactly how I imagined it, thank you.
Could you please put a short explanation above the `# pylint: disable=` line so that is clear why we don't do the import on top? Something like

```python
# We want this to be an optional feature and don't require every
# user to install the dependency. It may not even be available on
# all systems where we need the copr-cli to run.
```

cli/copr_cli/helpers.py
```diff
@@ -21,6 +27,46 @@ def cli_use_output_format(parser, default='json'):
...
+def submit_build_batches(levels, submit):
+    """
+    Submit each level of packages as a separate build batch
+    """
+    builds = []
+    previous_batch = None
+    for number, level in enumerate(levels):
+        is_level_printed = False
+        batch = None
+        for package in level:
```
FrostyX: Nitpick: I think that if you `enumerate(level)`, you can then drop the `is_level_printed` variable because you can then do something like

```python
if i == 0:
    print("level {0}:".format(number))
```

cli/copr_cli/main.py
```diff
@@ -539,6 +544,70 @@ def process_build(self, args, build_function, data):
...
+        """
+        try:
+            # pylint: disable=import-outside-toplevel
+            from coprtree.coprtree import resolve_dependencies
+            from coprtree.exceptions import CoprtreeError
+            from coprtree.models import BuildEnv, BuildTarget
+        except ImportError as err:
+            raise CoprException(
+                "The --with-deps option requires the coprtree library to be installed, "
+                "install it using sudo dnf install python3-coprtree"
```
FrostyX: Nitpick: Can we please wrap at 80 characters?

cli/copr_cli/main.py
sundaram123krishnan: Yes, it would skip the deps that are present in both copr project and the specified chroot.
All it does is download the `repodata` index from the specified chroot and copr project, and dnf under the hood will check their availability and prune them accordingly.

cli/copr_cli/main.py
FrostyX: Agree with @praiskup here. I think the `BuildTarget` object should be created in `action_build_pypi` and passed into this function the same way that the `build_function` is. That way we can easily extend the support for other ecosystems.

sundaram123krishnan: Yes, initialy i had thought of creating an abstraction, then I had thought of doing it whenever I add support for `rubygems`, which i am planning to do it soon.
Is that okay? @praiskup @FrostyX

sundaram123krishnan: The library also supports `metacpan`, which in copr is a custom method, so the abstraction should cover that as well. But for this PR i only scoped it for `pypi`

cli/copr_cli/main.py
```diff
@@ -539,6 +544,70 @@ def process_build(self, args, build_function, data):
...
+            data["pypi_package_version"] = package.version
+            return build_function(
+                ownername=username,
+                projectname=projectname,
+                project_dirname=project_dirname,
+                buildopts=buildopts,
+                **data,
+            )
+
+        builds = submit_build_batches(levels, submit)
```
FrostyX: We are dangerously getting into the functional programming territory here :D

We are calling the `submit_build_batches` function with a `submit` function as a parameter, which is defined as an inner function of this function, and inside it is calling a `build_function` which is not passed into it but rather taken from the scope of this function, into which it is passed as a parameter from somewhere else ... I mean, definitely tickled my fancy but not sure if we should refactor this somehow.

FrostyX (commented):
Thank you very much for the PR. As I expressed multiple times, I really like the feature, and it is implemented the way I imagined:

- Dependency resolution outsourced to the `coprtree` project
- All builds submitted at once as build batches and then one wait at the end
- Loose dependency on the `python3-coprtree` package that hits users only when trying to use the feature.

I didn't test yet but the demo looks very promising and is much appreciated :-)

cli/copr_cli/helpers.py
sundaram123krishnan: Yes, this is much better. Made the change. Thank You

cli/copr_cli/main.py
sundaram123krishnan: Lol let me think if I can come up with something better

sundaram123krishnan: I have flattened the function a bit, please take a look and let me know

cli/copr_cli/main.py
```diff
@@ -539,6 +544,82 @@ def process_build(self, args, build_function, data):
...
+        )
+        env = BuildEnv(chroot=args.chroots[0],
+                       copr_project="{0}/{1}".format(username, projectname))
+
+        if not target.version:
+            print("No package version specified, building the latest version "
+                  "of {0}".format(target.name))
+
+        try:
+            levels = list(resolve_dependencies(target, env))
```
praiskup, line 603: I am unable to find a multi-level package.  Can you give me some example?

sundaram123krishnan: - pyinfra
- poetry
- prefect
- pydantic-ai

sundaram123krishnan: `dbt-extractor` had deeper deps levels I think, 
I am afk right now, sry

cli/copr_cli/main.py
praiskup: My slight preference would be to do it now, so the method that submits the build is pypi agnostic.
Not a big deal, just a preference.

cli/copr_cli/main.py
```diff
@@ -539,6 +544,82 @@ def process_build(self, args, build_function, data):
...
+        """
+        # pylint: disable=too-many-locals
+        try:
+            # We want this to be an optional feature and don't require every
+            # user to install the dependency. It may not even be available on
+            # all systems where we need the copr-cli to run.
+            # pylint: disable=import-outside-toplevel
+            from coprtree.coprtree import resolve_dependencies
+            from coprtree.exceptions import CoprtreeError
+            from coprtree.models import BuildEnv, BuildTarget
```
praiskup, line 576: We should have Suggests/Recommends in the specfile.

cli/copr_cli/main.py
```diff
@@ -539,6 +551,82 @@ def process_build(self, args, build_function, data):
...
+            raise argparse.ArgumentTypeError(
+                "The --with-deps option requires exactly one --chroot for now"
+            )
+
+        username, projectname, project_dirname = self.parse_dirname(args.copr_repo)
+
+        target = BuildTarget(
+            provider=source_method.provider, # only pypi is supported for now
+            name=source_method.data["pypi_package_name"],
+            version=source_method.data["pypi_package_version"],
```
praiskup: I am not confident it's OK; namely, I meant before that I don't like the `pypi_` fields here.  What if source_method was a method instad, taking just package object (with name & version fields, as it comes from coprtree API)?  Something like `build_cb=pypi_cb`, where

```python
def pypi_cb(package, **kwargs):
    return build_proxy.create_from_pypi(
    pypi_package_name=package.name,
    pypi_package_version=package.version,
    **kwargs)
```

?

sundaram123krishnan: Ohh i see what you meant, yes definitely the whole `pypi_` should be removed from the method then, i thought we would be targeting that later, but yes, good to do it now itself

So the `SourceMethod` would be:
```python
SourceMethod = namedtuple("SourceMethod", ["provider", "build_function"])
```
where the `build_function` would be `pypi_cb`

sundaram123krishnan: I have made the changes @praiskup, please take a look and let me know

FrostyX: I was thinking about a change like this. Did you have the same idea in mind, @praiskup ?

```diff
--- a/cli/copr_cli/main.py
+++ b/cli/copr_cli/main.py
@@ -469,11 +469,18 @@ class Commands(object):
             "python_versions": args.pythonversions,
         }
         if args.with_deps:
+            from coprtree.models import BuildTarget
             source_method = SourceMethod(
                 provider="pypi.org",
-                build_function=self.pypi_cb,
+                build_function=self.client.build_proxy.create_from_pypi,
+                data=data,
             )
-            return self.process_build_with_deps(args, source_method)
+            target = BuildTarget(
+                provider="pypi.org",
+                name=args.packagename,
+                version=args.packageversion,
+            )
+            return self.process_build_with_deps(args, source_method, target)
         return self.process_build(args, self.client.build_proxy.create_from_pypi, data)

     @requires_api_auth
@@ -560,7 +567,7 @@ class Commands(object):
         self.print_build_info_and_wait(builds, args)


-    def process_build_with_deps(self, args, source_method):
+    def process_build_with_deps(self, args, source_method, target):
         """
         Resolve the dependency tree of the requested package, and submit each
         level of the tree as a separate batch.
@@ -587,11 +594,6 @@ class Commands(object):

         username, projectname, project_dirname = self.parse_dirname(args.copr_repo)

-        target = BuildTarget(
-            provider=source_method.provider, # only pypi is supported for now
-            name=args.packagename,
-            version=args.packageversion,
-        )
         env = BuildEnv(chroot=args.chroots[0],
                        copr_project="{0}/{1}".format(username, projectname))
```

The thing that I don't like about the current implementation is that it defines the `pypi_cb` function, which is the pattern that we would have to follow when adding support for more ecosystems.

sundaram123krishnan: I am not tagged here, but I can see one issue with this implementation:
By moving `from coprtree.models import BuildTarget` inside the `action_build_pypi` method, we lose our friendly message of `ImportError`
```python
            from coprtree.coprtree import resolve_dependencies
            from coprtree.exceptions import CoprtreeError
            from coprtree.models import BuildEnv
        except ImportError as err:
            raise CoprException(
                "The --with-deps option requires the coprtree library to be "
                "installed, install it using sudo dnf install python3-coprtree"
            ) from err
```

praiskup: Yeah, I was thinking of moving BuildTarget out like you suggest @FrostyX, but the current implementation seems good enough (considering other methods will also use `args.*` the very same way).  It lgtm.

cli/copr_cli/helpers.py
```diff
@@ -21,6 +22,38 @@ def cli_use_output_format(parser, default='json'):
...
+        help="Build the package dependencies for a given package using the "
+             "coprtree library. More information can be found at: "
+             "https://github.com/sundaram123krishnan/coprtree",
+    )
+
+
+def batch_buildopts(buildopts, batch, previous_batch):
+    """
+    Decide in which batch the current package should be built.
+    """
```
praiskup: I must admit I don't get the point of this method, though :-)

cli/copr_cli/main.py
```diff
@@ -539,6 +560,84 @@ def process_build(self, args, build_function, data):
...
+            levels = list(resolve_dependencies(target, env))
+        except CoprtreeError as err:
+            raise CoprException(
+                "Can not resolve the dependencies: {0}\n"
+                "Please report this at "
+                "https://github.com/sundaram123krishnan/coprtree/issues"
+                .format(err)) from err
+
+        builds = []
+        previous_batch = None
```
praiskup:
```python
build_opts = buildopts_from_args(args)
batch = build_opts.pop("with_build_id", None)
previous_batch = build_opts.pop("after_build_id", None)
```

cli/copr_cli/main.py
```diff
@@ -539,6 +560,84 @@ def process_build(self, args, build_function, data):
...
+        for number, level in enumerate(levels):
+            batch = None
+            for index, package in enumerate(level):
+                buildopts = batch_buildopts(
+                    buildopts_from_args(args), batch, previous_batch)
```
praiskup:
```python
actual_options = copy(build_opts)
actual_options.update(....)

File: cli/copr_cli/helpers.py (hunk 12 above)
Comment (praiskup) [reply cli/copr_cli/helpers.py]: I got it, but I don't think it's right -- it is better to keep the options provided by user working, unless I'm missing something.

Comment (sundaram123krishnan) [reply cli/copr_cli/helpers.py]: ~~If user provides their own `--with-build-id` or `--after-build-id`, then it will reorder the deps levels and cause problems~~

Comment (sundaram123krishnan) [reply cli/copr_cli/helpers.py]: My bad, it took some time for me to understand your comment, 
Yes definitely we should be having  `--with-build-id` and `--after-build-id` when specified by the user.
My comment in `batch_buildopts` in `helpers.py` was misleading and i shouldn't ignore those two values

File: cli/copr_cli/main.py (hunk 14 above)
Comment (sundaram123krishnan) [reply cli/copr_cli/main.py]: I also moved `batch=None` to the end of the level loop, so the `--with-build-id` is preserved

Comment (sundaram123krishnan) [reply cli/copr_cli/main.py]: Thank you for the detailed review!

File: cli/copr_cli/main.py (hunk 11 above)
Comment (FrostyX) [reply cli/copr_cli/main.py]: Ok, no problem with me, let's keep the current implementation. We can always refactor later if needed.

File: cli/copr_cli/main.py [hunk 15]
```diff
@@ -539,6 +560,91 @@ def process_build(self, args, build_function, data):
...
+
+        batch = build_opts.pop("with_build_id", None)
+        previous_batch = build_opts.pop("after_build_id", None)
+        for number, level in enumerate(levels):
+            for index, package in enumerate(level):
+                actual_options = copy(build_opts)
+                if batch is not None:
+                    actual_options.update({"with_build_id": batch})
+                elif previous_batch is not None:
+                    actual_options.update({"after_build_id": previous_batch})
```
praiskup, line 622: Thank you for the update!

I'm still not sure.  If we set `after_build_id`, Copr creates a separate batch for every build in given level.
It would be better to set "after_build_id"  for the first build, and then do "with_build_id".

praiskup:
otherwise looks good

cli/copr_cli/main.py
praiskup: Can you please update ASAP?  We'd like to merge & make it part of the next release...

sundaram123krishnan: Sure, I am afk let me get to a PC

sundaram123krishnan: From your blog:
https://pavel.raiskup.cz/blog/build-ordering-by-batches-in-copr.html
It says that `--after-build-id` always creates a new batch for a new build @praiskup

sundaram123krishnan: If batch is None, it will create a new batch for a level

praiskup: correct, each level should be a single batch :+1:

praiskup: What I mean, each level = one batch of _several builds_.  Right now each level is several batches... (that's what I dislike).  Does it make sense?

sundaram123krishnan: That's **precisely** what it does.
Each level is a batch and it contains multiple builds. Now there can be many levels for a package. 
Here's the full dependency tree that needs to be resolved in order to build `pydantic-ai`:
```
level 0: ['genai-prices', 'griffelib', 'logfire-api']
level 1: ['pydantic-graph']
level 2: ['pydantic-ai-slim']
level 3: ['pydantic-ai']
```
in order to build `pydantic-graph` (level 1) it depends on level 0. So level 0 is **one** batch build and level 1 (next batch)  will only start the build after it's previous level gets resolved, and so on....

sundaram123krishnan: image

sundaram123krishnan: Does it make sense ? @praiskup

praiskup: Ah, ha!  I missed the 'batch = batch or build.id' line.

Yes, that makes sense!

praiskup (approved):

sundaram123krishnan:
Thank you for the merge! @praiskup

PR #4412, opened 2026-07-31
backend: retry failed Pulp background tasks with exponential backoff

### LINKED ISSUE #4317: What if background/async PULP task fails?
Should we re-try when `self.client.wait_for_finished_task` returns failure?

praiskup:
Introduce request_and_wait() that unifies the send-request + wait-for-task pattern and retries the whole cycle on task failure using exponential backoff (capped at 5 min between retries, 24 h total).

Fixes: #4317
Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -32,6 +32,7 @@ BuildRequires: util-linux
 
 BuildRequires: python3-devel
 
+BuildRequires: python3-backoff
 BuildRequires: python3-copr
 BuildRequires: python3-copr-common >= %copr_common_version
 BuildRequires: python3-daemon
@@ -88,6 +89,7 @@ Requires:   python3-resalloc >= 3.0
 Requires:   python3-retask
 Requires:   python3-setproctitle
 Requires:   python3-tabulate
+Requires:   python3-backoff
 Requires:   python3-boto3
 Requires:   python3-cachetools
 Requires:   redis
--- a/backend/copr-backend.spec
+++ b/backend/copr-backend.spec
@@ -32,7 +32,6 @@ BuildRequires: util-linux
 
 BuildRequires: python3-devel
 
-BuildRequires: python3-backoff
 BuildRequires: python3-copr
 BuildRequires: python3-copr-common >= %copr_common_version
 BuildRequires: python3-daemon
@@ -89,7 +88,6 @@ Requires:   python3-resalloc >= 3.0
 Requires:   python3-retask
 Requires:   python3-setproctitle
 Requires:   python3-tabulate
-Requires:   python3-backoff
 Requires:   python3-boto3
 Requires:   python3-cachetools
 Requires:   redis
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -13,8 +13,6 @@ import hashlib
 from itertools import batched
 from urllib.parse import urlencode
 
-import backoff
-
 from copr_common.request import SafeRequest
 from copr_common.lock import Lock
 from copr_common.redis_helpers import get_redis_connection
@@ -219,21 +217,20 @@ class BatchedAddRemoveContent:
         """
         if self.check_processed():
             self.log.info("Task processed by other process")
-            return True
+            return
 
         data = self.options()
         dirs_to_delete = data.pop("dirs_to_delete", [])
 
         try:
             if not data["add_content_units"] and not data["remove_content_units"]:
-                return True
-            if not self.client.request_and_wait(
-                lambda: self.client.modify_repository_content(repository, **data),
-                f"modify repository content {repository}",
-            ):
-                return False
-
-            return self.client.publish(repository)
+                return
+            self.client.deliver_and_wait([
+                self.client.modify_repository_content(repository, **data),
+            ])
+            self.client.deliver_and_wait([
+                self.client.publish(repository),
+            ])
         finally:
             self._delete_dirs(dirs_to_delete)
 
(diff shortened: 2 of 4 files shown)


backend/copr_backend/pulp.py
```diff
@@ -572,16 +569,15 @@ def create_content_chunked(self, path, labels):
...
         # compared to standard uploads
-        task = response.json()["task"]
-        data = self.wait_for_finished_task(
-            task, f"upload chunked {path}",
+        data = self.request_and_wait(
+            lambda: self.send("POST", commit_url, data=commit_data),
+            f"upload chunked {path}",
         )
         if not data:
             raise RuntimeError(f"Failed to upload chunked {path}")
```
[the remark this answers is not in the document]
praiskup: Truth, @FrostyX we don't use this method, should I rather avoid touching this method?

backend/copr_backend/pulp.py
```diff
@@ -457,18 +455,20 @@ def update_distribution(self, distribution, publication=None, repository=None):
...
+        """
+        Update an RPM distribution and wait for the async task to finish.
+        """
+        return self.request_and_wait(
+            lambda: self._send_update_distribution(
+                distribution, publication=publication,
+                repository=repository,
+            ),
+            f"update distribution {distribution}",
+        )
```
FrostyX: I am on the fence about whether we want to merge the request operation and the wait operation into one `request_and_wait`. I know we don't currently wait for, e.g. the `PulpClient.delete_repository` and `PulpClient.delete_distribution` but we IMHO probably should. And if we look at how we call them

https://github.com/fedora-copr/copr/blob/00d2792d335c39672e7781292e4c6fa499d825e2/backend/copr_backend/storage.py#L530-L542

it would be a waste of time to `request_and_wait` for every call in the loop, and would probably be better to request all of them and then wait at the end.

Brainstorming here ... what if we instead return a `PulpResponse` object from the `update_distribution` method (and other methods), which would be something like this

```python
class PulpResponse:
    def __init__(self, response):
        self.response = response

    @property
    def is_task(self):
        return self.response == 202

    def wait(description):
        # We would use the backoff + wait_for_finished_task here
        # We can use `self.response.request` to access the original request
        # and resend it using:
        #    session = requests.Session()
        #    retry = session.send(response.request)
        ...
```

and therefore the usage could be 

```python
response = self.send("PATCH", url, data)
if response.is_task:
    return response.wait(f"update distribution {distribution}")
if not response.ok:
    self.log.error("Failed to update distribution %s: %s",
                   distribution, response.text)
    return False
return True
```

or even just

```python
response = self.send("PATCH", url, data)
return response.wait(f"update distribution {distribution}")
```

depending on how much logic would be in the `wait` method?

Or even a minimal fix with just changing `wait_for_finished_task` to take a `response` instead of a `task` so that it can re-request if the task failed?

Do you still prefer the `request_and_wait` implementation?

praiskup: Thank you very much for the **needed** feedback!

> I know we don't currently wait for, e.g. the PulpClient.delete_repository and PulpClient.delete_distribution but we IMHO probably should.

For sure.

> it would be a waste of time to request_and_wait for every call in the loop,

Do we want to have a vectorized method?

> def wait(description):

Important detail for me: I wouldn't like to call it "wait()" if the method re-requests.  I was initially confused reading your pseudocode (I wanted to complain it actually isn't solving the problem I am trying to fix).

> retry = session.send(response.request)

Is this 100% going to reproduce the original request?

> Do you still prefer the request_and_wait implementation?

As I said in the morning, I am not excited about the solution, just ... something has to happen, ideally before the release.

```
response = self.send("PATCH", url, data)
if response.is_task:
    return response.wait(f"update distribution {distribution}")
if not response.ok:
    self.log.error("Failed to update distribution %s: %s",
                   distribution, response.text)
    return False
return True
```

I don't like thinking about all of this logic ^^^ for every PulpClient request...  (compared to the request_and_wait, which is almost a one-liner).

I am fine with additional abstraction, but I dislike doing a request from a response; can we iterate more and have `s/PulpResponse/PulpRequest/` instead?

```
request = self.client.delete_distribution(distribution["pulp_href"])
retry_pulp_requests([request])  # no error checking, raising errors?
```

praiskup:
@FrostyX PTAL

backend/copr-backend.spec
```diff
@@ -88,6 +88,7 @@ Requires:   python3-resalloc >= 3.0
 Requires:   python3-retask
 Requires:   python3-setproctitle
 Requires:   python3-tabulate
+Requires:   python3-backoff
```
praiskup: no longer needed

backend/copr_backend/pulp.py
```diff
@@ -436,17 +441,14 @@ def create_distribution(self, name, repository, basepath=None):
...
-    def update_distribution(self, distribution, publication=None, repository=None):
+    def update_distribution(self, distribution, publication=None,
+                            repository=None):
         """
-        Update an RPM distribution
-        https://docs.pulpproject.org/pulp_rpm/restapi.html#tag/Distributions:-Rpm/operation/distributions_rpm_rpm_update
-
-        This allows us to point a distribution to either a publication or
-        a repository. Not both, that doesn't make sense and Pulp would raise
-        "Only one of the attributes 'repository' and 'publication' may be used simultaneously."
```
FrostyX, line 432: This is IMHO still a helpful comment

backend/copr_backend/pulp.py
```diff
@@ -217,26 +217,20 @@ def _execute_locked(self, repository):
...
-                task, f"modify repository content {repository}",
-            ):
-                return False
-
-            return self.client.publish(repository)
+                return
+            self.client.deliver_requests([
+                self.client.modify_repository_content(repository, **data),
+            ])
+            self.client.deliver_requests([
```
FrostyX: Nitpicking here, but what do you think of `deliver_and_wait` instead of `deliver_requests`? I like the word "deliver" because it kinda implies the retries on failure. But it would be nice to also be clear about the waiting for background task feature that it has.

FrostyX (commented):
This is definitely the best version we've had so far. I like this.

praiskup:
I'd prefer having #4414 merged first.

backend/copr_backend/pulp.py
praiskup: done

backend/copr_backend/pulp.py
praiskup: returned

praiskup:
hang???  

/packit test

praiskup:
/packit test

copr-dev repo needed a fix

PR #4420, opened 2026-08-06
Fix wait for Pulp publication when creating devel repository

### LINKED ISSUE #4417: copr create --disable_createrepo true is broken with PULP
For some reason, the initial build ends in the non-devel repository.

FrostyX:
Fix #4417

The bug was introduced in 2fe7094a6.

### DIFF
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -4,6 +4,7 @@ Support for various data storages, e.g. results directory on backend, Pulp, etc.
 
 import os
 import gc
+import time
 from tempfile import TemporaryDirectory
 from concurrent.futures import ThreadPoolExecutor, as_completed
 # We need to stop using the deprecated distutils module.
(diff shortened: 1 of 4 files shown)


backend/copr_backend/storage.py
```diff
@@ -358,6 +353,11 @@ def init_project(self, dirname, chroot, reason=None):
         # If a project enabled the manual createrepo mode, we need to create a
         # devel distribution to be consumed by other builds within the project
         if self.devel:
+            task = response.json()["task"]
+            if not self.client.wait_for_finished_task(
+                task, f"create distribution {distribution_name}",
+            ):
+                return False
             response = self.client.get_publication(repository_href)
```
praiskup, line 383: Can we have something like this?
```python

def wait_for_first_pulbication(repository_href):
    while True:
        response = self.client.get_publication(repository_href)
        if response.json()["results"]:
            return response
        sleep(10)

File: backend/copr_backend/storage.py [hunk 2]
```diff
@@ -149,6 +148,34 @@ def find_build_results(self, results_dir):
                 results.append(path)
         return results
 
+    def wait_for_repo(self, dirname, chroot, repo_id, baseurl):
+        """
+        Wait a while for initial createrepo, and eventually fail the build
+        if the waiting is not successful.
+        """
+        waiting_since = time.time()
+        while time.time() - waiting_since < 60:
```
praiskup, line 159: Ohh, we've been only waiting 60 seconds?  Interesting, I would intuitively give it at least 5 minutes.

backend/copr_backend/storage.py
```diff
@@ -345,6 +374,12 @@ def init_project(self, dirname, chroot, reason=None):
         # If a project enabled the manual createrepo mode, we need to create a
         # devel distribution to be consumed by other builds within the project
         if self.devel:
+            baseurl = "{0}/{1}/".format(
+                self.opts.pulp_content_url,
+                public_distribution_name,
+            )
+            self.wait_for_repo(dirname, chroot, "copr_base", baseurl)
```
praiskup, line 381: 👍

praiskup (approved):
Considering this gives us ✔️  LGTM!  Thanks.

backend/copr_backend/storage.py
FrostyX: Oh, I didn't read this part of the code since I was just moving it around. Yeah, that surprises me too. 60 is not that long.

FrostyX:
> Considering this gives us ✔️ LGTM! Thanks.

CI passed, so merging :-)

PR #4423, opened 2026-08-10
ci: provide PULP_SECRET_KEY for testing-farm use-case

praiskup:
To make tests work, we also need to provide this, per: https://pulpproject.org/pulpcore/docs/admin/reference/settings/#secret_key

This is a new thing, got to the ghcr.io/pulp/pulp:latest yesterday :-)

Relates: #4421


nikromen (approved):
pulp container is doing something

PR #4434, opened 2026-08-18
discussion: add live metrics for copr-rpmbuild (backporting to mock)

nikromen:
tldr; Do not merge.... since there was interest for collecting more data from builds, here's my experiment that I had planned to bring some day. The current system_monitor plugin in mock does not bring enough data, let's compare it with my solution here and potentionally discuss how to backport this to mock (I want initial discussion here to filter first potential bad decisions/ideas early)

this is experimental opt-in feature for sampling resource usage (memory, swap, CPU, i/o, PSI pressure, procces count) every N seconds, then alongside posting aggregate.

The motivation is to eventually feed this data to rpmeta to predict potentional bottleneck (whether build needs more CPU; i/o, ... and to what degree).

mock currently has system_monitor plugin, however that adds little (if any) benefitial value for rpmeta, as it only monitors peak ram, swap, and heaviest process. No CPU, I/O, network, ... to identify what is bottleneck (and for how long) we need periodic sampling (speaking of https://github.com/rpm-software-management/mock/pull/1748)

My idea on how to collect the data here is to wrap mock process itself with systemd-run to one named unit: `systemd-run --unit=XXX <opts>`, then sample the resulting cgroups. This catches every subprocesses with it.
One gotcha though: with isolation=nspawn, it escapes into its own cgroup instead, this is resolved via `--keep-unit`

I would like to hear you ideas on this approach, and whether you are fine with me trying to backport it to mock itself.
Also what you think of a strategy how to backport this: In mock it can't be wrapped with systemd-run though, instead it creates its own group on startup, move mock PID to it and use the ---keep-unit trick. It would also hook into the same lifecycle of hooks (logged then to state.log). This gets bootstrap, resolving deps, actual build and %check (with ability to tell when was which part I assume?). This would replace system_monitor current mechanism (or rather extend it to more hooks), which flaw is that it only catches one nspawn invocation - missing the whole dependency install, srpm rebuild for dynamic buildreqs and separate %check.

EDIT: this idea does not collect these yet, but I'd like to collect IOPS (rios/wios), disk space consumed (and potentionally network I/O during dependency resolution). This could be useful as well

### DIFF
--- a/rpmbuild/copr_rpmbuild/builders/mock.py
+++ b/rpmbuild/copr_rpmbuild/builders/mock.py
@@ -15,6 +15,10 @@ from copr_rpmbuild.helpers import (
     macros_for_task,
     mock_snippet_for_tags,
 )
+from copr_rpmbuild.resource_monitor import (
+    CgroupSampler,
+    write_resource_results,
+)
 
 log = logging.getLogger("__main__")
 
@@ -87,6 +91,7 @@ class MockBuilder(object):
             modules=self.module_setup_commands,
             copr_build_id=self.build_id,
             isolation=self.isolation,
+            resource_monitoring=self.copr_rpmbuild_config.resource_monitoring,
             macros=self.macros,
             mock_snippet=mock_snippet_for_tags(
                 self.copr_rpmbuild_config.tags_to_mock_snippet, self.tags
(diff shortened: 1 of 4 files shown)


praiskup:
>  system_monitor plugin in mock does not bring enough data

Yeah, plese fix.

> let's compare it with my solution here and potentionally discuss how to backport this to mock

TL;DR :-) feel free to backport whatever you wish.

nikromen:
ok

PR #4436, opened 2026-08-19
backend: add BuildTagCombinationLimit for multi-tag worker limits

praiskup:
The existing BuildTagLimit only matches a single tag at a time, which doesn't allow limiting builds that carry a specific combination of tags (e.g. ppc64le + on_demand_powerful).  Add a new BuildTagCombinationLimit that checks whether all tags in a given set are present on a task, with the config option
builds_max_workers_tag_combination using '+' to join tag names.

Assisted-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/backend/copr_backend/rpm_builds.py
+++ b/backend/copr_backend/rpm_builds.py
@@ -196,6 +196,19 @@ class BuildTagLimit(PredicateWorkerLimit):
         super().__init__(predicate, limit, name="tag_{}".format(tag))
 
 
+class BuildTagCombinationLimit(PredicateWorkerLimit):
+    """
+    Limit the amount of concurrently running builds that match all tags
+    in the given set simultaneously.
+    """
+    def __init__(self, tagset, limit):
+        self._tagset = frozenset(tagset)
+        def predicate(x):
+            return self._tagset.issubset(x.tags)
+        super().__init__(predicate, limit,
+                         name="tag_combination_{}".format("+".join(sorted(tagset))))
+
+
 class RPMBuildWorkerManager(WorkerManager):
     """
     Manager taking care of background build workers.
--- a/backend/conf/copr-be.conf.example
+++ b/backend/conf/copr-be.conf.example
@@ -55,6 +55,10 @@ sleeptime=30
 # Maximum number of concurrently running tasks per a build tag.
 #builds_max_workers_tag=Power9=5,Power8=10
 
+# Maximum number of concurrently running tasks per a combination of tags.
+# Use '+' to join tags in a combination.
+#builds_max_workers_tag_combination=ppc64le+on_demand_powerful=5
+
 # Maximum number of concurrent background processes spawned for handling
 # actions.
 #actions_max_workers=10
--- a/backend/copr_backend/daemons/build_dispatcher.py
+++ b/backend/copr_backend/daemons/build_dispatcher.py
@@ -8,6 +8,7 @@ from copr_backend.rpm_builds import (
     ArchitectureWorkerLimit,
     ArchitectureUserWorkerLimit,
     BuildTagLimit,
+    BuildTagCombinationLimit,
     UserSSHLimit,
     BlockedOwnersLimit,
     RPMBuildWorkerManager,
@@ -86,7 +87,7 @@ class BuildDispatcher(BackendDispatcher):
         super().__init__(backend_opts)
         self.max_workers = backend_opts.builds_max_workers
 
-        for tag_type in ["arch", "tag", "arch_per_owner"]:
+        for tag_type in ["arch", "tag", "arch_per_owner", "tag_combination"]:
             match tag_type:
                 case "arch":
                     lclass = ArchitectureWorkerLimit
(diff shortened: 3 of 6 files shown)


praiskup:
> Maximum test time '4h' exceeded.

I doubt I caused this?  Re-trying. 

/packit test

praiskup:
/packit test

Per failure: https://artifacts.dev.testing-farm.io/e9f5997b-5aa8-4670-a23d-405da5bb11b9/

praiskup:
```
16:39:45                 stdout: :: [ 16:38:40 ] :: [   PASS   ] :: Updating packages (Assert: '1' should equal '1')
16:39:45                 stdout: :: [ 16:38:40 ] :: [   PASS   ] :: Command 'copr-cli list-builds single-host-testing/TEST1787243578427488598Project10Fork > /tmp/tmp.C9tK75h1IY' (Expected 0, got 0)
16:39:45                 stdout: :: [ 16:38:40 ] :: [   PASS   ] :: A build was forked (Assert: '2' should equal '2')
16:39:45                 stdout: :: [ 16:39:43 ] :: [   PASS   ] :: Command 'yes | dnf copr enable tested-copr/single-host-testing/TEST1787243578427488598Project10Fork fedora-44-x86_64' (Expected 0, got 0)
16:39:45                 stdout: :: [ 16:39:44 ] :: [   FAIL   ] :: Command 'dnf install -y example' (Expected 0, got 1)
```

This is weird.

praiskup:
> This is weird.

This is #4440.

backend/copr_backend/rpm_builds.py
```diff
@@ -196,6 +196,19 @@ def predicate(x):
...
 
 
+class BuildTagCombinationLimit(PredicateWorkerLimit):
+    """
+    Limit the amount of concurrently running builds that match all tags
+    in the given set simultaneously.
+    """
+    def __init__(self, tagset, limit):
+        self._tagset = frozenset(tagset)
+        def predicate(x):
```
nikromen, line 206: nit: missing empty line before and after the nested funtion

nikromen (approved):

backend/copr_backend/rpm_builds.py
```diff
@@ -196,6 +196,19 @@ def predicate(x):
         super().__init__(predicate, limit, name="tag_{}".format(tag))
 
 
+class BuildTagCombinationLimit(PredicateWorkerLimit):
+    """
+    Limit the amount of concurrently running builds that match all tags
+    in the given set simultaneously.
+    """
+    def __init__(self, tagset, limit):
+        self._tagset = frozenset(tagset)
```
nikromen, line 205: hmmm due to frozenset here: this means `x86+powerful_builder` tag combination is different that `powerful_builder+x86`?? I would expect from this feature that these two are equal

nikromen (changes requested):
reverting my approve because I got confused by this https://github.com/fedora-copr/copr/pull/4436/changes#r3907449739

backend/copr_backend/rpm_builds.py
FrostyX: I don't  think they are different. This is how we initialize the objects:

```python
if lclass is BuildTagCombinationLimit:
    self.limits.append(lclass(tag.split("+"), limit))
```

So we are doing `frozenset(["x86", "powerful_builder"])` and since `frozenset` is a set, order doesn't matter.

FrostyX:
The code looks good to me but I somehow managed to crash this when testing locally:

Config:

```
builds_max_workers_tag_combination=arch_x86_64+arch_noarch=0
```

Leads to:

```
  File "/opt/copr/backend/run/copr-run-dispatcher-backend", line 35, in <module>
    _main()
    ~~~~~^^
  File "/opt/copr/backend/run/copr-run-dispatcher-backend", line 27, in _main
    opts = get_backend_opts()
  File "/opt/copr/backend/copr_backend/helpers.py", line 624, in get_backend_opts
    return config_reader.read()
           ~~~~~~~~~~~~~~~~~~^^
  File "/opt/copr/backend/copr_backend/helpers.py", line 258, in read
    opts = self._read_unsafe()
  File "/opt/copr/backend/copr_backend/helpers.py", line 393, in _read_unsafe
    opts.builds_limits = _get_limits_conf(cp)
                         ~~~~~~~~~~~~~~~~^^^^
  File "/opt/copr/backend/copr_backend/helpers.py", line 232, in _get_limits_conf
    raise CoprBackendError("Empty '{}' spec".format(option))
copr_backend.exceptions.CoprBackendError: Empty 'builds_max_workers_tag_combination' spec
```

Edit: It's the fact that I am setting the limit to zero. Not sure if that's meant to be supported but IMHO it's a reasonable value.

praiskup:
> It's the fact that I am setting the limit to zero. Not sure if that's meant to be supported but IMHO it's a reasonable value.

I don't disagree - going to add a separate commit (this is not a new thing, `limit=0` never worked).

praiskup:
ptal

FrostyX (approved):

nikromen (approved):

PR #4452, opened 2026-08-27
frontend: use engine.begin() in whooshers.on_commit() to persist writes

### LINKED ISSUE #4451: `update_indexes_quick` never updates
https://github.com/fedora-copr/copr/blob/b6f92bd964f8d9ceec10efc849cf317b329e127d/frontend/coprs_frontend/coprs/whoosheers.py#L106-L120

As per the [docs](https://docs.sqlalchemy.org/en/20/core/connections.html#commit-as-you-go), the connection should be explicitly committed or we can better use [`engine.begin()`](https://docs.sqlalchemy.org/en/20/core/connections.html#connect-and-begin-once-from-the-engine) shorthand

sundaram123krishnan:
Fixes: #4451


sundaram123krishnan:
@coderabbit resume

praiskup (approved):
Thank you.

PR #4453, opened 2026-08-27
openshift: add probes for all services

nikromen:
add probes for all services since app-sre CI requires it for adding saas.yaml file (and thus creating copr instance)

some of the probes is just simple pgrep -f <copr process that should be running> but that just check if process exists, not if it is healthy. For purposes to satisfy CI, I am making this simple, and creating follow-up (#4454) issue to implement proper HTTP /healthz endpoints for this (that will be simple anyway since the endpoints are easy to implement, but it would require release...)

### DIFF
--- a/openshift/helm/templates/backend.yaml
+++ b/openshift/helm/templates/backend.yaml
@@ -131,6 +131,11 @@ spec:
 {{ toYaml .Values.resources.backend.log | indent 12 }}
           command: ["/usr/bin/tini", "--"]
           args: ["/usr/bin/copr_run_logger.py"]
+          livenessProbe:
+            exec:
+              command: ["pgrep", "-f", "copr_run_logger"]
+            initialDelaySeconds: 10
+            periodSeconds: 30
           volumeMounts:
             - name: config
               mountPath: /etc/copr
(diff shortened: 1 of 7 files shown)


nikromen:
@FrostyX I think that after this to satisfy the CI, we would still need to add resources to values because every template has to reference it (cpu, memory limits...) - in separate PR. Currently they are pointing to values - e.g. https://github.com/fedora-copr/copr/blob/main/openshift/helm/templates/backend.yaml#L281 but missing in values, thus it would create nil exception

also we have allowed only images from `quay.io/copr` namespace but we also use pure `quay.io/centos7/nginx-120-centos7` so could we perhaps build our own image for it (`nginx_image: quay.io/centos7/nginx-120-centos7` in values and `name: quay.io/centos7/nginx-120-centos7` in copr-backend-httpd)?

openshift/helm/templates/backend.yaml
```diff
@@ -151,6 +156,11 @@ spec:
               mountPath: /var/lock/copr-backend
           command: ["/usr/bin/tini", "--"]
           args: ["/run-backend", "--sign-host", "copr-keygen", "/usr/bin/copr-run-dispatcher-backend", "actions"]
+          livenessProbe:
+            exec:
+              command: ["pgrep", "-f", "copr-run-dispatcher"]
```
praiskup, line 161: These probes are not very good in particular, it's better to detect the process works.  But we can elaborate on those in the future...

PR #4455, opened 2026-08-29
python: add function for getting a list of downloadable build results URLs

### LINKED ISSUE #4119: Helper function in python-copr to get downloadable RPM URLs
As per discussion with @LecrisUT, we could improve our API / python client.

> So on the tmt side, if you can provide the following interfaces to be Pulp backend independent, then it would help a lot:
>
> - List of build artifacts. Should be covered by `BuildChrootProxy.get_built_packages` afaict
> - Download location of the artifacts. Maybe this can be added to the `get_built_packages` dict
>
> Those are the only interfaces that I see that we are using.

FrostyX:
Fix #4119


praiskup:
Would be awesome to have a beaker/behave test for this (we would test both the backend and pulp storage).  Edit: ideally with explicit Epoch involved

praiskup:
✅ SUCCESS: runtest-download-build.sh - Completed at 11:11:17

PR #4458, opened 2026-08-31
openshift: eol centos7 for httpd + add resources

nikromen:
Rebuild backend_httpd on Fedora to align it with the rest of images under copr namespace on quay.io. Also adding some default values so appsre CI does not traceback on parsing values.yaml.


nikromen:
I will need to build and publish the image to quay, but before that @FrostyX could you please check that this works for your docker environment? I don't use docker so I can't check

docker/backend_httpd/Dockerfile
```diff
@@ -1,10 +1,19 @@
-# Crafted with the help of the SCLOrg docs:
-# https://github.com/sclorg/nginx-container/tree/master/1.20#3-prepare-an-application-inside-a-container
-
-FROM quay.io/centos7/nginx-120-centos7:centos7
+FROM registry.fedoraproject.org/fedora:43
```
praiskup, line 1: why don't we stay with RH supported nginx image?

docker/backend_httpd/Dockerfile
```diff
@@ -1,10 +1,19 @@
...
 
 MAINTAINER copr-devel@lists.fedorahosted.org
 
-COPY files/etc/nginx/conf.d/copr-be.conf /opt/app-root/etc/nginx.d/
+RUN dnf -y install nginx \
```
praiskup, line 5: Long-term, I dislike using Nginx here. RH Copr uses Apache, Fedora Copr uses lighttpd (nah, nobody else does so), and in the container world—which will be getting much more of our attention now—we use Nginx. If you're doing a major rewrite, I'd propose migrating to either lighty or (my preference) Apache.

nikromen: > —which will be getting much more of our attention now—we use Nginx. 

I probably dont understand here... to me that sentence says that we are migrating to container world (OC) where we should use nginx, but you would prefer apache? Am I reading it correctly? Since it is a bit contradicting

docker/backend_httpd/Dockerfile
nikromen: 3 reasons:
- we should pull images from one quay namespace (not hard requirement so it is just should)
- centos7 is EOL
- everything else in our stack is running on fedora

docker/backend_httpd/Dockerfile
praiskup: > where we should use nginx

Why?  We have tons of hacks for lighttpd & apache (redirects, auto-compression, etc.), and if we run nginx we have yet another HTTP server to care about.

docker/backend_httpd/Dockerfile
praiskup: > we should pull images from one quay namespace (not hard requirement so it is just should)

Citation needed?

> centos7 is EOL

Absolutely, does RH provide up2date nginx container?

> everything else in our stack is running on fedora

The more stuff we run on supported software (other teams) the better.

praiskup: https://github.com/sclorg/nginx-container ?

FrostyX: Personally, I am in the opposite camp. I am kinda proud of the fact, that we are running our servers on Fedora, so this is a welcome change from my side.

But since this is just a http server and nothing else, it IMHO doesn't matter that much.

docker/backend_httpd/Dockerfile
FrostyX: I hate the fact that we use nginx from the very beginning. It's nothing against nginx per se, it's just the fact that it is not that thing that we are using in production. So I agree with @praiskup, we should move to lighttpd or apache.

But it IMHO doesn't necesarily have to happen in this PR.

FrostyX (approved):

docker/backend_httpd/Dockerfile
praiskup: Well, the less we maintain the better.  With fedora-based images we are on our own (nothing against Fedora, it would be awesome to have Fedora-based nginx layered images to "just consume").

praiskup: I.e., I would just use `quay.io/sclorg/nginx-126-c10s` (it should work without additional changes? in the Dockerfile).

FrostyX: That's totally fine by me. The only quesiton is "we should pull images from one quay namespace (not hard requirement so it is just should)"

nikromen: I will send a link to slack channel since that is private URL...

about the nginx I still don't get it... are we really saying that in container world, in openshift, we should prefer nginx, but you don't like it anyway a don't want to use it?

nikromen:
I am merging this as is currently to unblock CI, I'll update nginx/apache issues raised in separate issue

docker/backend_httpd/Dockerfile
FrostyX: > are we really saying that in container world, in openshift, we should prefer nginx

Who is saying that? I am not saying that :D
I am saying that whatever we use in production, we should also use for development.

PR #4466, opened 2026-09-01
backend: don't upload zero bytes RPMs to Pulp

FrostyX:
Otherwise we get hit with an infinite loop of:

    Pulp: create_content: /api/v3/content/rpm/packages/upload/ /var/lib/copr/public_html/results/@freecad/nightly/fedora-30-aarch64/01121860-freecad/freecad-debuginfo-0.19-pre_18820.aarch64.rpm
    CFailed to create Pulp content for: /var/lib/copr/public_html/results/@freecad/nightly/fedora-30-aarch64/01121860-freecad/freecad-debuginfo-0.19-pre_18820.aarch64.rpm, Bad Request {"file":["The submitted file is empty."]} (attempt #595)

### DIFF
--- a/backend/run/copr-change-storage
+++ b/backend/run/copr-change-storage
@@ -178,6 +178,12 @@ def change_storage_for_project(project, dst, config):
                 to_upload = []
                 rpms = storage.find_build_results(resultdir)
                 for rpm in rpms:
+                    # It should never happen that a RPM package is 0 bytes, but
+                    # unfortunately sometimes it happened. Nobody knows why.
+                    if os.path.getsize(rpm) == 0:
+                        log.error("RPM package has zero bytes: %s", rpm)
+                        continue
+
                     # Was a package with this NEVRA already uploaded?
                     # We cannot simply uploaded.get(rpm) because the keys are
                     # full paths and we need to compare only basenames


backend/run/copr-change-storage
```diff
@@ -178,6 +178,12 @@ def change_storage_for_project(project, dst, config):
                 to_upload = []
                 rpms = storage.find_build_results(resultdir)
                 for rpm in rpms:
+                    # It should never happen that a RPM package is 0 bytes, but
+                    # unfortunately sometimes it happened. Nobody knows why.
+                    if os.path.getsize(rpm) == 0:
+                        log.error("RPM package has zero bytes: %s", rpm)
+                        continue
```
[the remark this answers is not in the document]
nikromen: I dont understand how the --project argument somehow bypasses the check?? or is this rather: hey if your build system is building during migration then there may be race?? probably... if yes then the bot does not have context that we stop builds for those we migrate. if that is the point of this comment then +1 to the PR

nikromen (approved):

backend/run/copr-change-storage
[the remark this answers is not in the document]
FrostyX: > The --project entrypoint bypasses the running-build and running-action checks. 

This is true but unrelated to this PR

>  A concurrent writer can cause a package to be skipped while it is zero bytes or upload an empty package.

I'd personally ignore this because when we are migrating a project, nobody else should touch the data

backend/run/copr-change-storage
```diff
@@ -178,6 +178,12 @@ def change_storage_for_project(project, dst, config):
                 to_upload = []
                 rpms = storage.find_build_results(resultdir)
                 for rpm in rpms:
+                    # It should never happen that a RPM package is 0 bytes, but
+                    # unfortunately sometimes it happened. Nobody knows why.
+                    if os.path.getsize(rpm) == 0:
+                        log.error("RPM package has zero bytes: %s", rpm)
```
praiskup, line 184: I'm not sure we should "almost silently" skip this (either user's or rpmbuild's) misbehavior.
Can we simply fail the build in this case, and push users to a) fix or b) report?

FrostyX: This is the migration script so it's too late here, there is nothing we can do. The package is broken even in the current backend repository so migrating it silently is IMHO the way to go.

But I agree that we should also make sure to fail a build if it produces zero bytes RPMs. But I didn't do it yet because I wasn't sure if we really need it. This could have been some temporary bug in the past, which has been resolved already (we are defaulting to the Pulp storage for some time now, and we haven't seen the traceback yet).

praiskup: aha, migration script ... ok, lgtm

praiskup (approved):

PR #4467, opened 2026-09-01
backend: respect disable createrepo when forking project to another

### LINKED ISSUE #4461: Repodata is refreshed after project fork despite `--disable_createrepo true` setting
When a Copr project is created with `--disable_createrepo true`, repository metadata (createrepo) should not be generated or updated automatically. However, uploading a prebuilt RPM to project A and subsequently forking project A into project B triggers a repodata update for project B. As a result, the backend generates metadata despite the explicit project setting.

Expected Behavior:

- Project B should honor `--disable_createrepo true`. Forking from project A should not trigger a createrepo process, and querying project B's repository via dnf should not find newly generated repository metadata.

Actual Behavior:

- createrepo runs on project B during or immediately following the fork, making the imported RPM queryable through dnf.

Steps to Reproduce:

- Create two projects with createrepo explicitly disabled:
```
copr create testproj-A --chroot custom-1-x86_64 --disable_createrepo true
copr create testproj-B --chroot custom-1-x86_64 --disable_createrepo true
```

- Upload a prebuilt binary RPM to the first project:
```
copr uploadrpm --chroot custom-1-x86_64 testproj-A jflex-1.7.0-19.el10.noarch.rpm
```

- Fork project A into project B:
```
copr fork --confirm testproj-A testproj-B
```

- Query the repository metadata for project B:
```
dnf -q --refresh --repofrompath testproj-B,http://coprbe.MY.CUSTOM.COPR.INSTANCE/results/mizdebsk/testproj-B/custom-1-x86_64/ --repo testproj-B rq
```

Output:
```
jflex-0:1.7.0-19.el10.noarch
```

Environment:

- Copr instance where direct RPM upload is allowed (not reproducible in Fedora Copr deployment).

nikromen:
this is however issue not related only to uploadrpm but to forking in general

not tested on stg yet

Fix #4461

### DIFF
--- a/backend/copr_backend/storage.py
+++ b/backend/copr_backend/storage.py
@@ -305,7 +305,8 @@ class BackendStorage(Storage):
             new_chroot_path = os.path.join(new_path, chroot)
             ensure_dir_exists(new_chroot_path, self.log)
 
-            if createrepo and not call_copr_repo(new_chroot_path, logger=self.log):
+            if createrepo and not call_copr_repo(new_chroot_path, devel=self.devel,
+                                                 logger=self.log):
                 return False
         return True
 
--- a/frontend/coprs_frontend/coprs/logic/actions_logic.py
+++ b/frontend/coprs_frontend/coprs/logic/actions_logic.py
@@ -364,6 +364,9 @@ class ActionsLogic(object):
             "copr": dst.name,
             "builds_map": builds_map,
             "storage": src.storage,
+            "ownername": dst.owner_name,
+            "projectname": dst.name,
+            "devel": dst.devel_mode,
         }
 
         action = models.Action(


backend/copr_backend/storage.py
```diff
@@ -305,7 +305,8 @@ def fork_project(self, src_fullname, dst_fullname, builds_map, createrepo=True):
             new_chroot_path = os.path.join(new_path, chroot)
             ensure_dir_exists(new_chroot_path, self.log)
 
-            if createrepo and not call_copr_repo(new_chroot_path, logger=self.log):
+            if createrepo and not call_copr_repo(new_chroot_path, devel=self.devel,
+                                                 logger=self.log):
```
FrostyX, line 309: This is only for the `BackendStorage`. I guess we will need the same fix for the `PulpStorage` as well?

nikromen:
leaving this as is before we align with mikolaj on this

frontend/coprs_frontend/coprs/logic/actions_logic.py
```diff
@@ -364,6 +364,9 @@ def send_fork_copr(cls, src, dst, builds_map):
             "copr": dst.name,
             "builds_map": builds_map,
             "storage": src.storage,
+            "ownername": dst.owner_name,
+            "projectname": dst.name,
+            "devel": dst.devel_mode,
```
praiskup, line 369: bump BE/FE protocol nunber?

frontend/coprs_frontend/coprs/logic/actions_logic.py
```diff
@@ -364,6 +364,9 @@ def send_fork_copr(cls, src, dst, builds_map):
             "copr": dst.name,
             "builds_map": builds_map,
             "storage": src.storage,
+            "ownername": dst.owner_name,
```
praiskup, line 367: Curious why we need `dst.owner_name` twice?

nikromen: `owner = self.ext_data.get("ownername")` in actions.py on line 94 takes it; however not related directly to this fix... but would be nice to fill that value and not having it none

frontend/coprs_frontend/coprs/logic/actions_logic.py
nikromen: before doing any follow-up action I want to clarify whether we really want to fix https://github.com/fedora-copr/copr/issues/4461 this way? Or we leave this behavior as is?

To me currently it makes sense to respect disable createrepo of destination fork project (the way how this PR points) however we have not come to clear conclusion, rather that this could be interpreted in numerous ways, so if at the end of the day this is not needed I'd prefer not to touch this and close the issue

praiskup:
I suppose this can be closed now, as @mizdebsk won't combine "manual createrepo" with "forking"

PR #4469, opened 2026-09-01
backend: don't batch duplicate NEVRAs into the same Pulp request

FrostyX:
This can happen on repeated mass rebuilds of the same packages and results in Pulp tracebacks:

    DuplicateContentInRepositoryError(dup_count, correlation_id)\n', 'error_code':
    387 'PLP0022', 'description': '[PLP0022] Found 0 duplicate contents in repository

Unfortunately the PRNs are different even though the NEVRAs are the same, so we can't use them for filtering. We need to query Pulp for the NEVRAs first, and make sure to use only the one from the newest build in case there are any duplicities.

### DIFF
--- a/backend/copr_backend/pulp.py
+++ b/backend/copr_backend/pulp.py
@@ -16,6 +16,7 @@ from urllib.parse import urlencode
 from copr_common.request import SafeRequest
 from copr_common.lock import Lock
 from copr_common.redis_helpers import get_redis_connection
+from copr_backend.helpers import format_filename
 
 
 # JSON data in a POST request can be large but they probably cannot be
@@ -194,9 +195,38 @@ class BatchedAddRemoveContent:
         rpms_to_add -= common
         rpms_to_remove -= common
 
+        # Make sure we are not trying to add the same NEVRAs in one request
+        rpms_to_add = self._unique_nevras(rpms_to_add)
+
         return {"add_content_units": list(rpms_to_add),
                 "remove_content_units": list(rpms_to_remove),
                 "dirs_to_delete": list(dirs_to_delete)}
+
+    def _unique_nevras(self, pulp_hrefs):
+        unique = {}
+        for pulp_href in pulp_hrefs:
+            data = self.client.get_by_href(pulp_href).json()
+
+            # We are intentionally leaving out epoch because Pulp cannot handle
+            # two different epochs in one repository
+            # https://github.com/pulp/pulp_rpm/issues/4239
+            nevra = format_filename(
+                data["name"],
+                data["version"],
+                data["release"],
+                "",
+                data["arch"],
+            )
+
+            if nevra in unique:
+                self.log.warning("Duplicate NEVRA in one request: %s ", nevra)
+
+            build_id = int(data["pulp_labels"]["build_id"])
+            if nevra not in unique or unique[nevra][1] < build_id:
+                unique[nevra] = (pulp_href, build_id)
+
+        self.log.info("Unique NEVRAs: %s", list(unique.keys()))
+        return {x[0] for x in unique.values()}
 
     def commit(self):
         """


FrostyX:
I didn't properly test it, but ... the code hopefuly makes sense?

backend/copr_backend/pulp.py
```diff
@@ -194,10 +194,25 @@ def options(self):
...
         return {"add_content_units": list(rpms_to_add),
                 "remove_content_units": list(rpms_to_remove),
                 "dirs_to_delete": list(dirs_to_delete)}
 
+    def _unique_nevras(self, prns):
+        unique = {}
+        for prn in prns:
+            data = self.client.get_by_href(prn).json()
+
+            filename = data["location_href"]
```
nikromen: location_href == NEVRA? or at least this the method name implies. AFAIU filename != NEVRA since epoch is typically missing. Would that be a problem?

nikromen: it would if user would submit in one go the same package with different epoch (real probability low though) to test something, we would mark it as duplicate. not blocking the PR because of this since we have bigger problem to solve, however it would be nice to have follow-up issue for thos

backend/copr_backend/pulp.py
```diff
@@ -194,10 +194,25 @@ def options(self):
...
+        rpms_to_add = self._unique_nevras(rpms_to_add)
+
         return {"add_content_units": list(rpms_to_add),
                 "remove_content_units": list(rpms_to_remove),
                 "dirs_to_delete": list(dirs_to_delete)}
 
+    def _unique_nevras(self, prns):
+        unique = {}
+        for prn in prns:
+            data = self.client.get_by_href(prn).json()
```
nikromen: would this work also for forks? `get_by_href` seems to parse only web urls, but fork PRN is something like `prn:blablabla` not a web url

nikromen (commented):
curious (maybe stupid? thinking about the current problem out loud) question:
- this solves issue when in batch there is multiple RPMs with same NEVRA submitted (pulp answers with "hey there is already dupe" and we got stuck). But what if the pkg with same NEVRA is already in pulp (in different, previous build) and user later spawns build with the same NEVRA. Will this situation repeat again?

in general lgtm - if the PRN will work with forks as well - I think this should help with the stuck batch situation.

backend/copr_backend/pulp.py
FrostyX: Location href == filename, so that's a good point about the epoch. We have also `data["name"]`, `data["version"]`, etc available, so we can actually put the NEVRA together if we want to.

backend/copr_backend/pulp.py
FrostyX: Good point here, I didn't try forks at all.

FrostyX:
> But what if the pkg with same NEVRA is already in pulp (in different, previous build) and user later spawns build with the same NEVRA. Will this situation repeat again?

This should be fine because that's what we do very often. The problem with duplicates is only when we try to add same NEVRAs in one request.

backend/copr_backend/pulp.py
FrostyX: Is it possible I just named the variable incorrectly? When forking a project it is also a set of hrefs ...

```
(Pdb) pp prns
{'/pulp/api/v3/content/rpm/packages/01a05ec5-5e28-77c4-bbd8-fca98a767bbe/',
 '/pulp/api/v3/content/rpm/packages/01a05ec5-5e9f-708a-9cb3-d35dc894c812/',
 '/pulp/api/v3/content/rpm/packages/01a05ec5-5f18-7aeb-a194-89cbb1b1dac7/',
 '/pulp/api/v3/content/rpm/packages/01a05ec5-5faa-72ca-8417-b6818eef01ea/'}
```

backend/copr_backend/pulp.py
praiskup: [PULP misbehaves for NEVRAs that differ just in Epoch](https://github.com/pulp/pulp_rpm/issues/4239).
But indeed, the filename might be very artificial string, reading the metadata and creating a combination of `name-version-release-arch` would be more solid

FrostyX: OK ok, let me update.

backend/copr_backend/pulp.py
```diff
@@ -194,10 +194,25 @@ def options(self):
...
+    def _unique_nevras(self, pulp_hrefs):
+        unique = {}
+        for pulp_href in pulp_hrefs:
+            data = self.client.get_by_href(pulp_href).json()
+
+            filename = data["location_href"]
+            build_id = int(data["pulp_labels"]["build_id"])
+
+            if filename not in unique or unique[filename][1] < build_id:
+                unique[filename] = (pulp_href, build_id)
```
praiskup: Having a WARNING report for duplicates would be awesome, actually.

praiskup: I wouldn't be afraid to INFO about all NVRAs that we add in the corresponding batch.

FrostyX:
Updated, PTAL

dkliban:
@FrostyX I recommend capturing the NEVRAs and pulp_created for each RPM at the time of upload. Right now we only store the pulp_href, but if the code captured the nevra and pulp_created, then you wouldn't have to query the REST API for this information later.

nikromen (approved):

FrostyX:
I am going to merge this so that we can get our production fixed. But I'll do a follow-up as @dkliban is suggesting.

backend/copr_backend/pulp.py
```diff
@@ -194,10 +195,36 @@ def options(self):
...
+    def _unique_nevras(self, pulp_hrefs):
+        unique = {}
+        for pulp_href in pulp_hrefs:
+            data = self.client.get_by_href(pulp_href).json()
+
+            nevra = format_filename(
+                data["name"],
+                data["version"],
+                data["release"],
+                data["epoch"],
```
praiskup: I wouldn't use epoch actually, because PULP ignores it as well (see my previous comment).

PR #4470, opened 2026-09-02
backend: resign when signing already signer rpm

### LINKED ISSUE #4465: Pre-signed RPMs uploaded via uploadrpm are not re-signed with Copr's GPG key
When uploading prebuilt binary RPMs via `copr uploadrpm`, Copr signs unsigned packages with the project's GPG key as expected. However, if an uploaded RPM is already signed with another GPG key, Copr does not re-sign it with its own key. The RPM served by the backend remains identical to the uploaded file.

All RPMs served by a Copr repository should be signed with the project's Copr GPG key. If an uploaded RPM already contains an external signature, Copr should re-sign the package.

nikromen:
this is noop for rpms that are not signed:
rpm --delsign ./path/to/rpm
exit code: 0

Fix #4465

### DIFF
--- a/backend/copr_backend/background_worker_build.py
+++ b/backend/copr_backend/background_worker_build.py
@@ -37,7 +37,7 @@ from copr_backend.helpers import (
 )
 from copr_backend.job import BuildJob
 from copr_backend.msgbus import MessageSender
-from copr_backend.sign import sign_rpms_in_dir, get_pubkey
+from copr_backend.sign import resign_rpms_in_dir, sign_rpms_in_dir, get_pubkey
 from copr_backend.sshcmd import SSHConnection, SSHConnectionError
 from copr_backend.vm_alloc import ResallocHostFactory
 from copr_backend.storage import storage_for_job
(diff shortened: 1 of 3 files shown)


backend/copr_backend/background_worker_build.py
```diff
@@ -717,7 +717,7 @@ def _sign_built_packages(self):
         self.log.info("Going to sign pkgs from source: %s in chroot: %s",
                       self.job.task_id, self.job.chroot_dir)
 
-        sign_rpms_in_dir(
+        resign_rpms_in_dir(
```
praiskup, line 720: Can we call resign only for uploaded RPMs? (I want to avoid unnecessary I/O consumption)
Please comment inline that resign_ (and unsign_) doesn't fail for unsigned RPMs.

nikromen: to me doing either rpm --delsign everytime or checking with

```
rpm -K <rpm package> | grep "digests signatures OK"
```

seems to be negligible difference... is it really that needed to check?

> Please comment inline that resign_ (and unsign_) doesn't fail for unsigned RPMs.

+1

backend/copr_backend/background_worker_build.py
```diff
@@ -717,14 +717,26 @@ def _sign_built_packages(self):
...
+                log=self.log,
+            )
+        else:
+            sign_rpms_in_dir(
+                self.job.project_owner,
+                self.job.project_name,
+                path,
+                self.job.chroot,
+                opts=self.opts,
+                log=self.log,
```
praiskup: can you use `action = resign_rpms_in_dir if self.job.prebuilt_rpm_urls else sign_rpms_in_dir`?

Otherwise lgtm

backend/tests/test_background_worker_build.py
```diff
@@ -353,6 +353,7 @@ def test_build_and_sign(mc_sign_one, f_build_rpm_sign_on, caplog):
     for call in expected_calls:
         assert call in mc_sign_one.call_args_list
     assert len(mc_sign_one.call_args_list) == 2
+
```
praiskup: unnecessary space change

praiskup (approved):
Thank you!

Issue #4462, opened 2026-09-01
Fork only copies the latest build of a package


mizdebsk:
When using the `copr fork` command to copy builds between projects, only the latest build of a package is copied to the target project. Previous builds of the same package present in the source project are omitted, resulting in an incomplete duplication of the project's repository.

Expected Behavior:

- Forking a project should copy all builds and package versions present in the source project over to the target project, ensuring an exact replica of the repository's contents.

Actual Behavior

- The `copr fork` command duplicates only the newest build of a package to the target project. Older builds are silently skipped and missing from the new project's repository.

Steps to Reproduce

- Create two new Copr projects (e.g., testproj-A and testproj-B):
```
copr create testproj-A --chroot custom-1-x86_64
copr create testproj-B --chroot custom-1-x86_64
```

- Upload 2 different versions of the same package (component):
```
copr uploadrpm --nowait --chroot custom-1-x86_64 testproj-A nvidia-imex-580.178.04-1.el9.x86_64.rpm
copr uploadrpm --nowait --chroot custom-1-x86_64 testproj-A nvidia-imex-595.91.07-1.el9.x86_64.rpm
```

- Verify that both builds exist in the source project:
```
dnf -q --refresh --repofrompath testproj-A,http://coprbe.SOME.COPR/results/mizdebsk/testproj-A/custom-1-x86_64/ --repo testproj-A rq
```

Output (Both versions present):
```
nvidia-imex-0:580.178.04-1.el9.x86_64
nvidia-imex-0:595.91.07-1.el9.x86_64
```

- Fork the source project into the target project:
```
copr fork --confirm testproj-A testproj-B
```

- Query the repository of the target project:
```
dnf -q --refresh --repofrompath testproj-B,http://coprbe.SOME.COPR/results/mizdebsk/testproj-B/custom-1-x86_64/ --repo testproj-B ```
```

Actual Output

Only the latest version of the package is present in the forked repository (testproj-B):

```
nvidia-imex-0:595.91.07-1.el9.x86_64
```

nikromen:
> The copr fork command duplicates only the newest build of a package to the target project. Older builds are silently skipped and missing from the new project's repository.

that is in fact standard copr behavior https://docs.copr.fedorainfracloud.org/koji_vs_copr.html#preserving-builds (mainly due to https://docs.copr.fedorainfracloud.org/user_documentation.html#how-long-do-you-keep-the-builds) this happens across forking everywhere

[...]

PR #4475, opened 2026-09-03
frontend: fork-all action

nikromen:
I want to discuss here whether the opt-in fork-all feature would work like this based on assumption either persistent=true or that we would have pulp delete after builds = 0?

ofc this is without the configuration and API implementation... but IMO this is all what is needed in the logic change and seems easy to me so I want to discuss whether there are possibly some nuances to this?

Relates: #4462


praiskup:
This approach seems OK to me.  API knob is needed, otherwise looks good.

nikromen:
@praiskup to clarift by the knob do you mean API field that allows this but we can disable it for copr instance in config? or rather knob in let's say database per project? that we will set manually perhaps?

praiskup:
Yeah, you are right that knob that can be toggled off/on via `psql`  is enough.

nikromen:
@praiskup updated the PR with the psql knob... is the idea something like it's implemented right now or something else?

frontend/coprs_frontend/coprs/models.py
```diff
@@ -394,6 +394,9 @@ class _CoprPublic(db.Model, helpers.Serializer):
     # optional tools to run after build
     fedora_review = db.Column(db.Boolean, default=False, nullable=False, server_default="0")
 
+    # fork all successful builds instead of only the latest per chroot (admin knob via psql)
+    allow_fork_all_builds = db.Column(db.Boolean, default=False, nullable=False, server_default="0")
+
```
FrostyX: I don't particularly like the idea of having users with some special privileges that are allowed to do things that others can't.

We definitely need some guard because it would be really easy to fork `@copr/PyPI` with its million builds a few times and destroy our database.

But what if we instead have a limit of how many builds we allow to fork and enable this feature for everybody?

frontend/coprs_frontend/coprs/models.py
```diff
@@ -394,6 +394,9 @@ class _CoprPublic(db.Model, helpers.Serializer):
     # optional tools to run after build
     fedora_review = db.Column(db.Boolean, default=False, nullable=False, server_default="0")
 
+    # fork all successful builds instead of only the latest per chroot (admin knob via psql)
+    allow_fork_all_builds = db.Column(db.Boolean, default=False, nullable=False, server_default="0")
```
FrostyX: I am not sure if I understand it correctly ... 

Does the knob say "when forking this project, fork all its builds" or "when forking this project, all user to specify they want to fork all builds"?

If the first one, the name is misleading and should be just `fork_all_builds`. If the second one, I think we are then using it incorrectly in the code.

frontend/coprs_frontend/alembic/versions/7a865da2b61c_add_allow_fork_all_builds_for_copr_.py
```diff
@@ -0,0 +1,28 @@
...
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('copr', sa.Column('allow_fork_all_builds', sa.Boolean(), server_default='0', nullable=False))
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column('copr', 'allow_fork_all_builds')
+    # ### end Alembic commands ###
```
FrostyX: We usually remove the alembic comments as a statement that we reviewed the auto generated code.

praiskup:
edit: wrong PR :facepalm:

frontend/coprs_frontend/coprs/models.py
nikromen: we already have `raise_if_cant_fork`

frontend/coprs_frontend/coprs/models.py
nikromen: deleted this and instead we have docs info in API docs page, PTAL

nikromen:
build seems like rawhide FTBFS? so unrelated failure

praiskup (approved):
Thank you, LGTM

praiskup:
I would ignore CI failure probably...  testing farm failure is weird (the architectures test passwed with backend storage, and failed just for pulp backend storage).

The RPM build failure seems to be some generic FTBFS:
```
+ ./manage.py test -vv
  File "/builddir/build/BUILD/copr-frontend-2.9.post1-build/copr-frontend-git-19.3e62080/coprs_frontend/./manage.py", line 13, in <module>
    from commands.flask3_wrapper import get_flask_wrapper_command
  File "/builddir/build/BUILD/copr-frontend-2.9.post1-build/copr-frontend-git-19.3e62080/coprs_frontend/commands/__init__.py", line 3, in <module>
    from coprs import models
  File "/builddir/build/BUILD/copr-frontend-2.9.post1-build/copr-frontend-git-19.3e62080/coprs_frontend/coprs/__init__.py", line 108, in <module>
    cache = Cache(app, config={
        'CACHE_REDIS_HOST': app.config["REDIS_HOST"],
        'CACHE_REDIS_PORT': app.config["REDIS_PORT"],
    })
  File "/usr/lib/python3.15/site-packages/flask_caching/__init__.py", line 211, in __init__
    self.init_app(app, config)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/flask_caching/__init__.py", line 280, in init_app
    self._set_cache(app, config)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/usr/lib/python3.15/site-packages/flask_caching/__init__.py", line 287, in _set_cache
    cache_factory = import_string(import_me)
  File "/usr/lib/python3.15/site-packages/werkzeug/utils.py", line 605, in import_string
    raise ImportStringError(import_name, e).with_traceback(
        sys.exc_info()[2]
    ) from None
  File "/usr/lib/python3.15/site-packages/werkzeug/utils.py", line 601, in import_string
    raise ImportError(e) from None
werkzeug.utils.ImportStringError: import_string() failed for 'flask_caching.backends.redis'. Possible reasons are:
```

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
```diff
@@ -396,7 +397,8 @@ def _common(ownername, projectname):
                         )
                     )
                 fcopr, _ = ComplexLogic.fork_copr(
-                    copr, flask.g.user, dstname=form.name.data, dstgroup=dstgroup
+                    copr, flask.g.user, dstname=form.name.data, dstgroup=dstgroup,
+                    all_builds=fork_all,
```
FrostyX: Not sure if this makes sense. Shouldn't we use `form.fork_all.data` here and drop the `fork_all` variable?

frontend/coprs_frontend/coprs/views/apiv3_ns/schema/schemas.py
```diff
@@ -541,6 +541,14 @@ class ProjectFork(InputSchema):
...
         )
     )
+    fork_all: Boolean = Boolean(
+        default=False,
+        description=(
+            "Fork all successful builds for each package and chroot, not only "
+            "the latest successful build per chroot. This may significantly "
+            "increase fork duration."
+        ),
+    )
```
FrostyX, line 551: Are the users accessing the API directly via `curl`/their own client implementation, or are they using `python-copr`?

nikromen: yes currently implementation is only frontend API...
depends on whether Mikolaj is OK with using forking all of the packages feature only via curl, or he needs python-copr implementation as well
@mizdebsk ^
@praiskup FYI

praiskup (approved):

frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_projects.py
nikromen: it was not in forms, so .data is not available. However I think it does not affect web fork so I added it.

Issue #4476, opened 2026-09-03
F46FailsToInstall on several Copr packages


FrostyX:
- [ ] https://bugzilla.redhat.com/show_bug.cgi?id=2517928
- [ ] https://bugzilla.redhat.com/show_bug.cgi?id=2517929
- [ ] https://bugzilla.redhat.com/show_bug.cgi?id=2517931
- [ ] https://bugzilla.redhat.com/show_bug.cgi?id=2517932

FrostyX:
The retired packages `python-XStatic-DataTables` and `python-XStatic-Patternfly` that are required for `copr-frontend` are trivial to fix, I filed for unretirement:

- https://forge.fedoraproject.org/releng/tickets/issues/13518
- https://forge.fedoraproject.org/releng/tickets/issues/13519

FrostyX:
We are now maintaining these two:

- https://src.fedoraproject.org/rpms/python-XStatic-Patternfly
- https://src.fedoraproject.org/rpms/python-XStatic-DataTables

It shouldn't be too much work. Excluding rebuild commits, retirement and unretirement commits, there was like 5 changes in 9 years. We will need to convert the specfiles to use the new python macros though.

FrostyX:
Unfortunatelly another set of `F45FailsToInstall` bugs appeared for these two packages:

- https://bugzilla.redhat.com/show_bug.cgi?id=2529369
- https://bugzilla.redhat.com/show_bug.cgi?id=2529370

PR #4478, opened 2026-09-04
Drop support for pyp2rpm

FrostyX:
See #4476
Fix RHBZ 2517932
Fix RHBZ 2517929

The pyp2rpm package is now retired in Fedora and it is probably not worth taking since the Python team is focusing on pyp2spec instead.

### DIFF
--- a/beaker-tests/Sanity/copr-cli-basic-operations/runtest-pypi.sh
+++ b/beaker-tests/Sanity/copr-cli-basic-operations/runtest-pypi.sh
@@ -41,10 +41,6 @@ rlJournalStart
     rlPhaseStartTest
         rlRun "copr-cli create --chroot $CHROOT $PROJECT"
 
-        # Submit pyp2rpm build directly.  Package is old and not buildable with
-        # pyp2spec (at least the last motionpaint version v1.52).
-        rlRun "copr-cli buildpypi $PROJECT --spec-generator pyp2rpm --template fedora --packagename motionpaint --pythonversions 3"
-
         # Submit pyp2spec build directly, generates python-pyp2spec package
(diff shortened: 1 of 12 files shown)


FrostyX:
This will require also migration that switches all existing packages from `pyp2rpm` to `pyp2spec` and also ASAP announce on copr-devel that we are going to do that with the migration to F45 and that they should make sure it will work for them.

praiskup (approved):

praiskup:
> we are going to do that with the migration to F45

Well, if we merge this PR, we are likely to deploy the change earlier than that :-/ what's your take on this?

praiskup:
> announce on copr-devel 

Please do this pre-merge :pray:

FrostyX:
Announcement on the mailing list
https://lists.fedoraproject.org/archives/list/copr-devel@lists.fedorahosted.org/thread/UADEQF3PLOXDFDWVAM6EOCJRO472NGTI/

> This will require also migration that switches all existing packages from pyp2rpm to pyp2spec

Still need to add the migration before we can merge.

FrostyX:
I added the migration, PTAL.

praiskup:
```
:: [ 13:47:31 ] :: [  BEGIN   ] :: Running 'curl https://127.0.0.1/status/stats'
:: [ 13:48:31 ] :: [   FAIL   ] :: Command 'curl https://127.0.0.1/status/stats' (Expected 0, got 22)
```
Stats are slower, not just in our staging..  but that's not related.

/packit test

praiskup (approved):
LGTM, thanks!  Glad to hear not many people are affected by the upcoming change.

PR #4479, opened 2026-09-04
backend: assign a Pulp content guard to distributions

praiskup:
Add a new `pulp_content_guard` backend configuration option holding the pulp_href of a content guard (typically a composite guard combining a few feature guards).  When set, the guard is assigned to every distribution copr creates in Pulp.

Also add the copr-pulp-set-content-guard script for (re)setting the content guard on all distributions of an existing <owner>/<project>.

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>


backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,84 @@
...
+        sys.exit(1)
+
+    client = PulpClient.create_from_config_file()
+
+    # Distributions are named <owner>/<project>/<chroot> (with an optional
+    # -devel suffix), so all distributions of a project share this prefix.
+    prefix = "{0}/".format(args.project.rstrip("/"))
+    response = client.list_distributions(prefix)
+    response.raise_for_status()
+    distributions = response.json()["results"]
```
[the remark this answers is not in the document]
praiskup: It's fine for this script, though it would be nice to place in an assert (in case there are more pages).  Can you suggest?

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,84 @@
...
+    if not content_guard:
+        print("Error: No content guard specified and pulp_content_guard is "
+              "not configured.")
+        sys.exit(1)
+
+    client = PulpClient.create_from_config_file()
+
+    # Distributions are named <owner>/<project>/<chroot> (with an optional
+    # -devel suffix), so all distributions of a project share this prefix.
+    prefix = "{0}/".format(args.project.rstrip("/"))
```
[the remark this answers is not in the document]
praiskup: how do I add one-line assert?

backend/run/copr-pulp-set-content-guard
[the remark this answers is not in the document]
praiskup: fixed, can you resolve?

backend/run/copr-pulp-set-content-guard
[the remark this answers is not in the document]
praiskup: fixed, can you resolve?

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,97 @@
...
+    # -devel suffix), so all distributions of a project share this prefix.
+    parts = args.project.rstrip("/").split("/")
+    if len(parts) != 2 or not all(parts):
+        parser.error("project must use the <owner>/<project> format")
+    prefix = f"{parts[0]}/{parts[1]}/"
+
+    response = client.list_distributions(prefix)
+    response.raise_for_status()
+    response_data = response.json()
+    assert response_data["next"] is None, \
```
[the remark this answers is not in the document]
praiskup: fixed, can you resolve?

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,97 @@
...
+        "More than one distribution page exists; refusing a partial content-guard update"
+    distributions = response_data["results"]
+
+    if not distributions:
+        print(f"No distributions found for {args.project}")
+        return
+
+    requests = []
+    for distribution in distributions:
+        print(f"Setting content guard on {distribution["name"]}")
```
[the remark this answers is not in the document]
praiskup: fixed, can your resolve?

praiskup:
Ready for review, PTAL

backend/conf/copr-be.conf.example
```diff
@@ -101,6 +101,14 @@ prune_days=14
 # some CDN caches (e.g. when RPMs in repository are re-signed).
 #aws_cloudfront_distribution=EX55ITR8LVMOH
 
+# Pulp content guard (pulp_href) assigned to every distribution that copr
+# creates in Pulp. Typically a composite content guard combining the feature
+# guards (e.g. RHEL-OS-x86_64, RHEL-OS-aarch64). When unset, distributions are
+# created without a content guard. The guard itself must be created in the Pulp
+# domain beforehand (feature and composite guards are not available in the Pulp
+# CLI, only via the REST API).
+#pulp_content_guard=/api/pulp/<domain>/api/v3/contentguards/core/composite/<uuid>/
```
FrostyX, line 110: The Pulp documentation site seems to be down at the moment, so a stupid question ... Is only one guard supported, or multiple guards? Meaning, ... should this be a list?

backend/copr_backend/pulp.py
```diff
@@ -454,7 +454,8 @@ def get_package_by_prn(self, prn):
         self.log.info("Pulp: get_by_prn: %s", uri)
         return self.send("GET", uri)
 
-    def create_distribution(self, name, repository, basepath=None):
+    def create_distribution(self, name, repository, basepath=None,
+                            content_guard=None):
```
FrostyX, line 472: I think we talked with @dkliban about having a default content guard for all distributions within a domain. So that we wouldn't have to think about it every time when creating a new distribution. I mean, this is completely fine, I was just worried about security in case we forget to set it or set it incorrectly.

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,98 @@
...
+        sys.exit(1)
+
+    client = PulpClient.create_from_config_file()
+
+    # Distributions are named <owner>/<project>/<chroot> (with an optional
+    # -devel suffix), so all distributions of a project share this prefix.
+    parts = args.project.rstrip("/").split("/")
+    if len(parts) != 2 or not all(parts):
+        parser.error("project must use the <owner>/<project> format")
+    prefix = f"{parts[0]}/{parts[1]}/"
```
FrostyX, line 68: Good, we have the `/` at the end, so we can't accidentally match distributions from another project 👍

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,98 @@
...
+
+    response = client.list_distributions(prefix)
+    response.raise_for_status()
+    response_data = response.json()
+    if response_data["next"] is not None:
+        raise RuntimeError("More than one distribution page exists; "
+                           "refusing a partial content-guard update")
```
FrostyX, line 75: Erroring out just in case is perfectly fine here, I think we should be getting just one page. But if you wanted to have a pagination, it should be as easy as

https://github.com/fedora-copr/copr/blob/12a1ee17442d494012454422befa9fef3f2da199/backend/run/copr-pulp-set-for-all#L12-L29

We can have a helper function for it in `pulp.py`.

backend/run/copr-pulp-set-content-guard
```diff
@@ -0,0 +1,98 @@
...
+        print(f"No distributions found for {args.project}")
+        return
+
+    requests = []
+    for distribution in distributions:
+        print(f"Setting content guard on {distribution['name']}")
+        requests.append(client.update_distribution(
+            distribution["pulp_href"],
+            content_guard=content_guard,
+            # Keep the old values (we use PATCH, None cleans the config).
```
FrostyX: The explanation is a little bit different. Pulp uses PATCH to only update the requested values while keeping everything else the same. For setting values and resetting everything else they use PUT.

The bug is here:

https://github.com/fedora-copr/copr/blob/12a1ee17442d494012454422befa9fef3f2da199/backend/copr_backend/pulp.py#L485-L488

We always send the `publication` and `repository` values even if they are both `None`.

FrostyX (commented):
Thank you for the PR,
overall it looks good but I have some comments.

backend/conf/copr-be.conf.example
praiskup: IIUC, pulp doesn't allow array of content guards.  I [had to create](https://redhat-internal.slack.com/archives/C04QH3LP0NQ/p1788530529877689) a composite guard (consisting of two).

backend/copr_backend/pulp.py
praiskup: I thought it is some future feature under development, and I share this sentiment with you.
The PR here is also "all-or-nothing" configuration type, though (once we configure, all (re-)created distributions will be protected...).

backend/run/copr-pulp-set-content-guard
praiskup: Being partlly lazy, and expecting to migrate rather small projects by this script, I would keep the current variant (this can be fixed anytime later).

backend/run/copr-pulp-set-content-guard
praiskup: :angel: I tried to rephrase, sorry for confusion.

FrostyX (approved):

praiskup:
CI failure is #4494 and #4493

PR #4485, opened 2026-09-09
keygen: generate RSA 4096 keys by default

bsmojver:
Copr generates a separate OpenPGP key for every project, and the default key length has been 2048 bits. Fedora's FUTURE crypto policy requires RSA keys of at least 3072 bits, so rpm rejects signatures made by these keys on any system running that policy:

    # update-crypto-policies --set FUTURE
    # dnf install <any-package-from-any-copr-project>

Bump the default to 4096, which is also what open-build-service switched its project keys to in early 2023.

The `key_length or 2048` fallback in create_new_key() is bumped as well, otherwise it would silently override the new default for any caller passing key_length=None. copr-backend never sends key_length in its /gen_key request, so in practice the server configuration decides.

While here, fix the docstrings claiming the key length is measured in bytes and that only 1024 and 2048 are supported.

This affects newly generated keys only. Existing projects keep their 2048 bit keys; rotating those requires re-signing already published packages and is left for a separate change.

Assisted-by: Claude Opus 5 <noreply@anthropic.com>


praiskup:
/packit test

@bsmojver what is your proposal for migration to rsa4096 keys for already built RPMs?  I suppose we have to help our users (repo consumers) to avoid unnecessary outages.

bsmojver:
Honestly, I don't have a migration plan - this change was scoped to new projects only.

Thinking about it now, maybe existing repos don't need a flag day. Given pubkey.gpg is just an armored keyring and rpm holds multiple gpg-pubkey entries, a project could publish both certs: keep existing RPMs signed with
the current 2048 key, sign new builds with a new 4096 one, and append the new cert to pubkey.gpg. Consumers pick it up via the normal import prompt and old packages keep verifying. Not sure whether dnf imports all blocks from pubkey.gpg - AI says it does.

So, maybe land this for new projects, then add an opt-in per-project key rotation action built on the existing script, so owners can migrate on their own schedule and announce it to their consumers. A blanket rotation only if 3072 ever becomes the DEFAULT policy minimum.

Does that sound reasonable, or is there a constraint on the backend side I'mnot seeing?

praiskup:
Thanks for the comment!  Ok, that sounds good to me.  Well, giving users "such a button" is relatively expensive, but we can run such a script for them if they care.

>  A blanket rotation only if 3072 ever becomes the DEFAULT policy minimum.

Any idea when this is going to happen (if at all)? If I understand correctly, the RPMs that need to be re-signed in this case are those built for distributions where FUTURE becomes the DEFAULT (some future version of Rawhide).

> Does that sound reasonable, or is there a constraint on the backend side I'mnot seeing?

I was afraid of old distributions, but RHEL6+ should accept rsa4096 keys: 

```
# gpg --list-packets /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release | grep -E "keyid|bits"
	pkey[0]: [4096 bits]
	pkey[1]: [17 bits]
	keyid: 199E2F91FD431D51
...
# rpm -qa --qf '%-30{NAME} %{SIGPGP:pgpsig}\n' | head
setup                          RSA/8, Tue 08 Nov 2016 09:51:41 AM EST, Key ID 199e2f91fd431d51
filesystem                     RSA/8, Mon 18 Jul 2011 07:32:35 AM EDT, Key ID 199e2f91fd431d51
ncurses-base                   RSA/8, Thu 05 Mar 2015 11:27:05 AM EST, Key ID 199e2f91fd431d51
nss-softokn-freebl             RSA/8, Wed 04 Dec 2019 08:26:18 PM EST, Key ID 199e2f91fd431d51
ncurses-libs                   RSA/8, Thu 05 Mar 2015 11:27:04 AM EST, Key ID 199e2f91fd431d51
nspr                           RSA/8, Wed 18 Sep 2019 08:08:45 AM EDT, Key ID 199e2f91fd431d51
libattr                        RSA/8, Fri 19 Aug 2011 08:16:48 AM EDT, Key ID 199e2f91fd431d51
zlib                           RSA/8, Tue 27 Nov 2012 04:58:54 AM EST, Key ID 199e2f91fd431d51
popt                           RSA/8, Mon 16 Aug 2010 03:23:36 PM EDT, Key ID 199e2f91fd431d51
libacl                         RSA/8, Mon 26 Jun 2017 04:53:03 AM EDT, Key ID 199e2f91fd431d51
```

PR #4486, opened 2026-09-09
rpmbuild: extract all sources not just specfile

### LINKED ISSUE #4325: Weird name/versin/release parsing failure for ruby (RHEL 9)
```
Exception appeared during handling spec file: /var/lib/copr-rpmbuild/results/ruby.spec
  File "/usr/lib/python3.14/site-packages/specfile/spec_parser.py", line 233, in get_rpm_spec
    return rpm.spec(tmp.name, flags)
           ~~~~~~~~^^^^^^^^^^^^^^^^^
ValueError: can't parse specfile


The above exception was the direct cause of the following exception:

  File "/usr/lib/python3.14/site-packages/copr_rpmbuild/automation/srpm_results.py", line 103, in get_package_info
    spec = Spec(specfile_path, macros)
  File "/usr/lib/python3.14/site-packages/copr_rpmbuild/helpers.py", line 418, in __init__
    self.spec = Specfile(path, macros=macros)
                ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/specfile/specfile.py", line 120, in __init__
    self._parser.parse(str(self))
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/specfile/spec_parser.py", line 387, in parse
    self.spec, self.tainted = self._do_parse(content, extra_macros)
                              ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.14/site-packages/specfile/spec_parser.py", line 301, in _do_parse
    spec = get_rpm_spec(content, rpm.RPMSPEC_ANYARCH | rpm.RPMSPEC_FORCE)
  File "/usr/lib/python3.14/site-packages/specfile/spec_parser.py", line 235, in get_rpm_spec
    raise RPMException(stderr=stderr) from e
specfile.exceptions.RPMException: /tmp/tmpf39jbisp: line 114: failed to load macro file /var/lib/copr-rpmbuild/results/macros.ruby
Querying NEVRA from SRPM header: /var/lib/copr-rpmbuild/results/ruby-3.0.7-162.el9.1.src.rpm
Package info: {
```

RH Copr results/praiskup/test-rhel-9.4/srpm-builds/00140074/builder-live.log.gz

FrostyX:
Fix #4325

The spec might do something like:

    Source4: macros.ruby
    %{load:%{SOURCE4}}

and therefore require the `Source4` to be available when trying to parse it using the `python3-specfile` library.

### DIFF
--- a/rpmbuild/main.py
+++ b/rpmbuild/main.py
@@ -206,8 +206,8 @@ def build_srpm(args, config):
     log.info("Output: {0}".format(
         list(entry.name for entry in os.scandir(resultdir))))
 
-    # extract spec file from SRPM
-    cmd = f"rpm2archive -n < {locate_srpm(resultdir)} | tar xf - '*.spec'"
+    # extract spec file and additional sources from the SRPM
+    cmd = f"rpm2archive -n < {locate_srpm(resultdir)} | tar xf - '*'"
     subprocess.run(cmd, shell=True, check=False, cwd=resultdir)
 
     with open(os.path.join(resultdir, 'success'), "w") as success:


rpmbuild/main.py
```diff
@@ -206,8 +206,8 @@ def build_srpm(args, config):
     log.info("Output: {0}".format(
         list(entry.name for entry in os.scandir(resultdir))))
 
-    # extract spec file from SRPM
-    cmd = f"rpm2archive -n < {locate_srpm(resultdir)} | tar xf - '*.spec'"
+    # extract spec file and additional sources from the SRPM
+    cmd = f"rpm2archive -n < {locate_srpm(resultdir)} | tar xf - '*'"
     subprocess.run(cmd, shell=True, check=False, cwd=resultdir)
```
nikromen, line 211: since now we do changes in this command and putting * in there without the spec could we delete the shell=True to mitigate the potential shell command injection?

I know this is now dangerous to us since this happens on builder but since we do change literally in the command right now I think it's worth fixing

praiskup: I'm not sure I like the fact that we unnecessarily  extract the bloat here :-( (all the sources files, etc...)

We can just do `tar xf -` (no wildcard needed?), and if we want to avoid shell=True, we can create a trivial `copr-extract-archive` script.

praiskup:
To read NEVR, the norpm (fault tolerant) parser should be enough.  The only culprit I'm afraid of is the current state of videly used rpmautospec logic.  It basically does this for us (for the DistGit method) to fill `%autorelease` contents:

```patch
diff --git a/star.spec b/star.spec
index 320026a..a5f5073 100644
--- a/star.spec
+++ b/star.spec
@@ -1,3 +1,13 @@
+## START: Set by rpmautospec
+## (rpmautospec version 0.8.4)
+## RPMAUTOSPEC: autorelease
+%define autorelease(e:s:pb:n) %{?-p:0.}%{lua:
+    release_number = 8;
+    base_release_number = tonumber(rpm.expand("%{?-b*}%{!?-b:1}"));
+    print(release_number + base_release_number - 1);
+}%{?-e:.%{-e*}}%{?-s:.%{-s*}}%{!?-n:%{?dist}}
+## END: Set by rpmautospec
+
 %global ALTERNATIVES %{_bindir}/alternatives
 %global version_schily 2024-03-21
```

Norpm can parse the getopt options (-b, -e, ...) if used in `Release`, the issue is that norpm deliberately doesn't evaluate %lua...

Sure, we could use `%[8 + %{?-b*:%{-b*}}%{!?-b:1} - 1]` instead, but this would only work with RPM v4.15+ (EPEL8+).  Work-around would be to fix rpmautospec to provide this:

```patch
diff --git a/star.spec b/star.spec
index a5f5073..9082321 100644
--- a/star.spec
+++ b/star.spec
@@ -1,11 +1,11 @@
 ## START: Set by rpmautospec
 ## (rpmautospec version 0.8.4)
 ## RPMAUTOSPEC: autorelease
-%define autorelease(e:s:pb:n) %{?-p:0.}%{lua:
+%define autorelease(e:s:pb:n) %{?-p:0.}%{?expr:%[8 + %{?-b*:%{-b*}}%{!?-b:1} - 1]}%{!?expr:%{lua:
     release_number = 8;
     base_release_number = tonumber(rpm.expand("%{?-b*}%{!?-b:1}"));
     print(release_number + base_release_number - 1);
-}%{?-e:.%{-e*}}%{?-s:.%{-s*}}%{!?-n:%{?dist}}
+}}%{?-e:.%{-e*}}%{?-s:.%{-s*}}%{!?-n:%{?dist}}
 ## END: Set by rpmautospec
```

IOW, if RPM supports expressions, use them.  Fallback to %lua only for older systems (EPEL7-).

Considering Copr just fills some metadata fields using python-specfile here (name, version, release, epoch) into the Build entry in copr-frontend, and this task doesn't even fail the build - it seems like an acceptable thing to move forward, but before we do so I'm eager to hear your opinions.

Issue #4216, opened 2026-03-12
pulp: one long obsolete package stayed around


knurd:
I don't care much, just noticed something that might be indicating that something is wrong somewhere:

I by chance noticed that pulp still has the following packages as well as binary packages build from it: https://packages.redhat.com/api/pulp-content/public-copr/@kernel-vanilla/mainline-wo-mergew/fedora-44-x86_64/Packages/k/kernel-6.19.0-0.rc8.260208.e7aa5724.457.vanilla.fc44.src.rpm

It's long obsolete and should not be there afaics (hope I don't miss anything), as can be seen by: https://copr.fedorainfracloud.org/coprs/g/kernel-vanilla/mainline-wo-mergew/builds/

FrostyX:
Hello @knurd,
thank you for the report.

There is currently a known bug when trying to delete multiple (more than 7) builds at once. For the reference #4187, and I am trying to fix it in PR #4214.

Is it possible you used that feature? If not, I need to start looking for another bug.

knurd:
> when trying to delete multiple (more than 7) builds at once. […] possible you used that feature?

Don't think I did that, but not totally sure, sorry.

FrostyX:
We will need to investigate than. Thank you :-)

praiskup:
> It's long obsolete and should not be there afaics (hope I don't miss anything)

If I interpret this correctly, the build is no longer present in Copr database.  Can you recall when/how you removed that one?

knurd:
> Can you recall when/how you removed that one?

That copr does usually build new packages daily and the old ones get removed due to "Max builds:  8" – but occasionally I delete a package manually for one reason or another. Can't remember if that might have been the case for that package, sorry.

praiskup:
Ok, no problem, thank you for the feedback!  The fact you use "max_builds" feature is an important piece.

knurd:
TWIMC, meanwhile there are eight old build that for some reason stay around in pulp and waste space (one from Feb, seven from Apr): https://packages.redhat.com/api/pulp-content/public-copr/@kernel-vanilla/mainline-wo-mergew/fedora-44-x86_64/Packages/k/

knurd:
Just a quick update:

> meanwhile there are eight old build that for some reason stay around in pulp and waste space (one from Feb, seven from Apr)

[...]

Issue #4071, opened 2025-12-12
Unexpected behavior of PULP's max_builds pruning mechanism


praiskup:
I'm following-up from #4044.

When a PULP repo contains nevras like:
```
python-copr-common 1.5.post1 1.git.3.3eb2324.fc43
python-copr-common 1.5.post1 1.git.3.776dffa.fc43
python-copr-common 1.5.post1 1.git.6.176eec3.fc43
python-copr-common 1.5.post1 1.git.7.2fef562.fc43
python-copr-common 1.5.post1 1.git.7.ab2cce0.fc43
```

And later I build (e.g., because I'm dropping commits in a pul-request):

```
/tmp/tito/python-copr-common-1.5.post1-1.git.3.2f3affb.fc43.src.rpm
```

Then Copr successfully builds, but the RPM is not hosted in PULP at all.  It's considered too old to host.

praiskup:
@dkliban we use this:
https://github.com/fedora-copr/copr/blob/538da4279cbedb34d8756b9d6627a0d4584fc00d/backend/copr_backend/pulp.py#L292

With the backend (old) storage, we used to keep all RPMs that were built within the last 14 days + one that had the latest nvr.

I think we discussed the time-based retention policy before, but I can't recall the conslusion?  This is a serious problem, though, people do "push force" in pull-requests all the time, and we need to **host the RPMs from the last build** no matter the nvr (at least for some time, ideall 14 days).

praiskup:
Can we, as a short term work-around, write a script that would set `retain_package_versions = 0` (or -1).
Let's brainstorm this on Wed with PULP folks, it's going to be a hard blocking issue for Packit, e.g.

praiskup:
Discussed in Jira: https://issues.redhat.com/browse/PULP-1056

praiskup:
Should we just always re-set the parameters using rpm_update?
https://pulpproject.org/pulp_rpm/restapi/#tag/Repositories:-Rpm/operation/repositories_rpm_rpm_update

FrostyX:
Going to continue in a new ticket

praiskup:
With @FrostyX we haven't found the new ticket :-) Seems we closed by accident during our meeting.

praiskup:
As I understand the status, the feature `retain_package_versions` is not the one we'd use -- we have to go with #4138 instead.  Right?

FrostyX:
[...]

PR #4489, opened 2026-09-11
backend: add script to set an attribute for all Pulp repositories

FrostyX:
See #4216
See #4071

Due to historical reasons, we have some projects that have `retain_package_versions` set to 5 and some projects that have it set to 0. More confusingly so, the inconsistency is on CoprChroot level and so that the cleanup can behave differently for two chroots within the same project.

    [copr@copr-be root][PROD]$ copr-pulp-query \
        --project @kernel-vanilla/mainline-wo-mergew \
        --chroot fedora-44-x86_64 \
        |grep retain_package_versions
     'retain_package_versions': 0,

    [copr@copr-be root][PROD]$ copr-pulp-query \
        --project @kernel-vanilla/mainline-wo-mergew \
        --chroot fedora-43-x86_64 \
        |grep retain_package_versions
     'retain_package_versions': 5,

Assisted-by: OpenAI Codex <codex@openai.com>

### DIFF
--- a/backend/run/copr-pulp-set-for-all
+++ b/backend/run/copr-pulp-set-for-all
@@ -1,0 +1,65 @@
+#!/usr/bin/python3
+
+"""
+Set retain_package_versions=0 on all Pulp RPM repositories.
+Rewrite to set any other attribute or make this general in the future.
+"""
+
+from urllib.parse import urlencode
+from copr_backend.pulp import PulpClient, PulpRequest
+
+
+def pulp_repositories(client):
+    """
+    Iterate over all Pulp RPM repositories.
+    """
+    limit = 100
+    offset = 0
+
+    while True:
+        query = urlencode({"limit": limit, "offset": offset})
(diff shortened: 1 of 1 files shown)


FrostyX:
Not sure if we want to merge this one or not. I can just upload it to production, run it once and throw it away. But posting it for a review first.

nikromen (approved):

backend/run/copr-pulp-set-for-all
```diff
@@ -0,0 +1,65 @@
...
+    offset = 0
+
+    while True:
+        query = urlencode({"limit": limit, "offset": offset})
+        response = client.send("GET", f"/api/v3/repositories/rpm/rpm/?{query}")
+        response.raise_for_status()
+        data = response.json()
+
+        repositories = data["results"]
+        yield from repositories
```
praiskup, line 26: nice, I never seen yield from

PR #4498, opened 2026-09-17
frontend: optimize /stats page

praiskup:


testing-farm/all-on-single-host.sh
```diff
@@ -25,6 +25,13 @@ export STORAGE
 
 export VENDOR="Testing Copr - user single-host-testing"
 
+# Warm up the /status/stats/ graph caches (BuildsStatistics / ActionsStatistics).
+# This is normally done hourly by the copr-frontend cron job, but on a freshly
+# provisioned Testing Farm machine that cron has not run yet, so the very first
+# request to /status/stats/ (see runtest-webui.sh) would compute hundreds of
+# time buckets in a single request and time out.  Populate the caches up front.
+sudo -i -u copr-fe copr-frontend update-graphs
```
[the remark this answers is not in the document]
praiskup: resolved

testing-farm/all-on-single-host.sh
```diff
@@ -25,6 +25,13 @@ export STORAGE
 
 export VENDOR="Testing Copr - user single-host-testing"
 
+# Warm up the /status/stats/ graph caches (BuildsStatistics / ActionsStatistics).
+# This is normally done hourly by the copr-frontend cron job, but on a freshly
+# provisioned Testing Farm machine that cron has not run yet, so the very first
+# request to /status/stats/ (see runtest-webui.sh) would compute hundreds of
+# time buckets in a single request and time out.  Populate the caches up front.
+sudo -i -u copr-fe copr-frontend update-graphs || exit 1
```
FrostyX, line 33: This is the only part of the PR that I find confusing. The Testing Farm machine has no data when it starts, and it only accumulates tens of builds and actions while running the tests. So even if the `/status/stats/` request was literally the last test that happens, it shouldn't take that long to build the cache from scratch?

Also, this command happens before running the tests and at that moment, we have no data yet? So it is confusing to me how this can help.

FrostyX (commented):
Thank you very much @praiskup, much appreciated. The failing CI was annoying.

testing-farm/all-on-single-host.sh
praiskup: The problem isn't that a particular query is slow ... IMO the sql queries for all the buckets are executed no matter if the data are thare, and there are _many_. 

> Also, this command happens before running the tests and at that moment, we have no data yet? So it is confusing to me how this can help.

Same reason as above :angel: not really 100% sure, but IMO it is enough that frontend is up&running.

FrostyX (approved):

testing-farm/all-on-single-host.sh
FrostyX: I still don't understand it, but if it works, it works :D

praiskup: Hmm, I'm not entirely sure, and you're not making me feel very confident :-D

To be honest, the script here only runs for about 3 seconds. It generates ~15k lines of logs, but on the frontend, we don't have a Redis logger :-/ so logging shouldn't be our bottleneck.

praiskup:
```
     SELECT               2441
     INSERT               516
```

I would simply blame the frontend load for our "initial" issues;  at the time we start testing, we run dozens of `.sh` scripts concurrently, each asking Frontend to do some stuff ... and under this storm, we do many transactions.

praiskup:
Added one more commit to minimize the number of commits (and thus fsync calls). PTAL.

I don't have definitive proof, but I suspect the system's "initial high load" was causing the frontend to time out on stats prior to this PR.  This change is strictly an optimization.

If I turn out to be wrong (which is certainly possible), we can explore other solutions.

nikromen (approved):
reviewed only changes after FrostyX's LGTM, if tests will pass, I am going to merge this

praiskup:
10:18:09                 stdout: âœ… SUCCESS: runtest-webui.sh - Completed at 10:18:09

praiskup:
seems like runtests-epel.sh hanged for some reason, but that seems to be orthogonal to this PR

PR #4500, opened 2026-09-17
frontend: fix fork-all package latest version

nikromen:
successful_build_chroots() yields builds newest-first. Fork-all then
assigns new IDs in that order, so an older source build can end up with
a higher ID in the fork.

Example: source builds 100 (v1.0) then 200 (v2.0). Forking newest-first
creates 301=v2.0, 302=v1.0; UI uses max(id) and shows v1.0 as latest.
Fork oldest→newest so max(id) still means the newest package version.


frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -167,6 +167,13 @@ def fork_copr(cls, copr, user, dstname, dstgroup=None, all_builds=False):
             if not builds:
                 continue
 
+            # get_packages_with_latest_builds_for_dir gets max(id) instead of actual
+            # NEVRA or something. Also who knows what other assumptions across the codebase
+            # are made, so keeping this sorted and forking exactly in the same order
+            # as the source breaks nothing.
+            if all_builds:
+                builds = dict(sorted(builds.items(), key=lambda item: item[0].id))
```
praiskup, line 175: Have you observed any issue with the corresponding sql query?

frontend/coprs_frontend/coprs/logic/complex_logic.py
```diff
@@ -167,6 +167,13 @@ def fork_copr(cls, copr, user, dstname, dstgroup=None, all_builds=False):
             if not builds:
                 continue
 
+            # get_packages_with_latest_builds_for_dir gets max(id) instead of actual
+            # NEVRA or something. Also who knows what other assumptions across the codebase
+            # are made, so keeping this sorted and forking exactly in the same order
+            # as the source breaks nothing.
```
praiskup, line 173: Sorry, but I don't get this comment, can you elaborate more?  If sorting matters, then for `not all_builds` and not vice versa.  :shrug:

praiskup: Aha, thinking again ... the feature bound to `Package.max_builds` keeps the newest build ids:
https://github.com/fedora-copr/copr/blob/9b79e1bd34d494f95da5eba435783b94f7847e24/frontend/coprs_frontend/coprs/logic/builds_logic.py#L1676

But I would specifically pinpoint this example so it is absoltely clear what you mean.

frontend/coprs_frontend/coprs/logic/complex_logic.py
praiskup: aha, we do this:
https://github.com/fedora-copr/copr/blob/5ed854f475cdc82d30ea65be7de109820c240fa1/frontend/coprs_frontend/coprs/logic/packages_logic.py#L417

Could we fix that and do a sorted sql query?

praiskup (approved):

PR #4504, opened 2026-09-18
frontend: allow setting module_hotfixes on external repositories

FrostyX:
Fix CPT-1264

The same way we allow setting priority on external repositories https://frostyx.cz/posts/prioritize-buildrequires-from-copr we can now also set the external_hotfixes value.


praiskup (approved):

frontend/coprs_frontend/coprs/repos.py
```diff
@@ -99,14 +100,22 @@ def parse_repo_params(repo, supported_keys=None):
...
         if k in supported_keys:
             # parse_qs returns values as lists, but we allow setting the param only once,
             # so we can take just first value from it
-            value = int(v[0]) if v[0].isnumeric() else v[0]
+            value = v[0]
+
+            if value.isnumeric():
+                value = int(value)
+            elif value.lower() == "true":
+                value = True
```
praiskup, line 115: in DNF this can be `1`

praiskup: but we can document what is and what isn't supported, I think our users may live with `true|True` only

FrostyX: `1` would be already taken by the previous statement

```python
if value.isnumeric():
    value = int(value)
```

nikromen (approved):

nikromen:
only docs changes were made from last TF run, another TF run not needed

Issue #2612, opened 2023-04-03
allow deleting a running build


praiskup:
Original issue: https://bugzilla.redhat.com/show_bug.cgi?id=1729683
Opened: 2019-07-13 13:51:58
Opened by: Till Hofmann

---

#### Till Hofmann commented at 2019-07-13 13:51:58:
Description of problem:
When deleting a running build, two messages are shown:
1. You can not delete build `971029` which is not finished. 
2. Builds 971029 have been deleted successfully.

Only one of them can be true (I think it's the first one).

How reproducible:
always

Steps to Reproduce:
1. Start a build
2. Delete the build while it's running
3.

Actual results:
The two error messages. Screenshot:
https://thofmann.fedorapeople.org/Screenshot_2019-07-13%20Builds%20for%20thofmann%20ros-testing.png

Expected results:
Ideally, the build would be canceled and then successfully deleted.

---

#### Pavel Raiskup commented at 2019-07-26 06:56:15:
Thanks for the report.

> Expected results:
> Ideally, the build would be canceled and then successfully deleted.

I agree, this is not going to be be easy change, but eventually we should
do what you propose.

---

#### Pavel Raiskup commented at 2020-12-14 14:32:57:
See also bug 1639788 (delete project with running builds).

---

#### Zbigniew Jędrzejewski-Szmek commented at 2022-02-15 09:10:02:
Also pending builds should be able to be deleted.

PR #4505, opened 2026-09-19
Add rebuild and purge method

qr243vbi:


qr243vbi:
@coderabbitai resume

python/copr/v3/proxies/build.py
```diff
@@ -414,6 +414,51 @@ def delete(self, build_id):
...
+        Delete a build, try cancel before delete
+
+        :param int build_id:
+        :return: Munch
+        """
+        try:
+            self.cancel(build_id)
+        except CoprRequestException:
+            pass
+        return self.delete(build_id)
```
FrostyX, line 429: This is basically https://github.com/fedora-copr/copr/issues/2612, right?

I think rather than having a separate `purge` method for this, we should probably add something like `force=False/True` for the `delete` method. In ideal world, we would pass that value through the API and let the server deal with it, but it may not be as simple as it sounds. So I think as a temporary workaround, it would be fine to add just the thing you are doing here:

```python
if force:
    try:
        self.cancel(build_id)
    except CoprRequestException:
        pass
```

python/copr/v3/proxies/build.py
```diff
@@ -414,6 +414,51 @@ def delete(self, build_id):
...
+                        ownername = build.ownername
+                    if projectname is None:
+                        projectname = build.projectname
+                    if project_dirname is None:
+                        project_dirname = build.project_dirname
+                    return self.create_from_url(ownername=ownername, projectname=projectname,
+                                    url=url, buildopts=buildopts, project_dirname=project_dirname)
+        raise CoprRequestException(
+            f"Source package is not available for build with id {build_id}", 404)
+
```
FrostyX, line 460: This is #708?

I think it would be better to have this on the server side. This is the code that handles the "Resubmit" button in the web UI:

https://github.com/fedora-copr/copr/blob/8589040bb211e0b59b41c4746f1e0b95b3c6eefe/frontend/coprs_frontend/coprs/views/coprs_ns/coprs_builds.py#L440-L456

It would be easy to add new API endpoint here, that does the same thing:

https://github.com/fedora-copr/copr/blob/8589040bb211e0b59b41c4746f1e0b95b3c6eefe/frontend/coprs_frontend/coprs/views/apiv3_ns/apiv3_builds.py#L623

The client code here could then be very simple and only send a request saying "hey resubmit the build 123".

FrostyX (commented):
Hello @qr243vbi,
thank you very much for the PR.

I have some comments regarding the implementation but both of these are very useful features. That's awesome.

In case you'll do changes to the server code, this is the way I am running it for a local development:
https://frostyx.cz/posts/copr-docker-compose-without-supervisord

